PR #8, opened 2018-01-25
Added ignore for Squish Coco's *csmes files to .gitignore

jcwbacker:


krlberry (dismissed):
Looks good! :+1:

krlberry:
This is good to go in whenever nightly is clean. (Not that this would change anything related to nightly being broken, just as a matter of policy.)

ihumphrey:
@jcwbacker @kberryUSGS I'm going to go ahead and merge this in. Nightly should be clean today, and as you've stated, this does not affect ISIS functionality at all, so it should be safe to merge in regardless (I agree with the matter of policy regarding clean nightly's, we should probably discuss PR's of different types and which are acceptable to merge into dev regarding the nightly status).

krlberry (approved):

PR #11, opened 2018-01-25
Added ReleaseNotes directory inside $ISISROOT/src/docsys/documents

twilson271828:
This PR adds the ReleaseNotes directory to $ISISROOT/src/docsys/documents/ as well as a README.md file inside that directory.  The isis3mgr createReleaseNotesXML script is expecting this directory to be in place before it is run, and it didn't make it over from the svn -> git migration because git doesn't like empty directories (hence, adding the README.md file inside it).  References #5267


ihumphrey:
I'm closing this PR as you've made changes to the createReleaseNotesXML php script to make this empty directory when it executes (as it is the only thing that populates and uses the ReleaseNotes directory).

PR #17, opened 2018-01-27
Removed easier ControlCubeGraphNode usage outside of ControlNet

jessemapel:
This PR removes all of the usages except for CnetEditorWidget. Its usage is more complicated and will need more work to address.


jessemapel:
Nevermind, this was merged in prior to the re-migration so it's already in!

PR #26, opened 2018-01-30
Updated ciss2isis documentation to reflect newly added targets in translation file. Fixes #4970.

### LINKED ISSUE #4970: Issues initializing camera model
**ISIS version(s) affected**: 6

**Description**  
We have been attempting to get the Ames Stereo Pipeline back up and running on an Ubuntu 14 machine, in order to produce CTX DEMs. Previously, you helped me upgrade ISIS from ISIS 3 to the most recent version, ISIS 6. Now, when we run "parallel_stereo" we get errors related to ISIS being unable to find the shape model file (error message below). It almost seems like it cannot file the data directory.

**How to reproduce**  

> parallel_stereo D17_033814_1526_XI_27S046W.map.cub D18_034091_1526_XI_27S046W.map.cub results/out

Error: **ERROR** Unable to initialize camera model from group [Instrument].
**PROGRAMMER ERROR** Unable to create a shape model from given target and pvl.
**I/O ERROR** Invalid shape model file [$base/dems/molaMarsPlanetaryRadius0005.cub] in Kernels group.
**ERROR** The given shape model file is not a valid ISIS DEM. Unable to open as an ISIS cube.
**I/O ERROR** Unable to open [$ISIS3DATA/base/dems/molaMarsPlanetaryRadius0005.cub].
**ERROR** The given shape model file is not a valid NAIF DSK file. Unable to construct a NAIF DSK shape model.
**USER ERROR** NAIF DSK file [$base/dems/molaMarsPlanetaryRadius0005.cub] does not exist.

**Possible Solution**  
As I mentioned earlier, I wonder if ISIS is no longer able to find the data directory, after my recent update of the software. However, when I type "echo $ISISDATA", it directs me to the correct folder.

Any help you could provide would be appreciated!

Thanks, Catherine

chrisryancombs:


celbal (changes requested):
Need a history comment in ciss2isis.xml

chrisryancombs:
Suggested changes made.

celbal (approved):

PR #38, opened 2018-02-02
Merged dev into CNetImprovements branch

jessemapel:


jessemapel:
Self-merging because I can

PR #40, opened 2018-02-03
LidarData json and binary serialization

ihumphrey:


ihumphrey:
@makaylas approved this PR

PR #66, opened 2018-02-10
Changed tgocassisunstitch to write all lines in a cube.

AgoinsUSGS:
tgocassisunstitch would previously write the last line in cubes as a line of null DN's.
tgocassisunstitch.cpp was modified to write up to and including the last line.

This causes inconsistencies with the old truthdata, however, because the old truthdata included the null lines.

For the sake of not breaking nightly, the updated truthdata for tgocassisunstitch tests have been placed in the following location:

/work/users/agoins/TGO_TRUTHDATA

It will rest here until this PR gets merged and it can be updated in the test directory.

### DIFF
--- a/isis/src/tgo/apps/tgocassisunstitch/tgocassisunstitch.cpp
+++ b/isis/src/tgo/apps/tgocassisunstitch/tgocassisunstitch.cpp
@@ -25,17 +25,17 @@ using namespace Isis;
 
 /**
  * Struct for storing information about a filter.
- * 
+ *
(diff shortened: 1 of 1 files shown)


krlberry (approved):
:+1: Thanks! This looks good. I'm approving it, but holding off on merging it because it needs a redmine ticket: https://trello.com/c/FLgesGba

krlberry:
Test data updated from `/work/users/agoins/TGO_TRUTHDATA`

PR #76, opened 2018-02-15
fits2isis tests now check the labels. Also fixed a few typos in the mdis2isis's default test. Fixes #4923.

kaitlyndlee:
fits2isis tests now check the labels. Also fixed a few typos in the mdis2isis's default test. Fixes #4923.

### DIFF
--- a/isis/src/base/apps/fits2isis/tsts/default/Makefile
+++ b/isis/src/base/apps/fits2isis/tsts/default/Makefile
@@ -4,10 +4,6 @@ include $(ISISROOT)/make/isismake.tsts
 
 commands:
 	$(APPNAME) from=$(INPUT)/WFPC2u5780205r_c0fx.fits \
-	to=$(OUTPUT)/fitsTruth.cub \
-	> /dev/null;
-	raw2isis from=$(INPUT)/WFPC2u5780205r_c0fx.fits \
-	to=$(OUTPUT)/rawTruth.cub \
-	SAMPLES=200 LINES=200 BANDS=4 SKIP=23040 \
-	BITTYPE=REAL BYTEORDER=MSB \
-	> /dev/null;
+	  to=$(OUTPUT)/fitsTruth.cub > /dev/null;
+	catlab from=$(OUTPUT)/fitsTruth.cub \
+		to=$(OUTPUT)/fitsTruth.pvl > /dev/null;
--- a/isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
+++ b/isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
@@ -10,9 +10,9 @@ commands:
 	  to=$(OUTPUT)/mdis2isisTruth.pvl> /dev/null
 	$(APPNAME) from=$(INPUT)/EN0089569526M.IMG \
 	  to=$(OUTPUT)/mdis2isisLutTruth.cub > /dev/null
-	catlab from=$(OUTPUT)/mdis2isisTruth.cub \
+	catlab from=$(OUTPUT)/mdis2isisLutTruth.cub \
 	  to=$(OUTPUT)/mdis2isisLutTruth.pvl> /dev/null
 	$(APPNAME) from=$(INPUT)/EN0131771763M.IMG \
 	  to=$(OUTPUT)/mdis2isisStrangeLabelTruth.cub > /dev/null
-	catlab from=$(OUTPUT)/mdis2isisTruth.cub \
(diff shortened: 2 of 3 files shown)


isis/src/base/apps/fits2isis/tsts/default/Makefile
```diff
@@ -4,10 +4,6 @@ include $(ISISROOT)/make/isismake.tsts
 
 commands:
 	$(APPNAME) from=$(INPUT)/WFPC2u5780205r_c0fx.fits \
-	to=$(OUTPUT)/fitsTruth.cub \
-	> /dev/null;
-	raw2isis from=$(INPUT)/WFPC2u5780205r_c0fx.fits \
```
celbal, line 9: Why did you remove this raw2isis call?

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
```diff
@@ -10,9 +10,9 @@ commands:
 	  to=$(OUTPUT)/mdis2isisTruth.pvl> /dev/null
 	$(APPNAME) from=$(INPUT)/EN0089569526M.IMG \
 	  to=$(OUTPUT)/mdis2isisLutTruth.cub > /dev/null
-	catlab from=$(OUTPUT)/mdis2isisTruth.cub \
+	catlab from=$(OUTPUT)/mdis2isisLutTruth.cub \
```
celbal, line 13: Why Lut?

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
```diff
@@ -10,9 +10,9 @@ commands:
...
 	  to=$(OUTPUT)/mdis2isisLutTruth.pvl> /dev/null
 	$(APPNAME) from=$(INPUT)/EN0131771763M.IMG \
 	  to=$(OUTPUT)/mdis2isisStrangeLabelTruth.cub > /dev/null
-	catlab from=$(OUTPUT)/mdis2isisTruth.cub \
+	catlab from=$(OUTPUT)/mdis2isisStrangeLabelTruth.cub \
```
celbal, line 17: Why StrangeLabel?

isis/src/base/apps/fits2isis/tsts/default/Makefile
kaitlyndlee: We don't do anything after the raw2isis call, so there is no need for it. The only reason I could see why we would call raw2isis is to check if the two output cubes are the same.

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
kaitlyndlee: I changed it to Lut because if we used mdis2isisTruth.cub, we would be checking the label for mdis2isisTruth.cub again, since the first cat call does this, instead of the LutTruth cube we just created.

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
kaitlyndlee: I changed it to StrangeLabel because if we used mdis2isisTruth.cub, we would be checking the label for mdis2isisTruth.cub again, since the first cat call does this, instead of the StrangeLabel cube we just created.

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
celbal: What does Lut stand for though?

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
celbal: Does the cube have a strange label then?

isis/src/messenger/apps/mdis2isis/tsts/default/Makefile
kaitlyndlee: I am not sure. The pvl files were named this way before I made my changes. I renamed the cubes being passed into catlab to make sure it was testing the correct cube.

celbal: Is there a reasonable replacement for "Lut" that makes more sense in the context of this test? If not that is okay.

kaitlyndlee: Hi Makayla, nothing really stood out to me in the pvl file that would give me a new name.

celbal (approved):

PR #79, opened 2018-02-15
Removed GIF from documentation and added history entry.

### LINKED ISSUE #1667: Add/Show/Edit column for sigma values in cneteditor
---

Author Name: **Stuart Sides** (@scsides)

Original Date: 2013-03-18

---

Add/Show/Edit column for sigma values

Apriori and adjusted SP lat/lon/radius sigmas are now in the point table, but are currently read-only. Apriori sigmas will need to be editable. 

APriori surface point sigmas are now editable. This will be available in the next version of astro.

kaitlyndlee:
The ability to export as a GIF in isis2std has been removed because Qt does not support GIF exports. Fixes #1667.


ihumphrey (commented):
Changes look good!

* GIF has been removed as a FORMAT option
* GIF has been removed from the application documentation
* A good user-oriented history comment is present in application documentation

ihumphrey (approved):

PR #84, opened 2018-02-21
Reverted ciss2isis.xml and ciss2isis.cpp

### LINKED ISSUE #5342: Autoseed example algorithm PVL have wrong syntax
Some of the example definition PVL files on [here](https://github.com/DOI-USGS/ISIS3/wiki/Autoseed)  have a space in a keyword, which is not allowed.

**Suggested Edit**
Replace 
`Minimum Thickness`
with
`MinimumThickness`

TracieSucharski:
Reverted ciss2isis.xml and ciss2isis.cpp to prior to ipce merg. Fixes #5342.


ihumphrey (approved):
Looks good

PR #94, opened 2018-03-08
Update socetframesettings:  fix OSIRIS-REx; rename variables; cleanup.

drbitboy:
N.B. This work was originally done in a custome socetframesettings-only
     Github repo branch here:

     https://github.com/drbitboy/isis_socetframesettings/tree/20171116/cleanup

Corrected ORX MapCam and ORX PolyCam isiscamToSocetcam matrix; other cleanup

- Add exception for non-MapCam, non-PolyCam ORX inputs; added many comments

Change variable names so the each name accurately describes the variable

Other cleanup:

- Use consistent naming for concepts
  - Ocentric; not Ocentric or BodyFixed
- Use SpiceDouble instead of double
- Use macros to initialize vectors and matrices
- Use twovec_c to calculate matrix instead of combinations of sin/cos-Lat/Lon
- Remove whitespace from blank lines
- Improve comments
- Remove else-if from assignments to isisFocalPlane2SocetPlate matrix; check
  that result is a rotation matrix
- Use vscl_c instead of loop to convert km to meters
- Eliminate radii variable
- Fix typo:  call to focalMap->DetectorLineOrigin() should be
  a call to focalMap->DetectorSampleOrigin()
- Add dummy function to to eliminate warning

### DIFF
--- a/isis/src/socet/apps/socetframesettings/socetframesettings.cpp
+++ b/isis/src/socet/apps/socetframesettings/socetframesettings.cpp
@@ -30,10 +30,10 @@
 using namespace std;
 using namespace Isis;
 
-void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
-                  double ographicCamPos[3], double omegaPhiKappa[3],
-                  double isisFocalPlane2SocetPlateTranspose[3][3]);
-      
+void getCamPosOPK(Spice &spice, QString spacecraftName, SpiceDouble et, Camera *cam,
+                  SpiceDouble ographicCamPos[3], SpiceDouble omegaPhiKappa[3],
+                  SpiceDouble isisFocalPlane2SocetPlateTranspose[3][3]);
+
 void IsisMain() {
 
   // Use a regular Process
@@ -76,9 +76,9 @@ void IsisMain() {
   // Set the image at the boresight pixel to get the ephemeris time and SPICE data at that image
   // location
   double detectorSampleOrigin = focalMap->DetectorSampleOrigin();
-  double detectorLineOrigin = focalMap->DetectorSampleOrigin();
+  double detectorLineOrigin = focalMap->DetectorLineOrigin();
   cam->SetImage(detectorSampleOrigin, detectorLineOrigin);
-  double et = cam->time().Et();
+  SpiceDouble et = cam->time().Et();
 
   Spice spice(*input);
   spice.setTime(et);
@@ -94,15 +94,18 @@ void IsisMain() {
     instrumentId = (QString) orig["InstrumentId"];
     spacecraftName = (QString) orig["SpacecraftName"];
   }
-     
+
+#define VEC3_ZEROS { 0.0, 0.0, 0.0 }
+#define MTX3x3_ZEROS { VEC3_ZEROS, VEC3_ZEROS, VEC3_ZEROS }
+
   // Get sensor position and orientation (opk) angles
-  double ographicCamPos[3] = {0.0, 0.0, 0.0};
-  double omegaPhiKappa[3] = {0.0, 0.0, 0.0};
-  double isisFocalPlane2SocetPlateTranspose[3][3] =
-      {{0.0, 0.0, 0.0}, {0.0, 0.0, 0.0}, {0.0, 0.0, 0.0}};
+  SpiceDouble ographicCamPos[3] = VEC3_ZEROS;
+  SpiceDouble omegaPhiKappa[3] = VEC3_ZEROS;
+  SpiceDouble isisFocalPlane2SocetPlateTranspose[3][3] = MTX3x3_ZEROS;
+
   getCamPosOPK(spice, spacecraftName, et, cam, ographicCamPos,
                omegaPhiKappa,isisFocalPlane2SocetPlateTranspose);
-  
+
   // Determine the SOCET Set camera calibration file
   QString socetCamFile = socetCameraCalibrationPath;
 
@@ -122,7 +125,7 @@ void IsisMain() {
       socetCamFile += "VIK2B.cam";
     }
   }
-  
+
   //----------------------------------------.-------------
   //TO DO: Uncomment these lines when MEX SRC is supported
   //----------------------------------------.-------------
@@ -166,7 +169,7 @@ void IsisMain() {
     PvlGroup bandBin = cube.label()->findGroup("BandBin", Pvl::Traverse);
     QString filter = (QString) bandBin["FilterName"];
     filter.replace("/", "_");
-    
+
     socetCamFile += "Cassini_ISSNA_";
     socetCamFile += filter;
     socetCamFile += ".cam";
@@ -188,8 +191,13 @@ void IsisMain() {
     if (instrumentId == "MapCam") {
       socetCamFile += "OCAMS_MapCam.cam";
     }
-    else {//polycam or samcam
+    else if (instrumentId == "PolyCam") {
       socetCamFile += "OCAMS_PolyCam.cam";
+    }
+    else {
+      QString msg = QString("The ISIS to SOCET Set translation of input image [%1] is currently "
+                            "not supported for instrument [%2].").arg(from).arg(instrumentId);
+      throw IException(IException::User, msg, _FILEINFO_);
     }
   }
   // Throw exception for unsupported camera
@@ -248,10 +256,10 @@ void IsisMain() {
   }
   double sizeX = numSamples / pixelSize;
   double sizeY = numLines / pixelSize;
-  
+
   // Make sure the Socet Set project name has the .prj extension
   if (socetProject.endsWith(".prj", Qt::CaseInsensitive) == false)  socetProject += ".prj";
-  
+
   // Find cube base name w/o extensions & establish the Socet Set support file name 
   // Note: I'm using the QFileInfo class because the baseName method in the ISIS
   // FileName class only strips the last extension, and we need the core name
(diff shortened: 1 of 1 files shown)


isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -94,15 +94,18 @@ void IsisMain() {
     instrumentId = (QString) orig["InstrumentId"];
     spacecraftName = (QString) orig["SpacecraftName"];
   }
-     
+
+#define VEC3_ZEROS { 0.0, 0.0, 0.0 }
+#define MTX3x3_ZEROS { VEC3_ZEROS, VEC3_ZEROS, VEC3_ZEROS }
```
jessemapel, line 99: We try not to use macros unless absolutely necessary. Please switch back to just using initializer lists for zero vectors and matrices.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -422,24 +430,228 @@ void IsisMain() {
...
-void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
-                  double ographicCamPos[3], double omegaPhiKappa[3],
-                  double isisFocalPlane2SocetPlateTranspose[3][3]) {
-
-  // Initialize the isisFocalPlane2SocetPlate matrix based on mission.
+////////////////////////////////////////////////////////////////////////
+
+// OVERVIEW
+
+// getCamPosOPK converts the geometry contained in ISIS CUB labels, and
```
jessemapel, line 437: ISIS 3 documentation refers to files with the .cub extension as cubes.  Please change all references to CUB to cube.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -422,24 +430,228 @@ void IsisMain() {
...
+//
+//    _ITRANSS is a triplet:  the 2nd value is the sample pixel/mm wrt +X;
+//                            the 3rd value is the sample pixel/mm wrt +Y.
+//
+//    _ITRANSL is a triplet:  the 2nd value is the line pixel/mm wrt +X;
+//                            the 3rd value is the line pixel/mm wrt +Y.
+//
+//
+//    The first values in _ITRANSS and _ITRANSL may be ignored as long as
+//    they are zero; refer to the ISIS documentation if they are otherwise.
```
jessemapel, line 509: The ITransS and ITransL vectors define an affine transformation. So, the first value is a constant offset. I think it's clearer to describe these as affine transformations from ISIS focal plane x,y to detector pixels.

It should also be noted that detector pixels and cube pixels are not the same. Detector pixels reference a mission specific detector origin and are un-binned. Cube pixels are relative to the first sample, line as (1,1) and may be binned in the sample and/or line direction.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -422,24 +430,228 @@ void IsisMain() {
...
+//     in SPICE.  Planetographic latitude and height are normal and
+//     relative, repsectively, to the surface of a target modeled as a
+//     spheroid that is a volume of revolution.  Only the radius at the
+//     intersection of the prime meridian and the equator, and the polar
+//     radius, are used; the intermediate radius, at the intersection of
+//     meridians +/-90degrees and the equator, of the tri-axial ellipsoid
+//     model is not used.
+//
+//     N.B. Planetographic coordinates are used ***ONLY*** for calculating
+//          the camera position; they are ***NOT*** used in not used in
```
jessemapel, line 578: This looks like a minor typo. "they are ***NOT*** used in ..."

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -471,110 +683,284 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
...
+
+  if (spacecraftName == "OSIRIS-REX") {
+    /* MapCam and PolyCam ISIS-to-SS Matrix swaps X and Y, inverts Z */
+    isisFocalPlane2SocetPlate[1][0] =  1.0;  // +Xisis => +Yss
+    isisFocalPlane2SocetPlate[0][1] =  1.0;  // +Yisis => +Xss
+    isisFocalPlane2SocetPlate[2][2] = -1.0;  // +Zisis => -Zss
+  }
+
+  // Confirm that matrix is now a rotation matrix
+  if (!isrot_c(isisFocalPlane2SocetPlate, 1e-10, 1e-10)) {
```
jessemapel, line 861: Checking if the rotation matrix is valid as a proxy for checking if the instrument is supported seems obtuse.  I'd much rather see an if/else or switch setup used here with the default/else case being this exception.  Then, a separate check can be done to verify that the rotation matrix is valid.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -471,110 +683,284 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
...
+    QString msg = QString("The ISIS to SOCET Set translation of input image is currently "
+                          "not supported for instrument [%1].").arg(spacecraftName);
+    throw IException(IException::User, msg, _FILEINFO_);
+  }
+
+  //////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////
+  // End of section setting isisFocalPlane2SocetPlate
+  //////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////
```
jessemapel, line 871: If this section is so large that the end of it needs to be indicated, it should be moved to its own method.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -471,110 +683,284 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
...
   Distance dRadii[3];
   spice.radii(dRadii);
-  double radii[3] = {0.0, 0.0, 0.0};
-  radii[0] = dRadii[0].meters();
-  radii[1] = dRadii[1].meters();
-  radii[2] = dRadii[2].meters();
 
-  // Calculate ographic coordinates of spacecraft position vector
+  // Calculate ographic coordinates of spacecraft position vector, in meters
+  SpiceDouble equatorRadius_m = dRadii[0].meters();
```
jessemapel, line 908: ISIS standards do not use underscore in local variable names. equatorialRadius_m should be changed to something like equatorialRadiusMeters.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -471,110 +683,284 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
...
+  SpiceDouble lsrToOcentricRotationMatrix[3][3] = MTX3x3_ZEROS;
+  SpiceDouble ocentricToLsrRotationMatrix[3][3] = MTX3x3_ZEROS;
+
+  // LSR:  Calculate matrix to rotate vectors from LSR frame to
+  //       Planetocentric Body-Fixed frame
+
+  twovec_c(instrumentPosition, 3, uvPlusZ, 2, ocentricToLsrRotationMatrix);
+  xpose_c(ocentricToLsrRotationMatrix, lsrToOcentricRotationMatrix);
+
+  /* OLD:
```
jessemapel, line 932: The old code should be removed to avoid clutter

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -471,110 +683,284 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
...
+  mxmt_c (isisFocalPlane2SocetPlate, isisFocalPlaneToOcentricRotationMatrix,
+          ocentricGroundToSocetPlateRotationMatrix);
+
+  // Populate the ocentric-to-LSR rotation matrix; it is a function of
+  // camera position only
+  SpiceDouble lsrToOcentricRotationMatrix[3][3] = MTX3x3_ZEROS;
+  SpiceDouble ocentricToLsrRotationMatrix[3][3] = MTX3x3_ZEROS;
+
+  // LSR:  Calculate matrix to rotate vectors from LSR frame to
+  //       Planetocentric Body-Fixed frame
```
jessemapel, line 927: Why is there "LSR:" at the start of this comment?

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -593,7 +979,8 @@ void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
 
   // Return the transpose of the isisFocalPlane2SocetPlate matrix for the FrameOffAxis sensor model
   xpose_c(isisFocalPlane2SocetPlate, isisFocalPlane2SocetPlateTranspose);
-  
+
   return;
 }
+int noop_intreturn() { return (int) IVALID_MAX4; } // Eliminate compiler warning about -Wunused-variable
```
jessemapel, line 985: Why is this here? Adding a no-op method to eliminate a compiler warning seems odd.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
```diff
@@ -422,24 +430,228 @@ void IsisMain() {
...
+//      +zLSR => Vector from target body cent to camera position. N.B. this
+//               vector is planetocentric and different from the
+//               plantographic (geodetic) coordinates described in (1) above.
+//
+//      +xLSR => Direction of east longitude at sub-camera point, equal to
+//               cross product of planetocentric North polar axis (+zBF)
+//               vector with +zLSR vector (above).
+//
+//      +yLSR => In same half-plane of xzLSR plane as +zBF (north, or
+//               positive rotation, pole)
```
jessemapel, line 613: I'd like to see this whole section of comments converted to doxygen style function documentation. We don't distribute doxygen documentation for functions in application code, so I would be okay with just converting to a block comment instead of single line comments if going completely to doxygen would be a significant amount of work.

jessemapel (changes requested):
Comments on the changes in general:
1. Please document your changes in the application XML file under the history element.
2. Do you have test data for this? We have some test images I could write test using, but I would need help verifying the output.

jessemapel:
Largely for reference when this gets merged in:
Redmine ticket #5235
https://fixit.wr.usgs.gov/fixit/issues/5235

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: Perhaps an earlier version listed the individual transformation steps like this.

It is not needed now.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: Well, I guess I am an odd person then (I've been called worse;-); I don't like noise that needs to be ignored when compiling.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: yup; it is only there for the review.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: agreed.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: Or just moved to the end of the file with a note referencing it here?

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: yup.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: sure.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: okay.

The only purpose here is to show the sign of the second value, so if it ever changes then this code can be updated; the rest is window dressing.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: where is the typo?

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
drbitboy: go for it!

drbitboy:
HI Jesse, Thanks for the review.  I am pretty busy.  I am more than happy to hand this branch over to USGS; do you think someone from your shop could make those cosmetic changes?  Best regards, Brian Carcich

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
jessemapel: It looks like "not used in" is repeated at the end of this line.

isis/src/socet/apps/socetframesettings/socetframesettings.cpp
jessemapel: I'm mainly concerned about suppressing a warning that is signalling something we should worry about. I'm going to remove this and see what the exact warning is. If IVALID_MAX4 is being defined somewhere and then not used, I don't think this is the correct place to address it.

jessemapel:
@drbitboy Brian, I am not sure if we have people available to do that right now. I've reached out to those in charge of the project/planning internally to see if that is possible. Some of those people are at LPSC right now, so we may not get a response back today.

drbitboy:
I might be able to get to it over the weekend.

Also, none of these fixes affect the code, so it is still a huge improvement over what was there before.  Do we want the perfect to be the enemy of the good?

drbitboy:
Also, I apologize for the typo of your name (now fixed).

jessemapel:
@drbitboy Because this is coming from a mission we have a work agreement with I'm going to be doing clean up on this PR and bringing it into the code base. I still need test data from you though. Can you please send me an ISIS cube, command line, and output socet set file that exercise these changes? You can upload that to the redmine ticket or e-mail it directly to me.

jessemapel:
@drbitboy Hi Brian, any word on test data for this?

jessemapel:
Work on this is continuing under #476

PR #100, opened 2018-03-14
Replaced cache calculations with Camera::LoadCache() call. Fixes #5353.

AgoinsUSGS:
VimsCamera was causing spiceinit to fail when ingesting cubes with TargetName == Sky.
The superclass Camera has a method called LoadCache() which accounts for this TargetName. 
Replacing the calculations with this call allows for cubes with "TargetName == Sky" to be ingested.


twilson271828 (commented):
Looks good!  I'm glad this was an easy fix.  Now that it works, you should delete the code I suggested you comment out since it's not needed.  Also, make sure to run app/unit/cat tests and place the output in /work/projects/m05353

twilson271828 (approved):
Looks good.  Just make sure to delete the commented out code.

PR #103, opened 2018-03-15
Removed the search bar and links that are not working on the ISIS website. Fixes #5185.

kaitlyndlee:
On the ISIS website, the search bar and links to help and advanced were not working. For the time being, it is better to completely remove the search bar and links. I commented them out just in case somebody wanted to go back and edit the code later. Fixes #5185.

### DIFF
--- a/isis/src/docsys/build/menu.xsl
+++ b/isis/src/docsys/build/menu.xsl
@@ -81,11 +81,11 @@ Deborah Lee Soltesz
     </div>
 
 
-    <hr/>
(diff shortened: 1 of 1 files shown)


celbal:
Why did you decide to comment these out rather than deleting them?

kaitlyndlee:
Stuart told me it would be better to comment it out, so that somebody in the future could add the search bar back with correct links, if they wanted to.

celbal (approved):

PR #108, opened 2018-03-21
Old build system working alongside cmake

jessemapel:
The top level and 3rdParty Makefiles got removed when it was planned that cmake would just replace the old system. Adding CMakeLists.txt to the src Makefile filter allows for building and testing with the old make system. Hyper can build but not test for right now.


jessemapel:
[The cmake system needs to be able to work with the old make system (for now)](https://trello.com/c/zLVW7Vzr/98-the-cmake-system-needs-to-be-able-to-work-with-the-old-make-system-for-now)

celbal (commented):
Have you tested making cmake with this build system as well?

jessemapel:
@makaylas I was able to build with cmake on fedora25 without conflict.

Kelvinrr:
Since we decided that the CMake system is no longer going to be backwards compatible with v00X libraries, this PR isn't needed anymore.

PR #120, opened 2018-03-29
Adding support for libisis python bindings

Kelvinrr:
~~**DO NOT MERGE YET, STILL NEEDS TESTING**~~
seems to work in linux
This adds a foundation for adding python wrappers to libisis. Also includes minor tweaks.


Kelvinrr:
@jlaura not exactly sure what you are asking. This is being PRed into the cmake branch since it hasn't been merged with dev yet.

jlaura:
@Kelvinrr This looks good here w.r.t. wrapper and Python bindings.  @cneubauerUSGS and/or @jessemapel can you all give a look at the CMake specific stuff.  I have not been involved enough to review that.  Thanks!

jlaura (commented):
:+1:

Kelvinrr:
still has the gcc issue of only supporting gcc 4 on linux. but that's another problem for another day.

jessemapel:
I just noticed I deleted my comment before sending it earlier.  I looked over the cmake changes and they all seem fine to me.

PR #128, opened 2018-04-06
Moved method documentation of FileName to the cpp file from the h file. Fixes #5230.

kaitlyndlee:
The method documentation was inside of the h file, instead of the cpp file. I also made some changes to the documentation to make sure everything was consistent.

### DIFF
--- a/isis/src/base/objs/FileName/FileName.cpp
+++ b/isis/src/base/objs/FileName/FileName.cpp
@@ -40,41 +40,99 @@ using namespace std;
 
 namespace Isis {
 
+  /**
+   * Constructs an empty FileName object.
+   */
   FileName::FileName() {
     m_d = new Data;
   }
 
-
+  /**
+   * Constructs a FileName object using a char pointer as a file name.
+   *
+   * @param file char pointer representing new filename
+   */
   FileName::FileName(const char *file) {
     m_d = new Data;
     m_d->setOriginal(file);
   }
 
-
+  /**
+   * Constructs a FileName object using a QString as a file name.
+   *
+   * @param file Qstring representing new filename
+   */
   FileName::FileName(const QString &file) {
     m_d = new Data;
     m_d->setOriginal(file);
   }
 
-
+  /**
+   * Constructs a copy of a FileName object using another FileName object.
+   *
+   * @param &other FileName object to copy.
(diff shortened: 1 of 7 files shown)


isis/src/base/objs/FileName/FileName.cpp
```diff
@@ -40,41 +40,99 @@ using namespace std;
...
+   */
   FileName::FileName() {
     m_d = new Data;
   }
 
-
+  /**
+   * Constructs a FileName object using a char pointer as a file name.
+   *
+   * @param *fileName char pointer representing new filename
```
celbal: This should be @param file

isis/src/base/objs/FileName/FileName.cpp
```diff
@@ -40,41 +40,99 @@ using namespace std;
...
   FileName::FileName(const char *file) {
     m_d = new Data;
     m_d->setOriginal(file);
   }
 
-
+  /**
+   * Constructs a FileName object using a QString as a file name.
+   *
+   * @param &fileName Qstring representing new filename
```
celbal: This should be @param file

isis/src/base/objs/FileName/FileName.cpp
```diff
@@ -87,32 +145,98 @@ namespace Isis {
...
   QString FileName::original() const {
     return m_d->original(true);
   }
 
-
+  /**
+   * Adds a new extension to the file name. If the current extension is the same as the
+   * new extension it will return an unchanged FileName object.
+   *
+   * @param &extension The new file extension to be added at the end of the file name after all
```
celbal: This should be @param newExtension

isis/src/base/objs/FileName/FileName.cpp
```diff
@@ -142,7 +270,13 @@ namespace Isis {
     return result;
   }
 
-
+  /**
+   * Sets all current file extensions to a new extension in the file name.
+   *
+   * @param &extension The new file extension to replace any current file extensions with
```
celbal: This should be @param newExtension

celbal:
Did you accidentally make commits to more files than FileName.cpp and FileName.h? Those are the only two files that should be changed.

PR #132, opened 2018-04-06
Fixed bug on install, working environment.yml

Kelvinrr:
- Fixes issue https://github.com/USGS-Astrogeology/isis3_cmake/issues/35
- updated environment.yml to work so that the platform specific spec files are not necessary (requires latest version of Anaconda). Opens the possibility for a ISIS3 recipe for conda. 
- Fixed bug where CXX flags were not set in the right order
- since CMAKE tends to handle rpaths just fine, we no longer need to use `finalizeInstalledOsxRpaths.py` so the calls have been removed


jlaura:
:+1:

PR #146, opened 2018-04-19
Fixed issue with apps core dumping when ISISROOT not set. Fixes #5360.

chrisryancombs:


isis/src/qisis/apps/ipce/ipce.cpp
```diff
@@ -35,7 +35,10 @@ using namespace std;
 using namespace Isis;
 
 int main(int argc, char *argv[]) {
-
+  if (getenv("ISISROOT") == NULL || QString(getenv("ISISROOT")) == "") {
+    std::cerr << "Please set ISISROOT before running any Isis applications" << std::endl;
```
celbal, line 39: Can this message be pulled into a different, common class that is shared with all of the qisis apps? That way there isn't duplicated code

chrisryancombs: The common class that this would be added to is QIsisApplication. There are a couple issues with putting the ISISROOT check in its constructor. 

First, qtie and ipce don't use QIsisApplication, and instead use QApplications. This means that 2/6 of the qisis applications would not be able to use the common check. 

Also, qview does a block of logic before creating the application, in which it handles sending a cube to an already open instance of qview. This block can error out due to ISISROOT not being set. A possible workaround for this would be to check twice, once in qview's main method, and a second time in QIsisApplication's constructor.

At the moment, the only solution I can see with less duplicated is the QIsisApplication constructor check, but that only eliminates 2/6 of the duplicates, and makes the applications' checks less uniform. I can implement this if its not a large concern.

celbal (approved):

PR #163, opened 2018-05-09
Minor fixes to get tgocassismos tests running and added an intial app…

krlberry:
… test


celbal (approved):
This will need to be cleaned up later (ie remove the commented out code)

PR #169, opened 2018-05-15
Ipce-merge dev into ipce 

TracieSucharski:


isis/src/qisis/objs/ImageFileListWidget/ImageFileListWidget.cpp
```diff
@@ -621,10 +623,11 @@ namespace Isis {
    */
   void ImageFileListWidget::save(QXmlStreamWriter &stream, Project *project,
                                  FileName newProjectRoot) const {
+
     stream.writeStartElement("imageFileList");
 
     // Write QSettings
-    stream.writeStartElement("widgetGeometry");
+//  stream.writeStartElement("widgetGeometry");
```
celbal, line 630: Why is this uncommented but the rest of this block is not? Can the rest of this be deleted?

TracieSucharski (commented):
We're hopefully going to use some of this commented code in the upcoming sprint when we convert views to docked widgets, so decided to leave commented rather than re-creating in a few weeks.

PR #176, opened 2018-05-17
Added PRODUCTID parameter to tgocassisrdrgen to allow users to update the product id

kaitlyndlee:
Users will be able to specify the product id value they want; if they want to update the value. If the input cube does not have the ProductId keyword, we create one, then set the value. Added onto the default test because we are going to use the same input.


krlberry:
@kdl222 Could you put a comment in here when this is ready to merge?

kaitlyndlee:
Still working on this.

kaitlyndlee:
Is all good now. Ready to be merged.

PR #180, opened 2018-05-18
Updated tgocassisrdrgen to check if a cube has a Mosaic group

kaitlyndlee:
rdrgen will now check if a cube as a Mosaic group if the Instrument group does not exist. Updated default test to remove ISIS version and modification_date.

### DIFF
--- a/isis/src/tgo/apps/tgocassisrdrgen/tsts/default/Makefile
+++ b/isis/src/tgo/apps/tgocassisrdrgen/tsts/default/Makefile
@@ -1,3 +1,10 @@
+# Verify that we can export a non-projected and non-mosaiked cube as a TGO CaSSIS image.
+# Test that we can override the product id with the PRODUCTID parameter.
+#
+# history 2018-05-18 Kaitlyn Lee - Changed input cube with an one that has the
+# 				ObservationId. Added the removal of the modification_date and ISIS version.
+# 				Added testing for PRODUCTID parameter.
+
 APPNAME = tgocassisrdrgen
 
 include $(ISISROOT)/make/isismake.tsts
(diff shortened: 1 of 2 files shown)


kdl222:
[rdrgen needs to check if Mosaic group exists](https://trello.com/c/s8pONO02/488-rdrgen-needs-to-check-if-mosaic-group-exists)

isis/src/tgo/apps/tgocassisrdrgen/tsts/default/Makefile
```diff
@@ -6,17 +13,14 @@ commands:
...
-		 > /dev/null;
-
-	# Compare the output cube from raw2isis to the original input cube to
-	# ensure that they are identical
-	cubediff from=$(INPUT)/CAS-MCO-2016-11-22T15.45.50.984-BLU-03000-B1.cub \
-	         from2=$(OUTPUT)/CAS-MCO-2016-11-22T15.45.50.984-BLU-03000-B1.cub \
-	         to=$(OUTPUT)/cubediffresults.txt \
+			 
+	# Reingest output and make sure the cubes are the same
+	raw2isis from=$(OUTPUT)/CAS-MCO-2016-11-22T15.45.50.984-BLU-03000-B1.img \
```
krlberry, line 18: Why switch from `tgocassis2isis` to `raw2isis`? And why did you take out the cubediff?

kaitlyndlee: Originally, it was actually using raw2isis, but I changed it to tgocassis2isis with the previous pull request. However, Jeanie said that we should only be testing rdrgen in its apptest, so I changed it back. I also took out the cubediff because I copied over the input cube into the truth folder, so it should automatically do a cubediff when testing the output.

krlberry: cool, thanks for explaining! I should probably have caught that second one, perfect!

krlberry: :+1:

krlberry (approved):

PR #182, opened 2018-05-19
Adding appTest for rdrgen for projected images

SgStapleton:
Test data can be found in /work/users/sgstapleton/testData/tgocassis/mapProjected

### DIFF
--- a/isis/src/tgo/apps/tgocassisrdrgen/tsts/mapProjected/Makefile
+++ b/isis/src/tgo/apps/tgocassisrdrgen/tsts/mapProjected/Makefile
@@ -1,0 +1,13 @@
+# This tests exporting a previously projected cube. 
+#
+# @history 2018-05-17 Summer Stapleton - Original version. 
+#
+
+EXPORT = tgocassisrdrgen
+
+include $(ISISROOT)/make/isismake.tsts
+
+commands:
+
+	$(EXPORT) $(TSTARGS) from=$(INPUT)/CAS-MCO-2016-11-26T22.32.23.582-BLU-03009-B1_projected.cub \
+            to=$(OUTPUT)/CAS-MCO-2016-11-26T22.32.23.582-BLU-03009-B1_projected.img > /dev/null;


isis/src/tgo/apps/tgocassisrdrgen/tsts/mapProjected/Makefile
```diff
@@ -0,0 +1,13 @@
...
+#
+
+EXPORT = tgocassisrdrgen
+
+include $(ISISROOT)/make/isismake.tsts
+
+commands:
+
+	$(EXPORT) $(TSTARGS) from=$(INPUT)/CAS-MCO-2016-11-26T22.32.23.582-BLU-03009-B1_projected.cub \
+            to=$(OUTPUT)/CAS-MCO-2016-11-26T22.32.23.582-BLU-03009-B1_projected.img > /dev/null;
```
krlberry, line 13: You might need to add something similar to the following section of your other PR (`isis/src/tgo/tsts/mapProjectedReingested/Makefile`)

```
 $(SED) 's+\Product_Observational.*>+\Product_Observational>+' \
 +           $$label > $${label%.xml}1.txt; \
 +	  $(SED) 's+\FSW_HEADER.*>+\FSW_HEADER>+' \
 +           $${label%.xml}1.txt > $${label%.xml}2.txt; \
 +	  $(SED) 's+\PEHK_HEADER.*>+\PEHK_HEADER>+' \
 +           $${label%.xml}2.txt > $${label%.xml}3.txt; \
 +	  $(SED) 's+\Modification_Detail.*>+\Modification_Detail>+' \
```
Since (unless something has changed, which is definitely possible), when we output a tag with many attributes, we can't control the order of the attributes, so the test will end up sometimes failing when the order of the attributes differs, even though this is valid output. 

ex: ` <FSW_HEADER UID="100797368" Timestamp="2f0154313ab1aa13" SequenceCounter="9" WindowCounter="3" ExposureTimestamp="2f0154313ab1aa13"/>` is your current output for that tag, but I think it's possible that 

` <FSW_HEADER WindowCounter="3"  UID="100797368" Timestamp="2f0154313ab1aa13" SequenceCounter="9" ExposureTimestamp="2f0154313ab1aa13"/>`

could also be output, so for the test, you can use sed commands similar to the ones above to just not output all the attributes in these cases to the text files that will actually be diffed by the test

krlberry (changes requested):

PR #183, opened 2018-05-21
tgocassisrdrgen will now add a ProductId keyword to a mosaic

kaitlyndlee:
The productId can be set with the parameter or set to the ObservationId for a mosaic. Renamed rdrgen's default test to raw.


kdl222:
[param (PRODUCTID) in rdrgen to overwrite default prod id needs to work for mosaicked product](https://trello.com/c/y71IOXA7/495-param-productid-in-rdrgen-to-overwrite-default-prod-id-needs-to-work-for-mosaicked-product)

isis/src/tgo/apps/tgocassisrdrgen/tgocassisrdrgen.cpp
```diff
@@ -57,22 +57,22 @@ void IsisMain() {
   // by the user, set it to the Observation Id.
   // This is added before the translation instead of adding it to the exported xml
   // because of the ease of editing pvl vs xml.
-  PvlGroup &instrumentGroup = label->findObject("IsisCube").findGroup("Instrument");
+  PvlGroup &targetGroup = isMosaic ? label->findObject("IsisCube").findGroup("Mosaic")
```
celbal: Why did you chose to use a binary operator?

kaitlyndlee: I originally had a method called writeProductId that would take in a group name, either Instrument or Mosaic. I would either have to pass in the user interface and the input cube label as parameters or have those as global variables. I was not sure about how this was usually done, so I did it this way instead. We also can not do an if-else where we check to see if isMosaic is true and then set targetGroup to find the Mosaic and else find the Instrument group because we can not have an uninitialized reference (that is if we declare a reference above the if-else).

celbal (approved):

PR #186, opened 2018-05-22
Removed Z from end of StartTime on re-ingested cubes

SgStapleton:
Would this potentially make recent changes to iTime obsolete as any ingestion program will need to handle this individually anyway?

And should I create a redmine ticket for this one? (I wasn't sure what to reference in the history comment without a ticket number...)

### DIFF
--- a/isis/src/base/objs/iTime/iTime.cpp
+++ b/isis/src/base/objs/iTime/iTime.cpp
@@ -47,29 +47,19 @@ namespace Isis {
   }
 
   /**
-   * Constructs a iTime object and initializes it to the time from the argument.
-   *  
-   * This constructor will (temporarily) strip the Z off of the end of an input UTC-formatted time 
-   * if it is present. This time format is used by PDS4, but not presently supported by NAIF calls, 
-   * in particular str2et_c, which is used by this class.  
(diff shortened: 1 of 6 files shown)


celbal:
Your changes do make the iTime changes obsolete. Kristin and I have talked about this and we think that this is the correct place to deal with the trailing Z. Could you please delete the changes to iTime and test this again?

SgStapleton:
Done. Everything appears to be passing.

krlberry:
@SgStapleton :+1: Looks good!

krlberry:
@SgStapleton One more question -- is there any test data that will need to be updated for this once the CaSSIS branch is merged into trunk, and if so, where is the updated test data.

jcwbacker:
@kberryUSGS @SgStapleton I think we should put the data in /work/projects/tgo/testData (for app/cat tests) and when it's merged into dev, the reviewer check the data in from there.

SgStapleton:
Just removed the history comment from the reverted changes to iTime (as this never went into dev anyway).

SgStapleton:
Other than iTime truthdata, this should only really be affecting our catTests that are testing the re-ingestion of exported .xmls. I have updated the catTest I created. I will let Adam know to update his.

PR #187, opened 2018-05-23
PROG: made BundleSolutionInfo default and copy constructors and assig…

kledmundson:
…nment operator private since it is derived from QObject; modified BundleAdjust methods bundleSolveInformation and solveCholeskyBR to return raw pointers to a BundleSolutionInfo object; modified init method of JigsawDialog to not set m_BundleSolutionInfo to NULL since it no longer owns it; also modified the destructor to not delete m_BundleSolutionInfo or set it to NULL.


TracieSucharski (approved):
Looks good, I did PR.

PR #188, opened 2018-05-24
PROG: modified calls to BundleAdjust::solveCholeskyBR() in deltack.cp…

kledmundson:
…p and jigsaw.cpp to return raw pointers to a BundleSolutionInfo object


TracieSucharski (approved):
Looks good.

PR #193, opened 2018-05-29
Replace the ControlCubeGraphNode ControlNet graph with the boost graph library.

krlberry:
I removed unneeded code from `ControlNet` as part of this card (https://trello.com/c/INYtCNHS) and commented-out the related code from the unit test. 

I didn't completely remove the code from the unit test yet because I thought we might want to later replace it with something that produces similar output, but using the new boost graph.


krlberry:
The removed code broke ControlMeasure.... fixing this.

krlberry:
Okay, this may be as far as I get today. This PR has the now unneeded methods deleted and the graph and some of its functionality added (see the trello card.) It also still contains much of the old graph code because including both the old and the new code allows it to compile and test.

krlberry:
Continuing work on this today

isis/src/control/objs/ControlNet/ControlNet.h
```diff
@@ -221,6 +237,25 @@ namespace Isis {
...
+        int strength = 0;
+      };
+
+      // typedefs to help cut down on templated type bloat
+      typedef boost::adjacency_list<boost::setS, boost::listS, boost::undirectedS, Image, Connection> Network;
+      typedef Network::vertex_descriptor ImageVertex;
+      typedef Network::edge_descriptor ImageConnection;
+      typedef std::map<ImageVertex, size_t> VertexIndexMap;
+      typedef boost::associative_property_map<VertexIndexMap> VertexIndexMapAdaptor;
+      typedef Network::out_edge_iterator ConnectionIterator;
```
jessemapel: The implementation of the graph is private, so the bundled property structs and the typedefs should all be private

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -369,9 +373,16 @@ namespace Isis {
...
-      if (!cubeGraphNodes->contains(sn)) {
+       if (!cubeGraphNodes->contains(sn)) {
         cubeGraphNodes->insert(sn, new ControlCubeGraphNode(sn));
       }
+      // if the graph doesn't have the sn
+      if (!m_vertexMap.contains(sn)) {
+        ImageVertex newVertex = boost::add_vertex(m_controlGraph); 
+        m_controlGraph[newVertex].serial = sn; 
+        m_vertexMap.insert(sn, newVertex);
+  //    boost::put(indexMapAdaptor, newVertex, i); <-- what does this do?
```
jessemapel: This basically assigns numbers to the vertices in the graph. The connected components algorithm needs this map in order to work when vertices are stored in a list or set. We may want to recreate this map everytime we call the connected components method instead of trying to store it and keep it up to date.

jessemapel: See the vertex_index_map parameter https://www.boost.org/doc/libs/1_67_0/libs/graph/doc/connected_components.html

isis/src/control/objs/ControlNet/ControlNet.h
```diff
@@ -221,6 +237,25 @@ namespace Isis {
       friend class ControlPoint;
 
     public:
+      struct Image {
+        QString serial;
+        QList<ControlMeasure* > measures;
```
jessemapel: When we remove measures from the network, we'll need to be find them in these lists and then remove them. Because this list can get quite large, we probably want to use something that is faster to remove things from such as a a map or hash.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -522,6 +586,12 @@ namespace Isis {
             node->addConnection(neighborNode, point);
             neighborNode->addConnection(node, point);
           }
+
+          // new graph:
+            ImageConnection connection = boost::add_edge(m_vertexMap[serial],
+                                                         m_vertexMap[sn],
+                                                         m_controlGraph).first;
+            m_controlGraph[connection].strength++;
```
jessemapel: We also need to add the measure to the list of measures in the vertex

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1067,14 +831,16 @@ namespace Isis {
...
+    return m_vertexMap.keys(); 
+//    return cubeGraphNodes->keys();
   }
 
 
   /**
    * @returns A list of all the cube graph nodes in the network
    */
   QList< ControlCubeGraphNode * > ControlNet::GetCubeGraphNodes() {
+    // Just delete function?
```
jessemapel: Yep, kill it. With the way the graph is managed in boost, we REALLY do not want to be handing out vertex or edge descriptors because they can be invalidated very quickly.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1086,11 +852,17 @@ namespace Isis {
...
-      throw IException(IException::Programmer, msg, _FILEINFO_);
+//  if (!cubeGraphNodes->contains(serialNumber)) {
+//    IString msg = "Cube Serial Number [" + serialNumber + "] not found in "
+//        "the network";
+//    throw IException(IException::Programmer, msg, _FILEINFO_);
+//  }
+    if (!m_vertexMap.contains(serialNumber)) {
+          IString msg = "Cube Serial Number [" + serialNumber + "] not found in "
+              "the network";
+          throw IException(IException::Programmer, msg, _FILEINFO_);
```
jessemapel: We should consider changing this to just return a bool instead of throwing an exception.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1112,7 +885,16 @@ namespace Isis {
    */
   QList< ControlMeasure * > ControlNet::GetValidMeasuresInCube(QString serialNumber) {
     ValidateSerialNumber(serialNumber);
-    return (*cubeGraphNodes)[serialNumber]->getValidMeasures();
+
+    QList< ControlMeasure * > validMeasures;
+
+    QList< ControlMeasure * > measureList = m_controlGraph[m_vertexMap[serialNumber]].measures;
```
jessemapel: We should just do a call to GetMeasuresInCube to generate the initial list

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1101,6 +873,7 @@ namespace Isis {
    */
   QList< ControlMeasure * > ControlNet::GetMeasuresInCube(QString serialNumber) {
     ValidateSerialNumber(serialNumber);
+//    return m_controlGraph[m_vertexMap[serialNumber]].measures;  <--- segfaults.
```
jessemapel: I'm concerned that we may have invalidated the vertexMap somehow. Is there anywhere else that we use the vertedMap to get a vertex descriptor by serial number and then use that descriptor?

jessemapel: It'll probably take going into a debug to figure this out. Do you have a good test case where this seg faults?

jessemapel: Looking over the code, we are using the vertedMap this way elsewhere, so I'm curious if the measure list is not being managed correctly.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1:

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: I can no longer reproduce the segfault problem after rebuilding everything. I'm not sure what I changed to fix this, but the problem seems to have disappeared.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: I did change this to returning a bool -- here are some followup questions: maybe or maybe not for the standup: 

* GetMeasuresInCube -- if can't validate SN, return an empty list or throw exception? 
* GetValidMeasuresInCube -- if can't validate SN, return an empty list or throw exception? 
* DeleteMeasuresWithID -- if we can't validate the SN, do we delete no measures, or throw an exception.

isis/src/control/objs/ControlNet/ControlNet.h
krlberry: :+1: Done.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1: Done.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -817,247 +742,67 @@ namespace Isis {
...
+    VertexIndexMapAdaptor componentAdaptor(componentMap);
+    size_t numComponents = boost::connected_components(m_controlGraph, componentAdaptor,
+                                                      boost::vertex_index_map(indexMapAdaptor));
+
+//    std::cout << "Number of components:" << std::endl;
+//    std::cout << numComponents << std::endl; 
+//    std::cout << "Size of return: " << componentMap.size() << std::endl; 
+
+    // This is not efficient, but works
+    QHash<int, QList<QString> > islandsByGroup;
```
jessemapel: You know how many islands you have at this point, so you can create the output list right here.

jessemapel: This way you don't need a second loop

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -369,16 +357,20 @@ namespace Isis {
     // make sure there is a node for every measure
     for (int i = 0; i < point->GetNumMeasures(); i++) {
       QString sn = point->GetMeasure(i)->GetCubeSerialNumber();
-      if (!cubeGraphNodes->contains(sn)) {
-        cubeGraphNodes->insert(sn, new ControlCubeGraphNode(sn));
+
+      // if the graph doesn't have the sn
+      if (!m_vertexMap.contains(sn)) {
+        ImageVertex newVertex = boost::add_vertex(m_controlGraph); 
+        m_controlGraph[newVertex].serial = sn;
```
jessemapel: Based on the documentation, you can combine this line with the one above it using add_vertex(properties, graph). It's probably an incredibly small optimization though.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -436,15 +455,19 @@ namespace Isis {
     // make sure there is a node for every measure in this measure's parent
```
jessemapel: I know this is existing functionality, but I'm confused by we have to check this here. If you're adding a whole new point, then it will use pointAdded. So, this is only adding a single new measure/image.

I think we can get rid of this check entirely.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -453,12 +476,11 @@ namespace Isis {
         ControlMeasure *cm = point->GetMeasure(i);
         if (!cm->IsIgnored()) {
           QString sn = cm->GetCubeSerialNumber();
-          ControlCubeGraphNode *neighborNode = (*cubeGraphNodes)[sn];
 
-          if (neighborNode != node) {
-            node->addConnection(neighborNode, point);
-            neighborNode->addConnection(node, point);
-          }
+          ImageConnection connection = boost::add_edge(m_vertexMap[serial],
```
jessemapel: It's probably good to add a comment everywhere that we add edges about how it doesn't always add an edge, that we are exploiting the fact that using a set forces unique edges and in the case that the edge exists, this just returns it.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -552,24 +573,24 @@ namespace Isis {
...
 
-    // Remove the measure from the node.  If this caused the node to be empty,
-    // then delete the node.
-    node->removeMeasure(measure);
-    if (!node->getMeasureCount()) {
-      delete node;
-      node = NULL;
-      cubeGraphNodes->remove(serial);
+    // Decrement the edge strength & remove edge if -> 0 happens in measureIgnored
+    // If this caused the node to be empty, then delete the node.
```
jessemapel: This is still going to be implemented right? The following code does something else, but we need to do both.

jessemapel: We need the code in the measureIgnored method here too.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -817,247 +742,67 @@ namespace Isis {
...
    * @returns The total number of edges in the bi-directional graph for images
    */
   int ControlNet::getEdgeCount() const {
-    int total = 0;
-    foreach (ControlCubeGraphNode * node, *cubeGraphNodes) {
-      total += node->getAdjacentNodes().size();
-    }
+    int total = boost::num_edges(m_controlGraph); 
 
     return total;
```
jessemapel: smoosh to 1 line function, `return boost::num_edges(c_controlGraph)`

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1085,12 +822,12 @@ namespace Isis {
...
-    if (!cubeGraphNodes->contains(serialNumber)) {
-      IString msg = "Cube Serial Number [" + serialNumber + "] not found in "
-          "the network";
-      throw IException(IException::Programmer, msg, _FILEINFO_);
+  bool ControlNet::ValidateSerialNumber(QString serialNumber) const {
+
+    if (!m_vertexMap.contains(serialNumber)) {
+      return false;
     }
+    return true;
```
jessemapel: make 1 liner, `return m_vertexMap.contains(serialNumber)`

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1100,19 +837,34 @@ namespace Isis {
...
-    ValidateSerialNumber(serialNumber);
-    return (*cubeGraphNodes)[serialNumber]->getMeasures();
+    if( ValidateSerialNumber(serialNumber) ) {
+      return m_controlGraph[m_vertexMap[serialNumber]].measures.values();
+    }
+    else {
+      IString msg = "Cube Serial Number [" + serialNumber + "] not found in "
+          "the network";
+      throw IException(IException::Programmer, msg, _FILEINFO_);
+    }
```
jessemapel: This will probably generate compiler warnings about possibly not returning. Make the exception in an if and then have the regular functionality after the if block.

jessemapel: Like in FindClosest

jessemapel (changes requested):

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: That's a good point, thank you! I will add that comment.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1:

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: @jessemapel My comment was in the wrong place and did need to be updated. I think `measureIgnored` will always get called inside of `measureDeleted` if the graph needs to be updated and the edge pruning will always take place in `measureIgnored`.... but this seems like an odd design, so I should probably look back at what _exactly_ happens to the old graph in these two functions.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1: Thanks!

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1:

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1:

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1: Okay, added this everywhere an add is done.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1: Thank you. I even knew this was the wrong way to do it, I was just blanking on what the right way was.

isis/src/control/objs/ControlNet/ControlNet.cpp
jessemapel: ah I didn't notice that. I agree that is rather odd functionality.

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: :+1: Just to confirm, I did take this out.

isis/src/control/objs/ControlCubeGraphNode/ControlCubeGraphNode.cpp
```diff
@@ -102,7 +102,7 @@ namespace Isis {
       ASSERT(0);
     }
 
-    measure->associatedCSN = NULL;
+//    measure->associatedCSN = NULL;
```
jessemapel, line 105: Should just delete these lines, not comment them

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -160,15 +152,8 @@ namespace Isis {
...
-      QHashIterator< QString, ControlCubeGraphNode * > i(*cubeGraphNodes);
-      while (i.hasNext()) {
-        i.next();
-        delete(*cubeGraphNodes)[i.key()];
-        (*cubeGraphNodes)[i.key()] = NULL;
-      }
-      cubeGraphNodes->clear();
-    }
+    m_controlGraph.clear();
+    m_vertexMap.clear();
```
jessemapel: out of an abundance of caution, we should swap these. There are null pointers in the map for a brief moment with this order.

jessemapel (approved):
Two tiny comments. I think this is solid right now. Once testing is done, it's good to go.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -401,29 +392,75 @@ namespace Isis {
...
-   * Updates the ControlCubeGraphNode for the measure's serial number to
-   * reflect the addition.  If there is currently no ControlCubeGraphNode for
-   * this measure's serial, then a new ControlCubeGraphNode is created with
+   * Used for verifying graph intergrity
+   *
+   * @returns A string representation of the ControlNet graph
+   */
+  QString ControlNet::GraphToString() const {
+    QString graphString; 
+    typedef boost::graph_traits<Network>::edge_iterator edge_iter;
```
jessemapel: I just realized, you should replace the typedef in the header for connectionIterator with this typedef.

isis/src/control/objs/ControlCubeGraphNode/ControlCubeGraphNode.cpp
jessemapel: Nevermind! This file is going away soon

krlberry:
@jessemapel Could you remind me what we decided was necessary for testing?

jessemapel:
run new unit test with old code, compare to new code

krlberry:
@jessemapel The only differences I see in a diff of the new truthdata vs. the output of the new unitTest run with the old code are things I would expect: 

* the different string output format
* serial numbers are no longer removed from the list
* the edge count for a single connection between 2 nodes is 1 instead of 2. (The old graph structure had double the edges since there was both (soruce, target) and (target, source).)

PR #195, opened 2018-05-30
Ipce docks 

TracieSucharski:
Changes made to following classes for refactored AbstractProjectItemView to inherit from QMainWindow which will allow the ipce interface to change from MDI to the views becoming docked windows within the ipce main window.  These classes need further cleanup.
AbstractProjectItemView
IpceMainWindow
Directory
CubeDnView


SgStapleton:
Looks good.

PR #203, opened 2018-06-01
Added health monitor gui, network vitals, app to view monitor

AgoinsUSGS:
Created NetworkVitals object that contains Control Network health information.
Created ControlHealthMonitor class to display all information from the NetworkVitals object.
Created temporary application "healthmonitor" to display these two classes.


AgoinsUSGS:
Hold off on merging this for now. Gotta push something up before merge

AgoinsUSGS:
This can be reviewed and merged now.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // Replace this with graph call!!!$!@$!@$!@$#@%#@$%#@
+    return true;
+  }
+
+  int ControlNetVitals::numIslands() {
+    // replace this with graph call!#@$!#%#@%*($#)
+    return 1;
+  }
+
+  QList<QString> ControlNetVitals::getIslands() {
```
jessemapel: This will need to return a QList of QLists of QStrings, the internal lists are the islands.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  int ControlNetVitals::numLockedPoints() {
+    return m_controlNet->GetNumEditLockPoints();
+  }
+
+  // REFACTOR
+  int ControlNetVitals::numFixedPoints() {
+    int count = 0;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->GetPointTypeString() == "Fixed") count++;
```
jessemapel: You should use GetType and then compare against ControlPoint::Fixed instead of getting it as a string.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+    return count;
+  }
+
+  // REFACTOR
+  int ControlNetVitals::numConstraintedPoints() {
+    int count = 0;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->GetPointTypeString() == "Constrained") count++;
```
jessemapel: You should use GetType and then compare against ControlPoint::Constrained instead of getting it as a string.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+    return count;
+  }
+
+  // REFACTOR
+  int ControlNetVitals::numFreePoints() {
+    int count = 0;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
```
jessemapel, line 84: You should use GetType and then compare against ControlPoint::Free instead of getting it as a string.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+      if (m_controlNet->GetMeasuresInCube(serial).size() < num) count++;
+    }
+    return count;
+  }
+  // REFACTOR
+  int ControlNetVitals::numImagesBelowHullTolerance(int tolerance) {
+    return 1;
+  }
+
+  QList<QString> ControlNetVitals::getCubeSerials() {
```
jessemapel, line 126: Is there a reason why you would use this instead of the ControlNet's method?

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+    return m_controlNet->GetCubeSerials();
+  }
+
+  QList<ControlPoint*> ControlNetVitals::getAllPoints() {
```
jessemapel, line 130: Same as getCubeSerials, why not just call the ControlNet's method. Do you see a usage where the caller doesn't have access to the network?

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->IsEditLocked()) lockedPoints.append(point);
+    }
+    return lockedPoints;
+  }
+
+  QList<ControlPoint*> ControlNetVitals::getFixedPoints() {
+    QList<ControlPoint*> fixedPoints;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->GetPointTypeString() == "Fixed") fixedPoints.append(point);
```
jessemapel: Same as the counting methods, use the actual enumeration instead of string comparisons.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  int ControlNetVitals::numPoints() {
+    return m_controlNet->GetNumPoints();
+  }
+
+  // REFACTOR
+  int ControlNetVitals::numIgnoredPoints() {
+    int count = 0;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->IsIgnored()) count++;
```
jessemapel: Don't use 1 line if statements. They are clean, but they obscure the fact that there's a check inside of the foreach loop.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,245 @@
...
+  QString ControlNetVitals::getNetworkId() {
+    return m_controlNet->GetNetworkId();
+  }
+
+  //
+  // ImageVitals ControlNetVitals::getImageVitals(QString serial) {
+  //   return NULL;
+  // }
+
+  void ControlNetVitals::validate() {
```
jessemapel, line 221: This method looks like GUI functionality. Is there a reason why it's not in the GUI?

jessemapel (changes requested):

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
AgoinsUSGS: The GUI doesn't interface with a ControlNet directly to grab the data it's representing. It interfaces with a ControlNetVital's object only, so we use a wrapper around the network in ControlNetVitals to grab data like this.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
AgoinsUSGS: I felt like the overall status of a network was relevant information to store in the NetworkVitals object to reflect the current sitting status of it's Control Network. This slot is connected to the netModified() slot of the ControlNet to recalculate the overall status of the network. If we decide that the status and health of a network should be determined by the GUI then I can absolutely move it there

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
jessemapel: :+1:

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
jessemapel: Sounds good

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -0,0 +1,259 @@
...
+      }
+    }
+    return count;
+  }
+
+  // REFACTOR
+  int ControlNetVitals::numFreePoints() {
+    int count = 0;
+    foreach(ControlPoint* point, m_controlNet->GetPoints()) {
+      if (point->GetType() == ControlPoint::Constrained) {
```
jessemapel: This looks like a typo, probably should compare against ControlPoint::Free

jessemapel (changes requested):
1 minor typo to fix still

AgoinsUSGS:
Just noticed, I had been using "Constrainted" instead of "Constrained" as well. fixed that too.

jessemapel (approved):

PR #204, opened 2018-06-01
Added comment explanation for removed setupExecution method.

chrisryancombs:

PR #214, opened 2018-06-07
Updated JigsawSetupDialog's general tab according to new designs

chrisryancombs:


ihumphrey:
@chrisryancombs @kledmundson 

**High-level Review:**

The validators for the line edits are a bit confusing.
When you first open JigsawSetup, the two line edits for "Latitude" and "Longitude" are empty and have a white background.
If you put a huge number into "Latitude", it turns red (suggesting an invalid value). The "OK" button at the bottom of the dialog becomes inactive and unclickable. 
If you remove the number so that the dialog is empty again, it is still red, preventing the user from clicking "OK".

The above behavior occurs with all of the line edits for the "General" tab, except for the "Output Control Net" line edit. This could be confusing if the user sets a value and changes it to an invalid value, and they don't know why it's invalid (i.e. they don't know the valid ranges).

---

I'm not sure what the best solution is for this, but here are some ideas:
* Have a tooltip (a hover text) that shows the allowable range for the line edits.
* Have a tooltip that shows the default value for the line edits.
* Have default values for some of the line edits:
    * Does it make sense for all the the line edits to have default values? For example, is the default value of the "Latitude" line edit an unset value or 0.0 (or something like that)?
* Consider changing the type of the line edits:
    * Maximum iterations could be a QSpinBox
    * Latitude / Longitude could be QDoubleSpinBox
    * Outlier Rejection Sigma Multiplier could be a QDoubleSpinBox
    * Maximum Likelihood Estimations could be QDoubleSpinBox
    * etc. since QSpinBox can enforce ranges easily

isis/src/qisis/objs/Directory/JigsawWorkOrder.cpp
```diff
@@ -92,6 +92,20 @@ namespace Isis {
   }
```
ihumphrey: Make sure to pull down upstream/ipceBundleWindow 's changes into here, PR #204 has been merged into ipceBundleWindow.

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.cpp
```diff
@@ -201,131 +225,156 @@ namespace Isis {
...
     m_ui->outlierRejectionMultiplierLineEdit->setText(toString(settings->outlierRejectionMultiplier()));
     m_ui->sigma0ThresholdLineEdit->setText(toString(settings->convergenceCriteriaThreshold()));
     m_ui->maximumIterationsLineEdit->setText(toString(settings->convergenceCriteriaMaximumIterations()));
 
 
-    m_ui->positionComboBox->setCurrentIndex(observationSolveSettings.instrumentPositionSolveOption());
+    // m_ui->positionComboBox->setCurrentIndex(observationSolveSettings.instrumentPositionSolveOption());
     m_ui->hermiteSplineCheckBox->setChecked(observationSolveSettings.solvePositionOverHermite());
     m_ui->spkDegreeSpinBox->setValue(observationSolveSettings.spkDegree());
+    std::cout<<"observationSolveSettings.spkDegree(): " << observationSolveSettings.spkDegree() <<std::endl;
```
ihumphrey: Make sure to remove this cout after you are finished testing.

ihumphrey (commented):
I'll re-review this after the high-level comment on the PR is looked at.

twilson271828 (commented):
A few more high level comments:  
*  The tooltips for Latitude/Longitude display valid ranges of 0:360 for both.  Shouldn't the range be [-90,90] for latitude, [-180,190] for longitude?  Or if a different convention is expected, it should be stated in the tooltip.  I'm not a big fan of spinner boxes in GUIs personally, so the tooltip solution seems adequate.  You can ask Ken what his preference is.
* The sigma checkbox and textbox next to it is grayed out on the project I opened (with the multiplier set to a default value of 3), and I'm not sure why, and there is no explanatory tooltip over it telling me either.
I was using the Themis project in /work/projects/IPCE/IpceData.


You seem to have addressed most of Ian's concerns, and I have no problems with your code.

Actually I want to modify what I said about the spinner boxes.  As long as the user can manually enter a value, and they are set to a sensible default value, then that would be fine.

twilson271828 (approved):
Code looks good, and you've addresses the issues I had with my previous review.  Good job!

PR #215, opened 2018-06-07
Ipce bundle window observation solve settings (preliminary)

ihumphrey:
This is a preliminary attempt to get the tree view into the jigsaw setup bundloe observation solve settings tab.

Note that nothing will show up right now: to fix that, comment out JigsawSetupDialog.cpp:1142:

osspm->setRoot(item);


ihumphrey:
Tyler approves :space_invader: :sparkler: :taco: :bell:

PR #216, opened 2018-06-07
Replaced cneteditor's usage of ControlCubeGraphNodes with serial numbers References #5434

jessemapel:
This is a rather massive pull request. The design of the CnetEditorWidget is not documented anywhere that I can find, so I had to kind of reverse engineer it. So, here's my understanding:

The control network information is stored in two trees: the image-image tree and the image-point tree. Both inherit from the AbstractTreeModel class that contains all of the tree functionality. The specific tree classes contain functionality related to what they actually store in the tree.

The image-image tree contains a parent item for each image. Then, those items have a child leaf for each other image that they share a control point with. This represents the connections between images in the control network.

The image-point tree again contains a parent item for each image. Then, those items have a child leaf for each control point in that image. This represents the individual points and measures in each image.

The tree items have a somewhat complex class structure. The top most class is the AbstractTreeItem which represents an item that contains some data and is located anywhere in the tree. There are two sets of classes that inherit off of AbstractTreeItem. The first set is AbstractParentItem and AbstractLeafItem which represent parent and leaf items respectively. The parent item class implements searching child items. The child item class mostly disallows anything related to looking for children of that item. The second set is AbstractImageItem, AbstractPointItem, AbstractMeasureItem, and AbstractNullDataItem. These classes represent items that contain a specific type of data. The actual implementation classes inherit from one class from each of these group. For example, the ImageParentItem inherits from AbstractImageItem and AbstractParentItem.

For this pull request the only one changed was the AbstractImageItem. Before it contained a pointer to a vertex in the ControlNet's graph. This was removed to isolate the graph implementation inside ControlNet. Now, it contains a pair of image serial number and a pointer to the control network. Ideally, this would just be a serial number. In order to access everything needed by the filters and tree models, we need both the control network and serial number. The filter design only allows us to pass one item to the filter, though. Hence, the two items are combined into a single pair item. The AbstractTreeItem class also, only returns a single piece of data and expects it to be a pointer. So, the ImageTreeItem has to create a new QString, ControlNet pair to ensure the pointer is valid.

The filter design is much simpler. There is a parent AbstractFilter class that represents something that filters. Below that are three more abstract filter classes: AbstractMultipleChoiceFilter, AbstractNumberFilter, and AbstractStringFilter. These classes represent filters that filter based on different groups of criteria. There are also several filters that inherit directly off of AbstractFilter. The main change to the filter classes was to modify the evaluate(image) method to take a pair of serial number and network instead of a graph vertex. There is very little changed in individual filter classes besides method signatures. The majority of the logic changes are in AbstractFilter.


isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1094,6 +1104,33 @@ namespace Isis {
...
+   *
+   * @param serialNumber the serial number of the image to find images adjacent to.
+   *
+   * @returns @b QList<QString> The serial numbers of all adjacent images.
+   *
+   * @TODO Replace this with updated graph functionality once boost graph is in.
+   */
+  QList< QString > ControlNet::getAdjacentImages(QString serialNumber) const {
+    if (!cubeGraphNodes->contains(serialNumber)) {
+      IString msg = "Cube Serial Number [" + serialNumber + "] not found in "
```
AgoinsUSGS: Should this be a QString?

AgoinsUSGS (changes requested):
Looks good from what I can tell. Just one comment on a usage of IString instead of QString.

AgoinsUSGS (approved):

jessemapel:
@AgoinsUSGS Can you review the changes to GraphToString. The idea is that we always need to output the graph structure in a sorted order for testing purposes. Our usage of QHashes makes this a challenge. Hence all of the QStringList::sort calls

AgoinsUSGS (approved):
Looks good

PR #218, opened 2018-06-07
Got the Health Monitor into IPCE.

AgoinsUSGS:
Created all necessary classes and methods to get the health monitor into ipce. 
Deleted the temporary healthmonitor application that was used to view the HealthMonitor pre-ipce.
Added some documentation for the code thus far. 
Keeping it real.


isis/make/config.linux-x86_64_Fedora25
```diff
@@ -115,7 +116,7 @@ QTINCDIR += -I$(ISIS3LOCAL)/include/qt/qt5.7.1/QtWidgets
 QTINCDIR += -I$(ISIS3LOCAL)/include/qt/qt5.7.1/QtXml
 QTINCDIR += -I$(ISIS3LOCAL)/include/qt/qt5.7.1/QtXmlPatterns
 QTLIBDIR = -L$(ISIS3LOCAL)/lib
-QTLIB    = -lQt5Core -lQt5Concurrent -lQt5XmlPatterns -lQt5Xml -lQt5Network -lQt5Sql -lQt5Gui -lQt5PrintSupport -lQt5Positioning -lQt5Qml -lQt5Quick -lQt5Sensors -lQt5Svg -lQt5Test -lQt5OpenGL -lQt5Widgets -lQt5Multimedia -lQt5MultimediaWidgets -lQt5WebChannel -lQt5WebEngine -lQt5WebEngineWidgets -lQt5DBus
+QTLIB    = -lQt5Core -lQt5Concurrent -lQt5XmlPatterns -lQt5Xml -lQt5Network -lQt5Sql -lQt5Gui -lQt5PrintSupport -lQt5Positioning -lQt5Qml -lQt5Quick -lQt5Sensors -lQt5Svg -lQt5Test -lQt5OpenGL -lQt5Widgets -lQt5Multimedia -lQt5MultimediaWidgets -lQt5WebChannel -lQt5WebEngine -lQt5WebEngineWidgets -lQt5DBus -lQt5Charts
```
jessemapel, line 119: We need to make sure that this is available on all of our system! Also, adding qt dependencies can cause a lot of issues because compiling qt is troublesome sometimes.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -223,23 +225,23 @@ namespace Isis {
     QString details = "";
     if (hasIslands()) {
       status = "Broken!";
-      details = "This network has " + toString(numIslands()) + " islands.";
+      details = "This network has " + toString(numIslands()) + " island(s).";
```
jessemapel: If this happens, there will always be at least 2 islands, so we can just always use the plural form.

isis/src/qisis/objs/ControlHealthMonitorWidget/ControlHealthMonitorWidget.cpp
```diff
@@ -0,0 +1,769 @@
...
+    if (m_vitals->getStatus() == "Broken!") updateStatus(0);
+    else if (m_vitals->getStatus() == "Weak!") updateStatus(1);
+    else if (m_vitals->getStatus() == "Healthy!") updateStatus(2);
+
+    QPieSeries *series = new QPieSeries();
+    series->append("Ignored", m_vitals->numIgnoredPoints());
+    series->append("Locked", m_vitals->numLockedPoints());
+    series->append("Free", m_vitals->numFreePoints());
+    series->append("Constrained", m_vitals->numConstrainedPoints());
+    series->append("Fixed", m_vitals->numFixedPoints());
```
jessemapel: these 5 things are not mutually exclusive. Ignored and locked are separate boolean qualities of the points. The Free/Constrained/Fixed properties are mutually exclusive.

jessemapel (changes requested):
Looks good. A few small tweaks. I also realized we don't have a unit test for ControlNetVitals. We need to add that. It can be in this pr or a separate pr after this to make it cleaner.

jessemapel (approved):

jessemapel:
Once gui testing is done for this, we can merge it.

PR #230, opened 2018-06-13
Added xml schema for img xmlns and fix mosaic reingestion.

krlberry:
This fixes a "cannot parse XML" error when trying to re-ingest.

### DIFF
--- a/isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
+++ b/isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
@@ -410,7 +410,8 @@ namespace Isis {
                 "http://pds.nasa.gov/pds4/sp/v1"); 
     }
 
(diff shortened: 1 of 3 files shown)


isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
```diff
@@ -410,6 +410,14 @@ namespace Isis {
                 "http://pds.nasa.gov/pds4/sp/v1"); 
     }
 
+    if (imageObject.compare("Array_2D_Image") == 0) {
```
jcwbacker: do we only add IMG for 2D? I think it also applies to 3D

krlberry:
For this to work, a couple of translation files were added / updated at `/work/projects/tgo/data/tgo/translations/`:

* tgoCassisMosaic.trn
* tgoCassisMosaicBandBin.trn

jcwbacker (approved):

PR #232, opened 2018-06-13
Added additional signals/slots to ControlNet, ControlPoint, ControlMeasure

AgoinsUSGS:
Added a bunch of additional slots to the aforementioned classes to allow for more efficient communication between the Control Net and the Network Vitals class. 

Referencing the Matrix that Jesse made, I went through all relevant modifications that can be made to these three classes and made an enum for the possible modifications that could be done. 
Then, whenever a modification is made we call the emitModified() signal and pass in data with respect to that specific modification.

For instance, if I change the type of a control point from Free to Constrained, the call would look something like:

parentNetwork->emitPointModified(this, ControlPoint::TypeChanged, Free, Constrained)

the pointModified and measureModified signals emit the object that was modified (the ControlPoint, ControlMeasure, etc), the Type of Modification, the OldValue, and the NewValue for that change.
QVariants are used to support multiple datatypes.

These signals are called from the ControlNetwork and are attached in ControlNetVitals to the validate() SLOT. Ideally after a signal is intercepted in the NetworkVitals class, we could parse the data emited with the signal and construct some sort of history object out of it and emit that as well. 

Because we have several different modification types we can then validate the NetworkVitals different based on what was modified and what type of modification was made. Because all of this is given to us in the signal we can Validate differently and only check for certain things so that we don't have to go through the entire revalidation process after every little change is made.


isis/src/control/objs/ControlMeasure/ControlMeasure.cpp
```diff
@@ -360,9 +365,12 @@ namespace Isis {
     if (IsEditLocked())
       return MeasureLocked;
 
+
     bool oldStatus = p_ignore;
     p_ignore = newIgnoreStatus;
 
+    Parent()->emitMeasureModified(this, IgnoredModified, oldStatus, p_ignore);
```
jessemapel: need to check that the parent isn't null before calling this.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -755,9 +793,11 @@ namespace Isis {
 
     // notify CubeSerialNumbers of the loss of this point
     foreach(ControlMeasure * measure, point->getMeasures()) {
-      measureDeleted(measure);
+      emit measureRemoved(measure);
```
jessemapel, line 796: This needs to be changed back to measureDeleted. This doesn't update the network's graph structure, and measureDeleted will emit the measureRemoved signal

isis/src/control/objs/ControlPoint/ControlPoint.cpp
```diff
@@ -306,7 +307,7 @@ namespace Isis {
 
     // notify parent network of the change
     if (parentNetwork) {
-      parentNetwork->measureDeleted(cm);
+      parentNetwork->emitMeasureRemoved(cm);
```
jessemapel, line 310: This need to go back to measureDeleted. The emit call doesn't update the graph and measureDeleted will emit the same signal still.

isis/src/control/objs/ControlPoint/ControlPoint.cpp
```diff
@@ -254,6 +254,7 @@ namespace Isis {
     // notify parent network if we have one
     if (parentNetwork) {
       parentNetwork->measureAdded(measure);
+      parentNetwork->emitNewMeasure(measure);
```
jessemapel, line 257: We don't need to emit new measure here, because measureAdded will emit that already

isis/src/control/objs/ControlNet/ControlNet.h
```diff
@@ -329,18 +350,30 @@ namespace Isis {
 
     signals:
       void networkStructureModified();
+      void networkModified(ControlNet::ModType type);
+      void pointModified(ControlPoint *point, ControlPoint::ModType type, QVariant oldValue, QVariant newValue);
+      void measureModified(ControlMeasure *measure, ControlMeasure::ModType type, QVariant oldValue, QVariant newValue);
+      void pointDeleted(ControlPoint *point);
+      void newPoint(ControlPoint *);
+      void newMeasure(ControlMeasure *);
+      void measureRemoved(ControlMeasure *);
```
jessemapel, line 359: Based on the other changes, emitNewMeasure and emitMeasureRemoved are no longer needed and can be removed

jessemapel (dismissed):
Changes to fix seg faults

PR #233, opened 2018-06-13
Fix cleanup of views and temporary fix to geometry/state settings.

TracieSucharski:
This fixes some seg faults when views are closed and also some save/restore problems when views are closed.


SgStapleton:
Looks legit

PR #235, opened 2018-06-14
Fixes version for nanoflann

jlaura:
Environment.yml had a typo in the nanoflann version.


Kelvinrr:
Ran into this the other day, not sure how this happened since the file is generated by conda.

PR #236, opened 2018-06-14
removed ComputeResidualsMillimeters method and moved computation of f…

kledmundson:
…ocalplane computedx, computedy into the ComputeResiduals method. This removes unneeded computations.


dcookastro:
Looks good!

PR #239, opened 2018-06-14
Added toolbars and menus to CnetEditorView. Fixes #5416

### LINKED ISSUE #5416: findFeaturesSegment.py Error
**ISIS version(s) affected**: isis8.1.0-RC2

**Description**  
The newly available findFeaturesSegment script errors when working on overlapping images when not all segments overlap, even when new parameters minarea and/or minthickness are set to values greater than 0.  

I tested on images noted in [here](https://github.com/DOI-USGS/ISIS3/pull/5091#issuecomment-1371632440) and displayed in the qmos figure.  

Initially I ran the test without any new script parameters assuming the script would test for zero overlap while working on segments, but that failed. I'm having a difficult time isolating the most relevant error message(s) because of all the output (and various warnings for deprecated/unsupported things), but I think the problem is mostly near the end of the script when it is trying to merge expected segment networks that don't exist. Here's a snippet with leading path info removed:

```
Traceback (most recent call last):
  File "isis8.1.0-RC2/bin/findFeaturesSegment.py", line 383, in <module>
    findFeaturesSegment(ui) 
  File "isis8.1.0-RC2/bin/findFeaturesSegment.py", line 354, in findFeaturesSegment
    output = output.get()
  File "isis8.1.0-RC2/lib/python3.9/multiprocessing/pool.py", line 771, in get
    raise self._value
  File "isis8.1.0-RC2/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "isis8.1.0-RC2/lib/python3.9/multiprocessing/pool.py", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "isis8.1.0-RC2/bin/findFeaturesSegment.py", line 234, in generate_cnet
    segmented_net = cnet.from_isis(new_params["ONET"])
  File "isis8.1.0-RC2/lib/python3.9/site-packages/plio/io/io_controlnetwork.py", line 114, in from_isis
    with IsisStore(path, mode='rb') as store:
  File "isis8.1.0-RC2/lib/python3.9/site-packages/plio/io/io_controlnetwork.py", line 206, in __init__
    self._open()
  File "isis8.1.0-RC2/lib/python3.9/site-packages/plio/io/io_controlnetwork.py", line 220, in _open
    self._handle = open(self._path, self._mode)
FileNotFoundError: [Errno 2] No such file or directory: 'M156671044LE_ff_1_from_images_segment2.net'
```
I do get some basic findfeatures output despite the above errors and get a segment1.net file (see more below on that).

When I chose values>0 for minarea and minthickness I got somewhat different errors.  I thought using minarea and minthickness would help in that the script would completely avoid the segments that had zero overlap and avoiding the expectation, but I don't think it even got that far.

```
DEBUG:root:From images overlapping Match: []
DEBUG:root:output: [None, None, None, None]
DEBUG:root:onets: []
Traceback (most recent call last):
  File "isis8.1.0-RC2/bin/findFeaturesSegment.py", line 383, in <module>
    findFeaturesSegment(ui) 
  File "isis8.1.0-RC2/bin/findFeaturesSegment.py", line 366, in findFeaturesSegment
    final_images = set.union(*tolists)
TypeError: unbound method set.union() needs an argument

```
The above is right after overlapstats output. There is no call to findfeatures that I can see.
It seems the new parameters don't work as intended or I don't understand how to use them.

I was not able to capture what was going to screen (using 'tee' and running the command directly in my shell), so I had to copy and paste all of it into a text file. (Although the amount of information going to screen is a lot, not having it is a problem - ideas on how to capture it? Essential for debugging.). It's clear from findfeatures program output (via Defaults/) that there are segments it could not find points for (normal 'no points' output and clean exit), but it did find some points for overlapping segments and that should be valid output for this script.

**How to reproduce**  
See my user work area Isis3Tests/FFSegmentScript/Defaults/ for input images, fromlist and command (run.cmd) along with what went to screen in Stdout.txt. 

See Isis3Tests/FFSegmentScript/PolyConstraints/ for similar but with values entered for minarea and minthickness. See Stdout.txt there for that screen grab.

The attached qmos screengrab is under the Defaults/ directory.  The purple footprint is the 'match' image and the blue and green footprints are in the fromlist.  Segment 1 for all images was the successful chunk and are the darker portions of the footprints and have the resulting segment1.net overlain on top.  Segment 2 seems to be the problem, but I can't tell if it's across the board or just for green seg 2 and match seg 2 poly's. 

FFScript_ErrorFootprints

**Possible Solution**
In the Defaults/ case, I can see the findfeatures call for M156671044LE_ff_1_from_images_segment2.net and that "No control points found!!", so maybe it is just the script having the expectation that there will be success and the output network will be created when that is not always going to be the case. So that might need to be a test if not already one.

But I also see via the findfeatures log (M156671044LE_ff.log) that it is indeed trying to work on segments that do not overlap at all which I think we should completely be avoiding.  So if not already testing on valid overlaps via the overlapstats info (don't know what is being passed since it's going to /tmp), then for sure via the minarea/minthickness parameters, but test two shows that has some issues still.

**Additional Context**
I first tested the updated script on a different 3 image set where all segments overlapped completely and a final network was produced.  I then ran a test on a larger set of data (dozens of images) that blew up and the above images came from that larger set (plus these are the same images in the mentioned PR). So currently, the script seems to run in default mode for the most ideal situation but not the more realistic complex situations.

kaitlyndlee:
Removed save active control net button and made it so that only one instance of the sorting dialog can exist at once.

### DIFF
--- a/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
+++ b/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
@@ -464,7 +464,6 @@ namespace Isis {
     QSize iconSize(25, 45);
     m_permToolBar->setIconSize(iconSize);
     m_permToolBar->setObjectName("PermanentToolBar");
-    m_permToolBar->setToolTip("PermanentToolBar");
 
     addToolBar(m_permToolBar);
 
--- a/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
+++ b/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
@@ -338,6 +338,7 @@ namespace Isis {
     m_fileMenuActions.append(exitAction);
     m_permToolBarActions.append(exitAction);
 
+    // 6-13-18 Kaitlyn Lee - Commented this out because it is not currently being used
     // QAction *saveNet = new QAction("&Save Active Control Network", this);
     // saveNet->setIcon( QIcon::fromTheme("document-save") );
     // saveNet->setShortcut(Qt::CTRL + Qt::Key_S);
--- a/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
+++ b/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
@@ -217,7 +217,6 @@ namespace Isis {
 
 
   void IpceMainWindow::cleanupViewDockList(QObject *obj) {
-
     QDockWidget *dock = static_cast<QDockWidget *>(obj);
     if (dock) {
       m_viewDocks.removeAll(dock);
@@ -226,14 +225,12 @@ namespace Isis {
 
 
   /**
-   * @description This slot is connected from Directory::viewClosed(QWidget *) signal.  It will
+   * This slot is connected from Directory::viewClosed(QWidget *) signal.  It will
    * close the given view and delete the view. This was written to handle
    *
-   * @param view QWidget*
-   *
+   * @param view QWidget* The view to close.
    */
   void IpceMainWindow::removeView(QWidget *view) {
-    
     view->close();
     delete view;
   }
@@ -781,5 +778,4 @@ namespace Isis {
   void IpceMainWindow::raiseWarningTab() {
     m_warningsDock->raise();
   }
-
 }
--- a/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
+++ b/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
@@ -338,31 +338,6 @@ namespace Isis {
     m_fileMenuActions.append(exitAction);
     m_permToolBarActions.append(exitAction);
 
-    // 6-13-18 Kaitlyn Lee - Commented this out because it is not currently being used
-    // QAction *saveNet = new QAction("&Save Active Control Network", this);
-    // saveNet->setIcon( QIcon::fromTheme("document-save") );
-    // saveNet->setShortcut(Qt::CTRL + Qt::Key_S);
-    // saveNet->setToolTip("Save current active control network");
-    // saveNet->setStatusTip("Save current active control network");
-    // QString whatsThis = "<b>Function:</b> Saves the current active<i>"
-    //                     "control network</i>";
-    // saveNet->setWhatsThis(whatsThis);
-    // connect(saveNet, SIGNAL(triggered()), m_directory, SLOT(saveActiveControl()));
-    // m_permToolBarActions.append(saveNet);
-
-//  m_saveAsNet = new QAction(QPixmap(toolIconDir() + "/mActionFileSaveAs.png"),
-//                            "Save Control Network &As...",
-//                            m_matchTool);
-//  m_saveAsNet->setToolTip("Save current control network to chosen file");
-//  m_saveAsNet->setStatusTip("Save current control network to chosen file");
-//  whatsThis = "<b>Function:</b> Saves the current <i>"
-//      "control network</i> under chosen filename";
-//  m_saveAsNet->setWhatsThis(whatsThis);
-//  connect(m_saveAsNet, SIGNAL(triggered()), this, SLOT(saveAsNet()));
-
-
-
-
     QAction *undoAction = m_directory->undoAction();
     undoAction->setShortcut(Qt::Key_Z | Qt::CTRL);
 
@@ -486,36 +461,19 @@ namespace Isis {
    */
   void IpceMainWindow::createToolBars() {
     m_permToolBar = new QToolBar(this);
-    //m_activeToolBar = new QToolBar(this);
-    //m_toolPad = new QToolBar(this);
-
     QSize iconSize(25, 45);
-
     m_permToolBar->setIconSize(iconSize);
-    //m_activeToolBar->setIconSize(iconSize);
-    //m_toolPad->setIconSize(iconSize);
-
     m_permToolBar->setObjectName("PermanentToolBar");
-    //m_activeToolBar->setObjectName("ActiveToolBar");
-    //m_toolPad->setObjectName("ToolPad");
+    m_permToolBar->setToolTip("PermanentToolBar");
 
     addToolBar(m_permToolBar);
-    //addToolBar(m_activeToolBar);
-    //addToolBar(m_toolPad);
 
     foreach ( QAction *action, m_directory->permToolBarActions() ) {
       m_permToolBar->addAction(action);
     }
 
     foreach (QAction *action, m_permToolBarActions) {
-      if (action->text() == "&Save Active Control Network") {
-        m_permToolBar->addSeparator();
-        m_permToolBar->addAction(action);
-        m_permToolBar->addSeparator();
-      }
-      else {
-        m_permToolBar->addAction(action);
-      }
+      m_permToolBar->addAction(action);
     }
   }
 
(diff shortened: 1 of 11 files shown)


isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -226,10 +226,10 @@ namespace Isis {
 
 
   /**
-   * @description This slot is connected from Directory::viewClosed(QWidget *) signal.  It will 
+   * @description This slot is connected from Directory::viewClosed(QWidget *) signal.  It will
    * close the given view and delete the view. This was written to handle
-   * 
-   * @param view QWidget* 
+   *
+   * @param view QWidget*
```
celbal: What is this parameter used for? It's not used in the method so why is it here?

isis/src/qisis/apps/ipce/IpceMainWindow.h
```diff
@@ -199,8 +200,8 @@ namespace Isis {
       static const int m_maxRecentProjects = 5;
 
       QToolBar *m_permToolBar; //!< The toolbar for actions that rarely need to be changed.
-      QToolBar *m_activeToolBar; //<! The toolbar for the actions of the current tool.
-      QToolBar *m_toolPad; //<! The toolbar for the actions that activate tools.
+      //QToolBar *m_activeToolBar; //<! The toolbar for the actions of the current tool.
```
celbal: If these are commented out can they be removed?

isis/src/qisis/objs/CnetEditorView/CnetEditorView.h
```diff
@@ -51,11 +51,13 @@ namespace Isis {
    * @author 2018-04-04 Tracie Sucharski
    *
    * @internal
-   *    @history 2018-06-1 Kaitlyn Lee - Because AbstractProjectItemView now inherits
-   *                                from QMainWindow, I added a dummy central widget
-   *                                and set its layout to the grid layout. We used to set
-   *                                the whole CnetEditorView widget's layout, now we only
-   *                                set the central widget's layout.
+   *    @history 2018-06-01 Kaitlyn Lee - Because AbstractProjectItemView now inherits from QMainWindow,
+   *                              I added a dummy central widget and set its layout to the grid layout.
```
celbal: If your history comment is multiple lines all lines after the first need to be indented 4 spaces after the start of your name. Meaning that the second line should line up with the l in your first name

isis/src/qisis/objs/CnetEditorView/CnetEditorView.h
```diff
@@ -79,6 +81,9 @@ class CnetEditorView : public AbstractProjectItemView {
     void load(XmlStackedHandlerReader *xmlReader);
     void save(QXmlStreamWriter &stream, Project *project, FileName newProjectRoot) const;
 
+    private:
```
celbal: You have two private sections. You should combine them.

isis/src/qisis/objs/CnetEditorWidget/CnetEditorSortConfigDialog.cpp
```diff
@@ -23,6 +23,8 @@ namespace Isis {
     CnetEditorWidget *cnetWidget) : QDialog(cnetWidget) {
     m_cnetWidget = cnetWidget;
 
+    setWindowTitle("Table Sorting Dialog");
```
celbal: Do we want this to be called Table Sorting Dialog? Why not Table Sorting?

isis/src/qisis/objs/CnetEditorWidget/CnetEditorWidget.cpp
```diff
@@ -75,6 +75,8 @@ namespace Isis {
 
     m_updatingSelection = false;
 
+    m_sortDialog = NULL;
```
makaylas: Because you newed something you also need to delete the object

isis/src/qisis/objs/Directory/Directory.h
```diff
@@ -380,7 +382,7 @@ namespace Isis {
       //void imagesAddedToProject(ImageList *images);
       void updateControlNetEditConnections();
 
-      void saveActiveControl();
+      //void saveActiveControl();
```
celbal: Can this be deleted?

celbal (changes requested):

celbal:
Also, please make sure that you have two blank lines in between methods

isis/src/qisis/apps/ipce/IpceMainWindow.h
kaitlyndlee: These are removed in another PR, since I did not want to do too much on this branch.

isis/src/qisis/objs/CnetEditorWidget/CnetEditorSortConfigDialog.cpp
kaitlyndlee: I was not sure what to name it. Tracie said it was okay for now, but I can change it to whatever is the best.

celbal: I personally don't like Dialog at the end

isis/src/qisis/apps/ipce/IpceMainWindow.h
celbal: Might as well remove them here, just to make sure that they are deleted. We don't want any more commented out code if we can avoid it.

isis/src/qisis/objs/CnetEditorView/CnetEditorView.h
TracieSucharski: This is done throughout Isis:  private methods in one private section vs member variables in another private section.  I can't find anything in coding standards regarding this.  Has this changed?

isis/src/qisis/objs/CnetEditorWidget/CnetEditorSortConfigDialog.cpp
TracieSucharski: This is sort of a gui standard (appending the Dialog to the class name).  TableSortingDialog indicates this is a gui dialog class.  To me, TableSorting indicates that the class actually performs the table sorting.

TracieSucharski (approved):

isis/src/qisis/objs/CnetEditorWidget/CnetEditorSortConfigDialog.cpp
kaitlyndlee: The window title is the title of the dialog that pops up in ipce, not the class itself.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
kaitlyndlee: It is used in the method; there's a line that closes the view with view->close().

isis/src/qisis/objs/Directory/Directory.h
kaitlyndlee: Yes, I deleted it.

isis/src/qisis/objs/CnetEditorWidget/CnetEditorSortConfigDialog.cpp
TracieSucharski: OOps, sorry thought you were talking about the class itself.  Yes, I definitely agree with Makayla the title should not have the word Dialog.  Sorry for my confusion.

celbal (approved):

PR #240, opened 2018-06-14
Cleaned up all views and parent class. Fixes #5440 #5389

kaitlyndlee:
Removed methods that were returning toolbar actions, set an active control on import (when only 1 is being imported, removed undoExecution() when importing a control net, and moved signals from Directory to the two views when an active control network is set.

### DIFF
--- a/isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
+++ b/isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
@@ -236,14 +236,12 @@ namespace Isis {
     zoomTool->activate(true);
   }
 
-
   /**
-   * This is a slot function which is called when directory emits a siganl to
-   * CubeDnView when an active control network is set. It enables the control
-   * network editor tool in the toolpad.
+   * A slot function that is called when directory emits a siganl that an active
+   * control network is set. It enables the control network editor tool in the
+   * toolpad and loads the network.
    */
   void CubeDnView::enableControlNetTool(bool value) {
-
     foreach (QAction * action, m_toolPad->actions()) {
       if (action->objectName() == "ControlNetTool") {
         action->setEnabled(value);
@@ -303,16 +301,6 @@ namespace Isis {
 
     return item->isShape();
   }
-
-  /**
-   * Returns a list of actions for the tool pad.
-   *
-   * @return @b QList<QAction*> The actions
-   */
-  QList<QAction *> CubeDnView::toolPadActions() {
-    return m_toolPad->actions();
-  }
-
 
   /**
    * Slot to connect to the currentChanged() signal from a selection
--- a/isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
+++ b/isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
@@ -115,21 +115,11 @@ namespace Isis {
     connect(m_workspace, SIGNAL( cubeViewportAdded(MdiCubeViewport *) ),
             this, SLOT( onCubeViewportAdded(MdiCubeViewport *) ) );
 
-
-    // !!!!!!!   TODO  LOOK AT THIS       !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-    // These connections should be at higher level, directory or project???? ::addCubeDnView().
-    Project *activeProject = directory->project();
-    // These connect signals listen to the active project, and if a control network
-    // or list of control networks is added, then the enableControlNetTool() function is called.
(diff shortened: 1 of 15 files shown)


isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
```diff
@@ -115,22 +115,12 @@ namespace Isis {
     connect(m_workspace, SIGNAL( cubeViewportAdded(MdiCubeViewport *) ),
             this, SLOT( onCubeViewportAdded(MdiCubeViewport *) ) );
 
-
-    // !!!!!!!   TODO  LOOK AT THIS       !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
celbal, line 119: What happened to this code? Was it moved to a higher level?

isis/src/qisis/objs/Footprint2DView/Footprint2DView.cpp
```diff
@@ -313,39 +316,21 @@ namespace Isis {
...
+   * control network is set. It enables the control network editor tool in the
+   * toolpad and loads the network.
+   * TODO 2018-06-14 Kaitlyn Lee - Commented out code that loads the network,
+   * since it does not actually load the network. Left it here to research further.
    */
-  QList<QAction *> Footprint2DView::toolPadActions() {
-    return m_toolPad->actions();
+  void Footprint2DView::enableControlNetTool(bool value) {
+    m_controlNetTool->setEnabled(value);
+    // if (value) {
```
celbal: Can this be deleted?

isis/src/qisis/objs/CubeDnView/CubeDnView.cpp
kaitlyndlee: Directory now emits a signal to CubeDnView saying that an active control network was set. There were duplicate connections in Directory and CubeDnView doing the same thing, so there was no need for this code.

isis/src/qisis/objs/Footprint2DView/Footprint2DView.cpp
kaitlyndlee: I left a comment above just saying I left it in there to remind myself how the behavior works and I will research it further once Tracie tells me how it should be done.

kdl222:
Wait to merge until further testing is done.

isis/src/qisis/objs/Directory/ImportControlNetWorkOrder.cpp
```diff
@@ -206,32 +207,12 @@ namespace Isis {
...
-      project()->directory()->model()->removeItem(currentItem);
-
-      foreach (Control *control, *m_list) {
-        delete control;
-      }
-      delete m_list;
-    }
+    // If one control network was imported, then activeControl() will set the
+    // active control to that control network
+    project()->activeControl();
```
TracieSucharski: I would just add a bit more to the comment, such as "If one control network was imported, then activeControl() will set the newly imported control net to the active if project does not curently have an active control set and there are no other control nets in the project."

isis/src/qisis/objs/Footprint2DView/Footprint2DView.cpp
TracieSucharski: Do you have this version built somewhere? I would like to test since I can't see anywhere the loadNetwork  slot is called which is the only way the network is loaded into footprint view.  Am I missing a call or connection somewhere? I know we talked about this and it looked like it might have been called twice, but now I'm not sure if it's ever called?

TracieSucharski: I think leaving the call to ControlNetTool::loadNetwork in Footprint2DView::enableControlNetTool would be the cleanest.

kaitlyndlee: I have a build in /work/projects/IPCE/m05440_m05389/isis/ Calling loadNetwork() here does not actually load the network. There is a method, displayNetwork(), that displays the network if the "Display" button is clicked in the active toolbar of the view. I was under the impression that Ken told us to keep the functionality of qmos, where a network is not loaded until you click on the CnetEditor tool.

TracieSucharski: Ok, I see how it is working. Thinking we might want to document this somewhere:
I think MosaicControlNetTool::displayNetwork() only turn the display of pt on and off. It assumes the network has already been set and the GraphicsItems created.  But, MosaicControlNetTool::updateTool call ::openControlNet which then calls ::loadNetwork.  I think ::updateTool is called when tool is enabled or chosen.  So, that is why we don't need to call loadNetwork.

PR #244, opened 2018-06-15
ddd2isis importing multiband ddd files. Fixes #703

### LINKED ISSUE #703: "caminfo" has concurrency problems when multiple instances use "CAMSTATS" option
---

Author Name: **Travis Addair** (Travis Addair)

Original Assignee: Travis Addair

---

"caminfo" can crash when multi-processing because its "CAMSTATS" option redirects the output of "camstats" to a temporary file, which may be used by other instances of the application.

kaitlyndlee:
ddd2isis used to not calculate the bit type of a ddd file correctly and would throw an exception if a user would try to import a multiband file. We received a python program that showed us how to correctly calculate the bit type. I also added a few tests and refactored the code.

### DIFF
--- a/isis/src/base/apps/ddd2isis/ddd2isis.cpp
+++ b/isis/src/base/apps/ddd2isis/ddd2isis.cpp
@@ -45,7 +45,7 @@ void IsisMain() {
    EndianSwapper swp("MSB");
 
   // Verify that the file is a ddd by reading in the first 4 bytes and
-  // comparing the magic numbers
+  // comparing the magic numbers. The magic number for a ddd file is 1659.
   readBytes.readLong = 0;
   fin.seekg(0);
   fin.read(readBytes.readChars, 4);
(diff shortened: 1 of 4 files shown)


isis/src/base/apps/ddd2isis/ddd2isis.cpp
```diff
@@ -41,86 +33,142 @@ void IsisMain() {
...
   fin.seekg(0);
   fin.read(readBytes.readChars, 4);
+  if( fin.fail() || fin.eof() ) {
+    string msg = "Could not read the magic number in the input file [" + from + "]";
+    throw IException(IException::Io, msg, _FILEINFO_);
+  }
   readBytes.readFloat = swp.Float(readBytes.readChars);
 
-  if(readBytes.readLong != 0x67B) {
+  if(readBytes.readLong != 1659) {
```
celbal, line 58: 1659 seems like a magic number. Can you add a comment to explain it?

kaitlyndlee: I added on to the comment above that chunk of code explaining the magic numbers.

kaitlyndlee: I talked to Stuart about clarifying who we got the magic number from, and he said not to put anyone's info in the comments. For now, just specifying that 1659 is the number that identifies a ddd is okay.

celbal (approved):

PR #254, opened 2018-06-18
BundleAdjust no longer outputs text to the command line for ipce. jigsaw is unaffected other than spacing changes. Fixes #4171.

celbal:


krlberry (approved):
:+1: Looks good!

PR #255, opened 2018-06-18
The geometry and state of ipce are now saved. Fixes #5436.

### LINKED ISSUE #5436: Kaguyasp2ascii doesn't recognize current SP *.spc files
I just downloaded some Kaguya Spectral Profiler (SP) *.spc files. When I try to convert them to ASCII using the "kaguyasp2ascii" ISIS routine, I get an error message: 

**USER ERROR** Input file
[SP_2C_03_01213_N243_E1515.spc] is not a valid Kaguya Spectral Profiler file.

That file has just been downloaded from their site so I don't know why it wouldn't be a valid SP file. Maybe their format has changed from the last iteration of "kaguyasp2ascii"?

Thanks...

celbal:


isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -636,49 +672,76 @@ namespace Isis {
    *   @history Ian Humphrey - Settings are now read on a project name basis. References #4358.
    *   @history Tyler Wilson 2017-11-02 - Settings now read recent projects.  References #4492.
    *   @history Tyler Wilson 2017-11-13 - Commented out a resize call near the end because it
-   *                                      was messing with the positions of widgets after a
-   *                                      project was loaded.  Fixes #5075.
+   *                was messing with the positions of widgets after a project was loaded.  
+   *                Fixes #5075.
+   *   @history Makayla Shepherd 2018-06-10 - Settings are read from the project root ipce.config.
```
krlberry, line 637: This isn't actually a comment on this line. It looks like the comments above here, lines 664 - 667 need to be updated to describe the new config file location as well.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -636,49 +672,76 @@ namespace Isis {
...
+    }
+    
     if (project->name() == "Project") {
       setWindowTitle("ipce");
+    }
+    else {
+      setWindowTitle( project->name() );
+      QString projName = project->name();
+      setWindowTitle(projName );
+    }
```
krlberry: Does `setWindowTitle( project->name() ) ;` not work directly? If not, can line 706 be removed? It looks to me like lines 707 and 708 do the same thing as 706?

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -571,12 +572,47 @@ namespace Isis {
...
+      throw IException(IException::Programmer, msg, _FILEINFO_);
+    }
+    QSettings projectSettings(FileName(project->newProjectRoot() + "/ipce.config").expanded(),
+        QSettings::NativeFormat);
+
+    projectSettings.setValue("geometry", QVariant(geometry()));
+    projectSettings.setValue("windowState", saveState());
+    projectSettings.sync();
+
+    //TODO Do we really need this? Isn't this a global setting? Can the user even change this?
```
krlberry: I don't know the answer to this "todo" but is there any chance it's been answered since the time this PR was put in?

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -636,49 +672,76 @@ namespace Isis {
...
+    else {
+      setWindowTitle( project->name() );
+      QString projName = project->name();
+      setWindowTitle(projName );
+    }
 
-      //  Read  generic geometry
-      if (settings.contains("geometry")) {
-        setGeometry(settings.value("geometry").value<QRect>()); 
+    QSettings projectSettings(FileName(filePath).expanded(), QSettings::NativeFormat);
```
krlberry, line 669: Should this be wrapped in some kind of a if FileName(filePath).fileExists()? From the if on line 697, it looks like it is possible to proceed without an existing filePath.

krlberry (changes requested):
A few comments and questions. I don't have a good background with this so apologies in advance of some of my comments don't make much sense.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
TracieSucharski: Yes, the user can change the maxThreadCount, so I think this should remain as a global setting.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
celbal: Whoops that was a mistake

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
celbal: No, if the file does not exist then QSettings will create the file. The reason that we check if the file exists on lines 693 and 697 are because we want the following chain to happen:
1. If projectroot()/ipce.config exists use that because that means there are project specific settings
2. If projectroot()/ipce.config does not exist but ~/.Isis/ipce.config exists then use those settings (meaning the settings you had when you closed ipce the last time
3. If neither exists then we want ipce to come up full screen because we are assuming that this is the first time the user has opened ipce. In that case QSettings will create the file but we don't really care at that point.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
makaylas: You are correct

celbal: You are correct. I fixed the issue

krlberry: :+1: Thanks!

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
krlberry: :+1:

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
krlberry: @makaylas Ah, makes sense now. Thanks for explaining! :+1: 

I didn't read this relevant part of the QSettings constructor documentation: "If the file doesn't already exist, it is created."

krlberry (approved):

PR #260, opened 2018-06-19
Created SortFilterProxyModel class

twilson271828:
This class allows for the sorting and filtering of image project items in the main project tree from within the JigsawSetupDialog BOSS tab treeview.


ihumphrey:
Built and ran through it really quick, looks good!

I did notice that it is possible to select anything in the project tree view, which then will show up in the observation solve settings tree view (e.g. if you select "Maps", the Maps item will appear in the observation solve settings tree view).

twilson271828:
On my latest modifications to ProjectItemModel::selectedChildItems(), I followed Ryan's suggestion to query the parent of selected indices to see if they have children, and if they do then they are added to the list of selected items.  I also added the grandparent because this is needed to properly set the root index in the treeview of the BOSS tab.

Most of the logic for selecting what to display has been taken out of SubtreeProxyModel::filterAcceptsRow and placed in selectChildItems.   What remains to be done is to filter out anything that isn't an image or an image list (Ian's comment).

twilson271828:
Latest push fixes the segfault caused by users clicking Bundle Adjustment and not selecting any images.
Also, if the user selects objects other than images or image lists, the root Images folder with all the images/image lists will be displayed on the BOSS tab treeview.

PR #270, opened 2018-06-21
Ipce bundle window obs settings

ihumphrey:


chrisryancombs (changes requested):
Looks good, but ipce cant build due to JigsawSetupDialog::on_applySettingsPushButton_clicked() being in header but not cpp

PR #275, opened 2018-06-22
ipce JigsawSetupDialog observation solve settings tab implementation

ihumphrey:


twilson271828 (commented):
Two questions:
When the SPK solve degree keeps getting incremented past degree k=2, the units are expressed as m/s^k (k=3,....).  Is this correct behavior?

Also, in the event that a user might want to ascribe some meaning to these higher degree variable coefficients, maybe we should allow them to edit the description?

twilson271828:
Ken says that he thinks expressing the units currently is correct, but he will look into it.  For now he is not concerned.  With respect to the second question, he does not think they should be editable.

He does want the boxes for SPKDEGREE/SPKSOLVEDEGREE and the corresponding camera pointing boxes to be disabled with default values set unless the user selects ALL in the solve options.  This is the same behavior that jigsaw currently exhibits.

ihumphrey:
@twilson271828 
I've addressed the solvedegree and degree boxes to be disabled unless ALL is set (for pointing and position, respectively).
Note that the "default" values are not always 2 (the way the table is implemented is that the solve degree helps determine the number of rows since ALL needs to look at the solve degree to add rows). I can changes this so the "default" values are always 2 when disabled, but that will take more work.

PR #277, opened 2018-06-22
Fixed ddd2isis breaking prog26

kaitlyndlee:
I also made a quick change to clarify the differences between the old and new headers.


krlberry (approved):
:+1: Looks good to me!

PR #286, opened 2018-06-25
Updated ControlNet and ControlNetVitals to fix problem with numImages…

krlberry:
…BelowMeasureThreshold not matching getImagesBelowMeasureThreshold. 

I verified using cneteditor that the number of images with < n measures matched the updated results in the unitTest

### DIFF
--- a/isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
+++ b/isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
@@ -613,9 +613,10 @@ namespace Isis {
 
     QMap<int, int>::const_iterator i = m_imageMeasureCounts.constBegin();
     while (i != m_imageMeasureCounts.constEnd()) {
-      if (i.key() < num ) {
-        count += i.value();
+      if (i.key() >= num ) {
+        break;
       }
+      count += i.value();
       ++i;
     }
     return count;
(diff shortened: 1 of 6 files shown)


isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
```diff
@@ -610,11 +610,13 @@ namespace Isis {
...
   int ControlNetVitals::numImagesBelowMeasureThreshold(int num) {
     int count = 0;
-    foreach(int measureCount, m_imageMeasureCounts) {
-      if (measureCount > num) {
-        break;
+
+    QMap<int, int>::const_iterator i = m_imageMeasureCounts.constBegin();
+    while (i != m_imageMeasureCounts.constEnd()) {
+      if (i.key() < num ) {
+        count += i.value();
```
jessemapel: QMap is always sorted, so you can just break if i.key() >= num

jessemapel (approved):
Looks good. One small thing I'd tweak, but I think it's ready regardless. So, feel free to leave it as is and merge it.

isis/src/control/objs/ControlNetVitals/ControlNetVitals.cpp
krlberry: :+1:

PR #287, opened 2018-06-26
JigsawSetupDialog observation solve settings validation

ihumphrey:
This PR adds OK and Apply Settings button enabling/disabling based on validation of the position and pointing a priori sigma tables in the JigsawSetupDialog's observation solve settings tab.

Now, whenever there is *any* invalid value in either of the tables, the OK and Apply Settings buttons are disabled. When all of the sigmas are valid, the OK and Apply Settings buttons are (re)enabled.

Also, the default sizing of the pointing table matches the default sizing of the position table, and any whitespace is trimmed when checking for FREE validation in the sigma value.


twilson271828 (changes requested):
Looks good.  Add  a short history entry to JigsawSetupDialog.h and I will merge it.

twilson271828 (approved):

ihumphrey:
@twilson271828 Thanks for reviewing it! I've added a history entry and fixed conflicts with the histories

PR #292, opened 2018-06-26
Fixed ControlNet not handling ignoring measures correctly

jessemapel:
This fixes the errors in cnetwinnow. I also added the edge strength to the graphToString method in ControlNet because it's very helpful for debugging and ensuring the integrity of the graph.

### DIFF
--- a/isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
+++ b/isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
@@ -311,7 +311,6 @@ void IsisMain() {
         measFlag[j] = true;  //test passed
     }
 
-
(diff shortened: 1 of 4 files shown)


krlberry (commented):
This isn't part of your new changes, for line 677, shouldn't that (the edge strength ++)  be inside of the if(edgeAdded) ?

krlberry (approved):
:+1: This all looks good!

jessemapel:
@kberryUSGS Increasing the edge strength should happen when adding a new edge and when the edge already exists.. So, it needs to stay outside of the if statement.

krlberry:
@jessemapel Yep, you're right that's the whole point of incrementing the edge, isn't it? :) Must have had a brain-o.

PR #295, opened 2018-06-26
Added support for additional apriori sigmas in updateBundleObservationSolveSettings

twilson271828:
Prior to this, the additional apriori sigmas were being read in by updateBundleObservationSolveSettings, but they weren't being sent to the BOSS object the function is modifying.  Now they are.  This required modifying BundleObservationSolveSettings::setInstrumentPositionSettings/setInstrumentPointingSettings and changing the function signature by adding a default parameter at the end.  The change is backwards compatible and will not break anything.


ihumphrey (commented):
looks good!

PR #296, opened 2018-06-26
Added number of islands to HealthMonitorWidget

krlberry:


AgoinsUSGS (changes requested):
So the ControlHealthMonitor::broken(), healthy(), and weak() methods are placeholder slots that set the network to those respective states to allow us to see the error messages. 

we need a QLabel somewhere in the interface that actually displays the dynamic number of islands like we do with point/image data in the respective tabs. 

My thought is that the island is graph stuff and should go into the graph tab, but visualization for the graph tab is outside of the scope of this sprint. 

The number of islands is already displayed in the broken error message in the ControlNetVitals class, so I'm not sure if we should let that represent the number of islands for now and then revisit how to visualize the islands when we do the graph tab next time we revisit this.


UPDATE
I have removed the broken(), weak(), healthy() slots from the widget in my latest PR

krlberry:
@AgoinsUSGS Thanks for the explanation. I'm just going to close this, since it no long / was never needed. Thanks!

PR #299, opened 2018-06-27
Cleanup Tasks for ipceCnetHealth

AgoinsUSGS:
Removed QPoint include in ControlMeasure.cpp
Removed commented out code in ControlNet::MeasureDeleted()

Documented ModType Enums in ControlPoint, Measure, Net.
Documented ControlNetVitals member variables.


krlberry (approved):
:+1:

PR #300, opened 2018-06-27
Changed IPCE JigsawRunWidget design according to new mockup

chrisryancombs:


isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.ui
```diff
@@ -330,69 +330,98 @@ Must be positive.</string>
...
+                   <widget class="QComboBox" name="maximumLikelihoodModel1ComboBox">
+                    <property name="sizePolicy">
+                     <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+                      <horstretch>0</horstretch>
+                      <verstretch>0</verstretch>
+                     </sizepolicy>
+                    </property>
+                    <item>
+                     <property name="text">
+                      <string>NONE</string>
```
ihumphrey, line 343: The elements of the *maximumLikelihoodModel1-3ComboBox's* can cause a PROGRAMMER ERROR to occur, depending on what the user selects.

If you go into the JigsawSetupDialog, select the "HUBER MODIFIED" from the 1st model, then hit OK. The following error occurs:

```
**PROGRAMMER ERROR** Unknown maximum likelihood model name HUBER MODIFIED.
```

This is related to the *MaxiumLikelihoodWFunctions::stringToModel(QString)* method that is called at *JigsawSetupDialog::bundleSettings()* (lines ~600 to ~615). The "HUBER MODIFIED" value does not have a valid conversion defined in *MaxiumLikelihoodWFunctions::stringToModel(QString)*.

You could fix this by changing "HUBER MODIFIED" to "HUBER_MODIFIED".

Another way you could fix it is to programmatically set the elements of the maximumLikelihoodModel comboboxes in the JigsawSetupDialog constructor. If you set the elements there, you can use enum defined in the MaximumLikelihoodWFunctions class to determine the string. That way, the values in these comboboxes are decoupled from JigsawSetupDialog.

You could do this in the constructor with something like:

```
JigsawSetupDialog::JigsawSetupDialog() {
...

// outlier rejection
...
QStringList model1;
model1.append("NONE");
model1.append(MaximumLikelihoodWFunctions::modelToString(Model::Huber);
model1.append(MaximumLikelihoodWFunctions::modelToString(Model::HuberModified);

QStringList model2(model1);
model1.append(MaximumLikelihoodWFunctions::modelToString(Model::Welsch);
model1.append(MaximumLikelihoodWFunctions::modelToString(Model::Chen);

QStringList &model3 = model2;

m_ui->maximumLikelihoodModel1ComboBox.insertItems(0, model1);
m_ui->maximumLikelihoodModel2ComboBox.insertItems(1, model2);
m_ui->maximumLikelihoodModel2ComboBox.insertItems(2, model3);

File: isis/src/qisis/objs/JigsawRunWidget/JigsawRunWidget.cpp [hunk 2]
```diff
@@ -267,14 +259,135 @@ namespace Isis {
     }
     else {
       // Make sure to abort the bundle if it is currently running.
-      m_bundleAdjust->abortBundle();
-      m_bRunning = false;
       m_ui->JigsawRunButton->setText("&Aborting...");
+      m_ui->statusOutputLabel->setText("Aborting...");
```
ihumphrey, line 263: You may have already asked or talked to Ken about this, but do we want to discard the output on the left side if we abort a bundle.

For example, if a user is running a bundle and they see something in the left hand output where they want to abort the bundle, it will change to "aborting..." and discard the iteration outputs. Do we want that or do we want to keep the iteration outputs?

isis/src/qisis/objs/JigsawRunWidget/JigsawRunWidget.cpp
```diff
@@ -615,8 +641,30 @@ namespace Isis {
     m_bundleSolutionInfo = bundleSolutionInfo;
 
     // Since results are available, the user can accept (save) or reject(discard) the results.
-    m_accept->setEnabled(true);
-    m_reject->setEnabled(true);
+    m_ui->JigsawAcceptButton->setEnabled(true);
+  }
+
+
+  void JigsawRunWidget::closeEvent(QCloseEvent *event) {
```
ihumphrey: Make sure to add doxygen documentation for this method

ihumphrey (commented):
Looks good. There was one way i found to get an exception thrown (note that i did not thoroughly test everything), and i've documented it in the review.

PR #302, opened 2018-06-28
Removed partial derivative matrices as member variables and now decla…

kledmundson:
…re them as static in the applyConstraint method. Now coeff_range_image matrix is resized if number of image parameters changes. This is consistent with BundleAdjust::computePartials. Added IExceptions to verify m_rangeObserved and m_rangeComputed are positive.


dcookastro (approved):
The changes look good.

dcookastro:
Did you see any run time improvements from the changes?

PR #303, opened 2018-06-28
Added libs to env file that may be missing on some Linux machines 

Kelvinrr:
Found after testing from Trent. 

Unfortunately, one of the required libs libglu1 does not seem to exist on anaconda cloud, se we either need to create that recipe or simply require people to install it using  other package managers.


Kelvinrr:
for some reason this includes changes from #298 as if it was never merged?

PR #304, opened 2018-06-28
Added m_numPoints and m_numMeasures member variable to cnetvitals.

AgoinsUSGS:
Made local variables for these values rather than using the getters in ControlNet to reduce coupling with ControlNet and allow for accurate updates for these values.

NOTE: This branch is built off the cubeDn branch on my fork, in which a PR is currently up. That PR needs to be merged before this.


krlberry (commented):
:+1: Looks good except for the same comment as the last PR. (Why is the name for the new QAction "Sup"?)

AgoinsUSGS:
The random QAction has been removed

krlberry (approved):
:+1:

PR #309, opened 2018-06-29
Made ControlPoint::operator= call setters, ensure points aren't ignored when grabbing them.

AgoinsUSGS:
ControlPoint::operator= now calls setters when copying data from a Control Point so that the proper signals/slots get called.
ControlNetVitals now doesn't include ignored points when grabbing specific point lists (points constrained, < 3 measures, etc.)

NOTE: It looks like this branch needs to have ipceCnetHealth merged into it before merging this PR.


isis/src/control/objs/ControlPoint/ControlPoint.cpp
```diff
@@ -1945,23 +1955,26 @@ namespace Isis {
...
+      SetIgnored(other.ignore);
+      SetAprioriSurfacePointSource(other.aprioriSurfacePointSource);
+      SetAprioriSurfacePointSourceFile(other.aprioriSurfacePointSourceFile);
+      SetAprioriRadiusSource(other.aprioriRadiusSource);
+      SetAprioriRadiusSourceFile(other.aprioriRadiusSourceFile);
+      SetAprioriSurfacePoint(other.aprioriSurfacePoint);
+      SetAdjustedSurfacePoint(other.adjustedSurfacePoint);
+
+      // Set edit lock last so the it doesn't interfere with copying the other fields over.
+      SetEditLock(other.editLock);
```
jessemapel: You also need to reset to the old edit lock status before setting the edit lock. As it's currently implemented, if the point was locked and now is unlocked, we wouldn't notify because prior to this call the point is unlocked, not locked.

jessemapel (changes requested):
This either needs to be PR'd into ipceCnetHealth or the changes need to be rebased from your ipceCnetHealth fork onto ipce. We cannot merge ipceCnetHealth into ipce until it is clean.

jessemapel:
We also need to add signals to the ControlMeasure assignment operator in the same way.

PR #312, opened 2018-06-29
addEdge(), removeEdge(), and documentation added to ControlNet

krlberry:


isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -554,16 +602,7 @@ namespace Isis {
...
-            // If the edge already exists, this just returns it. (The use of a set
-            // forces the edges to be unique.)
-            ImageConnection connection;
-            bool edgeAdded;
-            boost::tie(connection, edgeAdded) = boost::add_edge(m_vertexMap[serial],
-                                                       m_vertexMap[sn],
-                                                       m_controlGraph);
-            m_controlGraph[connection].strength++;
-
+            bool edgeAdded = addEdge(serial, sn);
```
krlberry, line 615: For whatever reason, I thought doing it this way was clearer than:

```
if( addEdge(serial, sn) ){
  emit networkModified(GraphModified);
}
```
Let me know if you prefer this way, as I wasn't completely sure which would be better.

jessemapel: I think it's better to include the signals in the addEdges and removeEdge methods. That way we always notify when the graph changes.

jessemapel (changes requested):

isis/src/control/objs/ControlNet/ControlNet.cpp
krlberry: Agreed, but -- putting the signals in the methods changes the unit test output for `ControlNet` which... seems incorrect. I'll call you on hangouts about this later.

jessemapel: sounds good

jessemapel (approved):

PR #313, opened 2018-06-29
(DO NOT MERGE; REVIEW ONLY) colors set in BOSS tab

ihumphrey:
This probably needs to be redesigned.

Feel free to look at the implementation. Also, *make sure to pull down these changes and test them for how the functionality is working in the observation solve settings tab*. Note that a maximum of 7 colors is used for this initial proof-of-concept.


isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.cpp
```diff
@@ -1832,15 +1831,90 @@ namespace Isis {
...
+
+    // Change color for new solve settings
+    QColor newColor(Qt::white);
+
+    QColor red(Qt::red);
+    qDebug() << "red hex: " << red.name();
+    red.setNamedColor("#CCCCCC");
+    qDebug() << "red hex2: " << red.name();
+
+    QSet<QString> availableColors = colorSet.subtract(existingColorNames);
```
chrisryancombs: Right now we are limited to seven BOSS colors, but we could expand this to rotate over the colors. Maybe you could maintain an index for the set and modulo by the length to loop over it as many times as you want.

We could combine the current "base color" with an offset to randomly create unique colors.
Here is a link to some good ideas on color generation: http://devmag.org.za/2012/07/29/how-to-choose-colours-procedurally-algorithms/. The offset color generation is under the Algorithms section, in the random offset subsection.

With this idea we would loop through the color set, creating unique colors in a set order. For example, if I had three hues in my "badColorSet": 

Red -> Blue -> Yellow -> Red w/ offset 1 -> Blue w/ offset 1 -> Yellow w/ offset 1 -> Red w/ offset 2...

celbal: That's a cool idea! But I think that might be a stretch goal for now.

isis/src/qisis/objs/SortFilterProxyModel/SortFilterProxyModel.cpp
```diff
@@ -90,6 +90,39 @@ namespace Isis {
  }
 
 
+  QVariant SortFilterProxyModel::data(const QModelIndex &index, int role) const {
+    if (!index.isValid()) return QVariant();
+
```
ihumphrey, line 103: something I noticed when running this, if we want the proxy model to show the icons (the little hot-air balloon icons next to images on the project tree), we weill need to add a check for the _Qt::DecorationRole_ and call the QSortFilterProxyModel::data function (like we do for the Qt::DisplayRole)

isis/src/qisis/objs/SortFilterProxyModel/SortFilterProxyModel.cpp
```diff
@@ -90,6 +90,39 @@ namespace Isis {
  }
 
 
+  QVariant SortFilterProxyModel::data(const QModelIndex &index, int role) const {
```
ihumphrey, line 101: we should document this method

isis/src/qisis/objs/ProjectItemModel/ProjectItemModel.cpp
```diff
@@ -912,6 +912,23 @@ namespace Isis {
   }
 
 
+  QVariant ProjectItemModel::data(const QModelIndex &index, int role) const {
```
ihumphrey, line 925: we should document this method

isis/src/qisis/objs/ProjectItemModel/ProjectItemModel.cpp
```diff
@@ -912,6 +912,23 @@ namespace Isis {
...
+    if (index.isValid())
+      if (role == Qt::BackgroundRole || role == Qt::ForegroundRole) return QVariant();
```
ihumphrey: we don't really need the Qt::ForegroundRole check here. we might want to add a comment about what this check is doing as well (something like _this causes any view that uses this model directly to ignore the background colors being set through setData() with Qt::BackgroundRole. this is used so the SortFilterProxyModel can set background colors and have them displayed through attached views (the observation solve settings tree view in the JigsawSetupDialog)._

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.h
```diff
@@ -189,10 +191,18 @@ namespace Isis {
...
+    QColor generateBOSSColor();
+
+    // void updateGroups(QList<QStandardItem *> &selectedItems);
+    QList<QBrush *> m_colors;
+
   private:
     Ui::JigsawSetupDialog *m_ui;
     Project *m_project;
     BundleSettingsQsp m_bundleSettings; /**< The BundleSettings Object created by this dialog */
+    int m_colorIter;
```
ihumphrey: add documentation about this member

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.h
```diff
@@ -189,10 +191,18 @@ namespace Isis {
...
+
+    QList<QColor> m_baseColors;
```
ihumphrey: add documentation about this member

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.h
```diff
@@ -189,10 +191,18 @@ namespace Isis {
 
     void createObservationSolveSettingsTreeView();
 
+    QColor generateBOSSColor();
+
+    // void updateGroups(QList<QStandardItem *> &selectedItems);
+    QList<QBrush *> m_colors;
```
ihumphrey: i would remove _m_brush_, as i don't think it is being used anywhere anymore.

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.cpp
```diff
@@ -1729,6 +1762,30 @@ namespace Isis {
   }
 
 
+  QColor JigsawSetupDialog::generateBOSSColor() {
```
ihumphrey, line 1780: add method documentation

isis/src/qisis/objs/JigsawSetupDialog/JigsawSetupDialog.cpp
```diff
@@ -99,10 +101,12 @@ namespace Isis {
     if (!m_project->directory()->model()->selectedItems().isEmpty()) {
       foreach (ProjectItem * projItem, m_project->directory()->model()->selectedItems()) {
         if (projItem->isImage()) {
+          projItem->setData(QVariant(QColor(Qt::transparent)), Qt::UserRole+10);
```
ihumphrey: **_(this isn't something you need to do, i just thought of this as an idea that could possibly generalize our UserRoles)._**

Something we could do regarding the _Qt::UserRole+10_ is moving it to a public enum inside of ProjectItemModel.

```c++
// In ProjectItemModel header
enum UserRole {
  BundleObservationSolveSettingsColorRole = Qt::UserRole+2;
};
```

Then, in in JigsawSetupDialog.cpp and SortFlterProxyModel.cpp, we could replace the Qt::UserRole+1 with ProjectItemModel::BundleObservationSolveSettingsColorRole (which is long, the name could be anything meaningful for that role).

Alternatively, we could add the enum to a new header, like UserRoles.h, and put it in an UserRole namespace under the Isis namespace; or a globals header (Isis.h), but I'm not sure what the implications of that would be.

I believe that ProjectItemModel/ProjectItem already implicitly defines a data role, Qt::UserRole+1, in as a default parameter for ProjectItemModel::findItemData(..., role=Qt::UserRole+1) and ProjectItem::findItemData(..., role=Qt::UserRole+1). We could add a _DataRole = Qt::UserRole+1_ to the ProjectItemModel UserRole enum and replace the default role parameter values for findItemData with ProjectItemModel::DataRole;

isis/src/control/objs/BundleUtilities/BundleObservationSolveSettings.h
```diff
@@ -206,6 +210,8 @@ class BundleObservationSolveSettings {
           QStringList m_xmlHandlerAprioriSigmas;
       };
 
+    private:
+      QColor m_color;
```
ihumphrey: add documentation for this member

isis/src/control/objs/BundleUtilities/BundleObservationSolveSettings.cpp
```diff
@@ -1055,4 +1074,14 @@ namespace Isis {
     }
     return XmlStackedHandler::endElement(namespaceURI, localName, qName);
   }
+
+
+  QColor BundleObservationSolveSettings::color() const {
```
ihumphrey, line 1104: add method documentation

isis/src/control/objs/BundleUtilities/BundleObservationSolveSettings.cpp
```diff
@@ -1055,4 +1074,14 @@ namespace Isis {
...
+    return m_color;
+  }
+
+
+  void BundleObservationSolveSettings::setColor(const QColor &color) {
```
ihumphrey, line 1114: add method documentation

ihumphrey:
i tested the changes with the multiple sensor data set:

```
ipce
import images -> /work/users/ihumphrey/BOSSColors/multipleSensors.lis
import control net -> /work/users/ihumphrey/BOSSColors/AS15_LandingSide_Metric-NAC-LO5H-FIN.net
project -> bundle adjustment...
click "setup" on jigsaw run dialog
go to observation solve settings tab
expand import1 node
select first three images (the 5106 images)
apply settings
```

Only the first 5106 image will have a background color. These 5106 images are all part of one observation, so that may be the issue?

**Edit:** _it is not an issue with the application of the solve settings (they are set for all three of the images), its just the color is only set for the first one)_

celbal:
This is being closed as the branch this would be pulled into is being deleted.

jessemapel:
We may want this in the future. reopening for now

PR #314, opened 2018-06-30
Added population of JSD BOSS tab with selected images settings

chrisryancombs:


chrisryancombs:
Added this PR just in case the weekend people need it

celbal:
Using your changes when I select an image on the BOSS tab the ANGLES row on the Instrument Pointing Solve Settings is set to a special pixel when it was 0.0 before.

My data: /work/users/mshepherd/bundleReady

Let me know if this does not make sense

celbal:
The default table loading is not working correctly when you select an image. 

The first table should show nothing as the default setting is NONE. The second table should show only the ANGLES row. This happens correctly when there is not an image selected.

As soon as you select an image all of the rows show up and then weirdness happens because only the setting that is selected from the drop down in saved. It's very confusing if you don't know what's going on.

celbal:
Whoops did not mean to close that. Sorry!

PR #316, opened 2018-07-01
Switch definitions of PDS4 functions to match names. 

krlberry:


isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
```diff
@@ -450,9 +450,9 @@ namespace Isis {
   }
 
 
-  void ProcessExportPds4::translateBandBinSpectrumUniform(Pvl &inputLabel) {
+  void ProcessExportPds4::translateBandBinSpectrumBinSet(Pvl &inputLabel) {
```
makaylas: Can these be combined into one function? And all you would have to do is pass in the name of the translation file.

That way you don't have duplicate code. (This is super minor)

PR #320, opened 2018-07-02
Ipce bundle window speedup setup

ihumphrey:
This is solution 2 (adding members to Image for serialNumber and observationNumber, then storing their values after the first call to serialNumber() / observationNumber() )


krlberry (approved):
:+1: This looks great!

PR #325, opened 2018-07-06
Tweaked ControlNetVitals to keep island computation and counting in sync

jessemapel:
The valid measures in image counting is currently not working correctly, but there is not time to fix it right now so it's been documented for work next year.


krlberry (approved):
:+1: Looks good! I'm glad you moved the network modified signals into the `AddEdge` and `RemoveEdge` functions, it does make more sense in there. When I tried that (but without your other changes) the number of islands detected in the `ControlNet` unitTest changed. That's not happening with this set of changes, right?

krlberry:
That problem was fixed.

PR #329, opened 2018-07-06
Fixed ambiguous shortcuts that occured after conversion to QMainWindow.

kaitlyndlee:
Actions/widgets will be enabled/disabled when a user enters/leaves a view. I also cleaned up IpceMainWindow again.


SgStapleton:
Everything looks good and looks like testing has been completed.

PR #331, opened 2018-07-07
Removed old set target call with radii from ControlNet

jessemapel:


dcookastro (approved):
Will removing the mutex block cause the problem Tracie had with importing two control nets in IPCE to return? The rest looks good except for a comment in the ControlNet unitTest that refers to the radii.  I can fix the comment.

jessemapel:
@dcookastro Removing the mutex won't cause any problems because it doesn't wrap a SPICE call anymore.

PR #333, opened 2018-07-09
Changed the behavior of BundleSettings::observationSolveSettings(...)

twilson271828:
In the BOSS selection tree of the JSD , if you select a second group of images after applying BOSS settings to an initial group of images,  an error comes up that says "Unable to find BundleObservationSolveSettings."  and the GUI will not let you set BOSS settings for the second group of images.  This PR fixes that by changing the behavior of  BundleSettings::observationSolveSettings(QString observationNumber) by just returning a default BOSS object instead of throwing an exception.

### DIFF
--- a/isis/src/control/objs/BundleSettings/BundleSettings.cpp
+++ b/isis/src/control/objs/BundleSettings/BundleSettings.cpp
@@ -450,14 +450,17 @@ namespace Isis {
   BundleObservationSolveSettings
       BundleSettings::observationSolveSettings(QString observationNumber) const {
 
+    BundleObservationSolveSettings defaultSolveSettings;
+
     for (int i = 0; i < numberSolveSettings(); i++) {
       if (m_observationSolveSettings[i].observationNumbers().contains(observationNumber)) {
         return m_observationSolveSettings[i];
       }
     }
-    QString msg = "Unable to find BundleObservationSolveSettings for observation number ["
-                  + observationNumber + "].";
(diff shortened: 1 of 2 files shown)


isis/src/control/objs/BundleSettings/BundleSettings.cpp
```diff
@@ -450,14 +450,17 @@ namespace Isis {
   BundleObservationSolveSettings
       BundleSettings::observationSolveSettings(QString observationNumber) const {
 
+    BundleObservationSolveSettings defaultSolveSettings;
```
chrisryancombs, line 453: Do we want to return a default here? I think that it would make more sense to return NULL when a BundleSettings object doesn't have a BOSS corresponding to the given observation number.

chrisryancombs: If you  open the setup dialog with a new group of images and hit OK, the images wont have these defaults applied to them. Maybe a better solution would be to create a constructor for JigsawSetupDialog that takes in a BundleSettings object. In this constructor, we could add the case for images that were not part of the original setup.

chrisryancombs: The handling for this case could also be added to the loadSettings method, if a constructor seems overkill.

chrisryancombs:
I added the fix I described above. this solution should work with the change you already made.

twilson271828 (commented):
Definitely an improvement over my first solution.

PR #335, opened 2018-07-10
Ipce docks templates

SgStapleton:
This includes the new QComboBox implementation in ControlPointEditWidget, the new SaveAs option in TemplateEditorWidget, and the separation of map and registration template files within the project.


isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.h
```diff
@@ -231,7 +237,7 @@ namespace Isis {
       QPointer<QWidget> m_templateEditorWidget; //!< Template editor widget
       bool m_templateModified; //!< Indicates if the registration template was edited
 
-      QPointer<QLabel> m_templateFileNameLabel; //!< Label for the template filename
+      QPointer<QComboBox> m_templateComboBox; //!< ComboBox of importd registration templates
```
krlberry: Single character typo: Could you change `importd` to `imported` in the comment here?

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -2386,6 +2409,59 @@ namespace Isis {
...
+  /**
+  * Appends the filename to the registrations path (unless this is the default template) and calls 
+  * setTemplateFile for the control measure
+  *
+  * @param filename This is the import directory and the filename of the template file selected from
+  *                 the QComboBox.
+  */
+  void ControlPointEditWidget::setTemplateFile(QString filename) {
+    QString expandedFileName = filename;
+    if(!filename.startsWith("$base")){
```
krlberry, line 2436: Will the template file ever be in a different area non-project area like `$hayabusa`?

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -2386,6 +2409,59 @@ namespace Isis {
...
+  
+  
+  /**
+  * Reset the selected template in the template combobox if the template selected by the user does 
+  * not satisfy requirements for the control measure.
+  *
+  * @param fielName The filename that was previously selected in the template combo box
+  */
+  void ControlPointEditWidget::resetTemplateComboBox(QString fileName) {
+    if(fileName.startsWith("$base")) {
```
krlberry: (same comment as above)

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -2386,6 +2409,59 @@ namespace Isis {
...
+      loadTemplateFile(expandedFileName);
+    }
+  }
+  
+  
+  /**
+  * Reset the selected template in the template combobox if the template selected by the user does 
+  * not satisfy requirements for the control measure.
+  *
+  * @param fielName The filename that was previously selected in the template combo box
```
krlberry: fielName -> fileName

isis/src/qisis/objs/Directory/Directory.cpp
```diff
@@ -902,9 +907,6 @@ namespace Isis {
 
     connect( result, SIGNAL( destroyed(QObject *) ),
              this, SLOT( cleanupMatrixViewWidgets(QObject *) ) );
-
```
krlberry: Why were these removed? (I'm basically just double-checking to make sure that these were) deliberately removed. There's a lot of work in this PR and I couldn't figure out what other change caused this one.)

krlberry (commented):
Looks good! A few very minor typos I found, and a question: 
For all of the lines in which there's something of the form 
value == "string" 
I just want to double-check on use of the case-sensitive test for equality -- will any of these values ever be capitalized (or in any way have a different case)?

isis/src/qisis/objs/Directory/ImportMapTemplateWorkOrder.cpp
```diff
@@ -107,47 +107,25 @@ namespace Isis {
...
     QStringList templateFileNames;
 
     templateFileNames = QFileDialog::getOpenFileNames(
         qobject_cast<QWidget *>(parent()),
-        "Import " + itemType,
+        "Import Map Templates",
         QString(),
-        filterText,
-        &selectedFilter);
+        "Maps (*.map);; All Files (*)");
```
krlberry, line 120: Is it valid to have a mapfile with a filename like `map.pvl`? If not, this makes sense.

isis/src/qisis/objs/Directory/Directory.cpp
SgStapleton: Yeah, sorry - this was an unrelated fix. These connections were not needed.

isis/src/qisis/objs/Directory/ImportMapTemplateWorkOrder.cpp
SgStapleton: It is valid. I guess I am not sure of the concern here...? (If it is about the "Maps (*.map)..." line, this is simply a filter of the qfiledialog box. They can certainly get to .pvl files by choosing the "All Files" filter.)

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
SgStapleton: Not as of right now. The default template file in $base is hard-coded. The only other files that will exist as an option in this combobox are imported files. If other files are added in the future, this logic will need to change. Do you think this is reasonable now, or should I find another way of distinguishing?

SgStapleton:
All instances of
value == "string"
currently will not have any other capitalization scheme and so do not need to be made case-insensitive.

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
krlberry: No, you don't need to change anything. If `$base` is hard coded for the template file area, the rest of this makes sense, and I think that it's okay that if this changes in the future, the other logic will need to be updated, too.

isis/src/qisis/objs/Directory/Directory.cpp
krlberry: No problem! Just double-checking.

isis/src/qisis/objs/Directory/ImportMapTemplateWorkOrder.cpp
krlberry: Then, no problem! I just wasn't picturing this correctly. Thanks!

krlberry:
@SgStapleton :+1: Thanks! This looks good, so I'll go ahead and merge it.

PR #336, opened 2018-07-10
Removes the parent widget of a cneteditorview when discarding changes and changing an active control

kaitlyndlee:
When a user makes changes to an active control, but wants to change the active control and discards these changes, the old CnetEditorView used to not be deleted completely. Also, if a ControlPointEditView is open and a user discards changes, the view will close.

### DIFF
--- a/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
+++ b/isis/src/qisis/apps/ipce/IpceMainWindow.cpp
@@ -173,7 +173,7 @@ namespace Isis {
     initializeActions();
     createMenus();
     createToolBars();
-    
+
     QStringList args = QCoreApplication::arguments();
 
     if (args.count() == 2) {
@@ -232,8 +232,11 @@ namespace Isis {
    * @param view QWidget* The view to close.
    */
   void IpceMainWindow::removeView(QWidget *view) {
-    view->close();
+    QDockWidget *parentDock = qobject_cast<QDockWidget *>(view->parent());
+    removeDockWidget(parentDock);
+    m_viewDocks.removeAll(parentDock);
     delete view;
+    delete parentDock;
   }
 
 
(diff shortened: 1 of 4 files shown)


isis/src/qisis/apps/ipce/IpceMainWindow.h
```diff
@@ -159,9 +159,10 @@ namespace Isis {
    *                           state can be reset after the IpceMainWindow::show() causes resize and
    *                           move events which in turn cause the project clean flag to be false
    *                           even though the project has just opened.
-   *   @history 2018-07-05 Kaitlyn Lee - Added tilViews() and the menu option to tile all docked/undocked
-   *                           and tabbed/untabbed views.
-
+   *   @history 2018-07-09 Kaitlyn Lee - Added tilViews() and the menu option to tile all docked/undocked
```
krlberry: Should `tilViews()` be `tileViews()` ?

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
```diff
@@ -225,8 +225,11 @@ namespace Isis {
    * @param view QWidget* The view to close.
    */
   void IpceMainWindow::removeView(QWidget *view) {
-    view->close();
+    QDockWidget *parentDock = qobject_cast<QDockWidget *>(view->parent());
```
krlberry, line 228: Is there always a parent dock widget when this is called? Is `view->close()` ever still necessary to call?

krlberry (commented):
One typo and a question that is probably me just not understanding how this works.

isis/src/qisis/apps/ipce/IpceMainWindow.cpp
kaitlyndlee: Yes, there will always be a parent dock widget because a view is stored inside of a QDockWidget when it is opened. When we call removeDockWidget(), this closes the dock widget and its children, including the view.  So, we do not need to call view->close() as well.

krlberry: :+1: Thanks!

PR #337, opened 2018-07-10
selectedBOSSImages() will return a list of all images and image lists if no images are selected in the project tree

kaitlyndlee:


twilson271828 (changes requested):
Looks good, Kaitlyn.  My only request is that you add
a return items; line before you exit the if statement.  There's no need to go through the rest of the function at this point.

twilson271828 (approved):

PR #342, opened 2018-07-12
remove unused ProjectItemImageProxyModel class

ihumphrey:
This was a placeholder class added during ipceBundleWindow work that never got used, so it can be safely removed.


SgStapleton (approved):
Simple deletion

PR #347, opened 2018-07-12
Fix to segfault on astrovm4

SgStapleton:


isis/src/base/objs/Camera/Camera.h
```diff
@@ -328,6 +331,8 @@ namespace Isis {
       CameraDetectorMap *DetectorMap();
       CameraGroundMap *GroundMap();
       CameraSkyMap *SkyMap();
+      
+      QString instrumentId(); // The "InstrumentId" as it appears on the cube
```
kaitlyndlee: Do we need this comment, since m_instrumentId has the same comment?

isis/src/qisis/objs/GuiCamera/GuiCamera.h
```diff
@@ -82,7 +84,7 @@ namespace Isis {
 
       QString id() const;
 
-//    Camera *camera();
+      QString instrumentId(); // The "InstrumentId" as it appears on the cube
```
kaitlyndlee: Do we need this comment, since m_instrumentId has the same comment?

kaitlyndlee (commented):
Two quick things to consider, but other than that, looks good!

PR #361, opened 2018-07-18
DO NOT MERGE - TrackingTable Addition

SgStapleton:
This adds the TrackingTable object for the re-working of how tracking will be handled with an external tracking cube.


jessemapel:
@SgStapleton Can you PR this onto our mosaic branch so that other people can work with the API at least?

isis/src/base/objs/TrackingTable/TrackingTable.h
```diff
@@ -0,0 +1,67 @@
...
+   * @author 2004-09-01 Jeff Anderson
+   *
+   * @internal
+   *   @history 2005-03-18 Elizabeth Ribelin - Added documentation to class
+   */
+  class TrackingTable{
+    
+    public:
+      
+      static QString tableName = "InputImages";
```
jessemapel, line 45: This needs to be moved outside of the class and we should rename it to something like trackingTableName because then it will be accessible as "Isis::trackingTableName" similar to how special pixel values can be accessed.

It also needs to be const not static, my bad!

jessemapel: See SpecialPixel.h for how this should be done.

PR #362, opened 2018-07-19
Added 4-byte unsigned integer special pixel vals to  SpecialPixel.h

twilson271828:
I am emulating what has already been done for 2-byte unsigned shorts to maintain some consistency (even though splitting up the bytes between high and low values is weird).

### DIFF
--- a/isis/src/base/objs/Blobber/Blobber.h
+++ b/isis/src/base/objs/Blobber/Blobber.h
@@ -108,6 +108,8 @@ namespace Isis {
    *                           coding standards. Added padding to control
    *                           statements. References #1169.
    *   @history 2017-08-30 Summer Stapleton - Updated documentation. References #4807.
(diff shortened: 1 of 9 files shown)


isis/src/base/objs/Blobber/Blobber.h
```diff
@@ -262,6 +264,7 @@ namespace Isis {
 //  Low/level I/O and conversion methods
       void loadDouble(Table &tbl);
       void loadInteger(Table &tbl);
+      double int2ToDouble(unsigned int value) const;
       double int2ToDouble(int value) const;
```
jessemapel, line 268: Why are these taking 4 byte ints when the method name says 2 byte? Should these take short ints and then we have separate functions for 4 byte ints?

jessemapel (approved):

PR #368, opened 2018-07-19
New TrackingTable object

SgStapleton:


isis/src/base/objs/TrackingTable/TrackingTable.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "Table.h"
+#include "TableField.h"
+#include "TableRecord.h"
+
+using namespace std;
+namespace Isis {
+
+
+  /**
+  * @brief Default constructor
```
jessemapel: It doesn't matter, but the @brief tag is not really needed. Doxygen will pull the first sentence for the brief documentation when @brief isn't used.

isis/src/base/objs/TrackingTable/TrackingTable.cpp
```diff
@@ -0,0 +1,165 @@
...
+  * @ brief Constructs a TrackingTable given a Table object. The Table is used to populate 
+  *         m_fileList.
+  *
+  * @param table The Table object to pull the filenames and serial numbers from
+  */
+  TrackingTable::TrackingTable(Table table) {
+    
+    for (int i=0; i < table.Records(); i++) {
+      TableRecord record = table[i];
+      QString nameField = QString(record["FileName"]).toLatin1().data();
```
jessemapel: We don't need the toLatin1 and data calls here, they convert a QString into a Char*. Then, the QString assignment operator converts it back to a QString.

isis/src/base/objs/TrackingTable/TrackingTable.h
```diff
@@ -0,0 +1,74 @@
...
+   * indices in this table. Each record within this table will contain the filename of an
+   * associated cube, that cube's serial number, and the DN value associated with this cube within
+   * the tracking band (which should also be its index in the table).
+   *
+   * @author 2018-07-19 Jesse Mapel & Summer Stapleton
+   *
+   * @internal
+   */
+   
+  const QString tableName = "InputImages";
```
jessemapel: Because this variable is now a part of the Isis namespace, it should have a more descriptive name. Outside of this class, "tableName" could mean anything. It should be changed to something to like "trackingTableName"

isis/src/base/objs/TrackingTable/TrackingTable.h
```diff
@@ -0,0 +1,74 @@
...
+            
+      TrackingTable();
+      
+      TrackingTable(Table table);
+      
+      ~TrackingTable();
+      
+      Table toTable();
+      
+      // oldDnToNewDn(); TODO (maybe?)
```
jessemapel: I think we decided not to do this here, so this should be removed

isis/src/base/objs/TrackingTable/TrackingTable.h
```diff
@@ -0,0 +1,74 @@
...
+   * This table will currently be stored in the label of a separate cube. This tracking cube will
+   * also contain a single tracking band. The DN values stored in this band will correlate to the
+   * indices in this table. Each record within this table will contain the filename of an
+   * associated cube, that cube's serial number, and the DN value associated with this cube within
+   * the tracking band (which should also be its index in the table).
+   *
+   * @author 2018-07-19 Jesse Mapel & Summer Stapleton
+   *
+   * @internal
+   */
```
jessemapel, line 49: This documentation needs to be moved down to just before the class declaration

isis/src/base/objs/TrackingTable/unitTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+  cout << endl;
+
+  cout << "Testing the toTable method ..." << endl;
+  
+  Table tableOut = trackingTable.toTable();
+  
+  cout << "First record: " << QString(tableOut[0][0]) << ", " << QString(tableOut[0][1]) << endl;
+  cout << "Second record: " << QString(tableOut[1][0]) << ", " << QString(tableOut[1][1]) << endl;
+  cout << "Third record: " << QString(tableOut[2][0]) << ", " << QString(tableOut[2][1]) << endl;  
+  cout << "Fourth record: " << QString(tableOut[3][0]) << ", " << QString(tableOut[3][1]) << endl;
```
jessemapel: We should also test that we can re-ingest the output Table object and it's the same object.

jessemapel (changes requested):
Small changes plus some extra testing stuff:

Should test constructing from a table with a filename that doesn't end in .cub

Test writing to a Table object when the longest string is a serial number

isis/src/base/objs/TrackingTable/TrackingTable.h
SgStapleton: Done

SgStapleton:
All changes made.

isis/src/base/objs/Cube/Cube.h
```diff
@@ -221,9 +221,10 @@ namespace Isis {
          * |     *     *     *     *    |*
          * |****************************|*
          * |Tile *Tile *Tile *Tile *Tile|*
+         * | 10  * 11  * 12  * 13  * 14 |*
+         * |     *     *     *     *    |
          * ------------------------------*
-         * * 10  * 11  * 12  * 13  * 14  *
-         * *******************************
+
```
SgStapleton: This was just a typo in the comments that was really bugging me. I'd seen it a couple times and decided it was time to fix it

jessemapel (approved):
Looks good!

PR #369, opened 2018-07-20
A few minor tweaks to documentation tags which were causing PHP warnings in the createReleaseNotesXML script.

twilson271828:
There is also a change to SpecialPixel.h which I committed by mistake and undid by checking out the original from upstream.  Hopefully this fix will take care of the remaining PHP warnings in the createReleaseNotesXML script.


ihumphrey (dismissed):
Looks good, I'm a little confused with the ProjectItemImageProxyModel class being there though, it should have been deleted in commit 4eaafdbb06f3e61539c020ecdd8ddead380f8150

twilson271828:
I've deleted both files from this PR.  I think the original PR was prior to the commit you reference, and I forgot to pull from upstream.

ihumphrey (approved):

PR #372, opened 2018-07-23
[DO NOT MERGE] Modified ProcessMosaic to write to a tracking cube instead of a tracking band.

jessemapel:
I'm looking for some feedback on this. I can't figure out why it is not actually writing to the tracking cube. When I debug through the unit test and stop after it finishes the first processing the mosaic is correct, but the tracking cube is still all NULLs.

### DIFF
--- a/isis/src/base/objs/ProcessMosaic/ProcessMosaic.cpp
+++ b/isis/src/base/objs/ProcessMosaic/ProcessMosaic.cpp
@@ -50,6 +50,7 @@ namespace Isis {
 
     // Initialize the structure Track Info
     m_trackingEnabled    = false;
+    m_trackingCube = NULL;
     m_createOutputMosaic   = false;
     m_bandPriorityBandNumber  = 0;
     m_bandPriorityKeyName  = "";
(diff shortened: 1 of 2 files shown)


isis/src/base/objs/ProcessMosaic/ProcessMosaic.cpp
```diff
@@ -264,8 +268,44 @@ namespace Isis {
...
+            QString trackingPath = FileName(OutputCubes[0]->fileName()).path();
+            QString trackingFile = OutputCubes[0]->group("Tracking").findKeyword("FileName")[0];
+            m_trackingCube->create(trackingPath + trackingFile);
+
+            // TODO INITIALIZE TRACKING TABLE FROM TRACKING CUBE
+          }
+          else {
+            QString msg = "Tracking cannot be enabled when adding to an existing mosaic "
+            "that does not already have a tracking cube. Mosaics with a tracking band must "
+            "be have the tracking band extracted into an external tracking cube.";
```
SgStapleton, line 304: "must be have"

PR #374, opened 2018-07-24
Created program to extract the tracking band from a mosaic

kaitlyndlee:


isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+void findTrackBand(UserInterface ui, QString &copyBands, QString &trackBand);
+void createMosaicCube(UserInterface ui, QString bands);
+void createTrackCube(UserInterface ui, QString trackingBand);
+void copyPixels(Buffer &in, Buffer &out);
+void copyTrackPixels(Buffer &in, Buffer &out);
+
+void IsisMain() {
+  UserInterface &ui = Application::GetUserInterface();
+  QString copyBands;
+  QString trackBand;
```
jessemapel: It feels really weird passing a list of integers around as a string. I'd much rather see these as lists of integers

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (cubeObject.hasGroup("BandBin")) {
+      PvlGroup &bandBinGroup = cubeObject.findGroup("BandBin");
+      PvlKeyword &currentKeyword = bandBinGroup[0];
+      for (int i = 0; i < currentKeyword.size(); i++) {
+        if (currentKeyword[i] != "TRACKING") {
+          copyBands += QString::number(i + 1); // Make it 1 based
+          copyBands += ",";
+        }
+        else {
+          trackBand = QString::number(i + 1);
```
jessemapel: You don't need to return both the tracking bands and the non-tracking bands. You can just return the tracking band index and then any band that isn't that index is a non-tracking (copy) band.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+ */
+void findTrackBand(UserInterface ui, QString &copyBands, QString &trackBand) {
+  Cube inputCube = Cube(ui.GetFileName("FROM"));
+  Pvl *inputLabel = inputCube.label();
+
+  if (inputLabel->hasObject("IsisCube")) {
+    PvlObject &cubeObject = inputLabel->findObject("IsisCube");
+    if (cubeObject.hasGroup("BandBin")) {
+      PvlGroup &bandBinGroup = cubeObject.findGroup("BandBin");
+      PvlKeyword &currentKeyword = bandBinGroup[0];
```
jessemapel: This will error if the BandBin group has no keywords. I doubt that's something that will happen normally, but you may want to handle that scenario with a more descriptive error message than the PvlGroup index out of bands message.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+ * @param trackBand Index of the tracking band
+ */
+void findTrackBand(UserInterface ui, QString &copyBands, QString &trackBand) {
+  Cube inputCube = Cube(ui.GetFileName("FROM"));
+  Pvl *inputLabel = inputCube.label();
+
+  if (inputLabel->hasObject("IsisCube")) {
+    PvlObject &cubeObject = inputLabel->findObject("IsisCube");
+    if (cubeObject.hasGroup("BandBin")) {
+      PvlGroup &bandBinGroup = cubeObject.findGroup("BandBin");
```
jessemapel: You can skip to here using `Cube::hasGroup` and `Cube::group`, that way you don't have to look for the IsisCube object in the cube label.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+ * Finds the index of the tracking band in the BandBin group and stores the index
+ * in trackBand. The other bands' indices are stored in copyBands.
+ * We store the indices in QStrings because we use them as cube attributes when
+ * we process the mosaic and tracking cubes.
+ *
+ * @param ui        The user interface
+ * @param copyBands Indices of the bands to copy over to the mosaic cube
+ * @param trackBand Index of the tracking band
+ */
+void findTrackBand(UserInterface ui, QString &copyBands, QString &trackBand) {
```
jessemapel: You should directly pass the cube filename instead of the user interface. You could also just pass a pointer to the cube. Passing the user interface is opaque, it's not clear what it's going to be used for. This way, if we ever move this out of this application code we can just copy and past it over.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Creates the mosaic cube by copying the input cube without the tracking band.
+ * Then, removes the tracking table from the mosaic label and adds a group pointing
+ * to the tracking cube.
+ *
+ * @param ui    The user interface
+ * @param bands The indices of the bands that are not the tracking band
+ */
+void createMosaicCube(UserInterface ui, QString bands) {
```
jessemapel: See previous comment about passing the user interface. Just pass the filenames.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+  if (mosaicLabel->hasObject("Table")) {
+    PvlObject &tableObject = mosaicLabel->findObject("Table");
+    if (tableObject.hasKeyword("Name")) {
+      PvlKeyword &tableName = tableObject.findKeyword("Name", Pvl::Traverse);
+      if (tableName[0] == "InputImages") {
+        mosaicLabel->deleteObject("Table");
+      }
+    }
+  }
```
jessemapel: If the cube has two tables and the second table is the tracking table, this won't find it!.  I'd suggest using `Cube::deleteBlob` to remove the old tracking table.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+  // Add Tracking Group to the mosaic cube
+  PvlGroup trackingGroup = PvlGroup("Tracking");
+  PvlKeyword trackingName = PvlKeyword("Filename");
+  FileName cubeName = FileName(ui.GetFileName("TO"));
+  trackingName.setValue(cubeName.baseName() + "_tracking.cub"); //Strip off path and add _tracking
```
jessemapel: I don't remember if this removes any path to the file. The `Filename` keyword should be just the tracking cube file name and not any path.  For example, the TO keyword could be `some/path/to/some/dir/output.cub`, you want the `Filename` keyword to be `output_tracking.cub`, not `some/path/to/some/dir/output_tracking.cub`

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+  trackingGroup.addKeyword(trackingName);
+  mosaicLabel->addGroup(trackingGroup);
```
jessemapel: This will add the tracking group outside of the IsisCube object. Use `cube::putGroup` to add the tracking group.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Creates the tracking cube by copying the input cube with only the tracking band.
+ * Then, goes through each pixel and subtracts the input cube's pixel type's min value
+ * because ProcesMosaic used to add the min value to each pixel.
+ *
+ * @param ui           The user interface
+ * @param trackingBand The index of the tracking band
+ */
+void createTrackCube(UserInterface ui, QString trackBand) {
```
jessemapel: Again, pass the filenames, not the user interface

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+  FileName cubeName = FileName(ui.GetFileName("TO"));
+  // Strip off any extensions and add _tracking
+  QString trackingName = cubeName.path() + "/" + cubeName.baseName() + "_tracking.cub";
+  Cube inputCube = Cube(ui.GetFileName("FROM"));
+  int numSample = inputCube.sampleCount();
+  int numLine = inputCube.lineCount();
+
+  QString outAttString = "+UNSIGNEDINTEGER+" + QString::number(VALID_MINUI4) + ":";
+  outAttString += QString::number(VALID_MAXUI4);
+  CubeAttributeOutput outAtt = CubeAttributeOutput(outAttString);
```
jessemapel: You can create a CubeAttributeOutput object and then set the pixel type, min, and max through its own methods instead of by using the UI string.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    case 4:
+      offset = FLOAT_MIN;
+      defaultVal = INULL4;
+      break;
+
+    default:
+      QString msg = "Invalid Pixel Type [" + QString(in.PixelType()) + "]";
+      throw IException(IException::Programmer, msg, _FILEINFO_);
+  }
```
jessemapel: This offset is going to be the same within a single execution. I'd like to see the offset computed outside of this method in createTrackCube and then passed in through a global variable or a functor.

jessemapel: Same with the default value.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+  for (int i = 0; i < in.size(); i++) {
+    if (in[i] == (float) defaultVal) {
+      out[i] = NULLUI4;  // Set to the unsigned 4 byte Null value
+    }
+    else {
+      out[i] = ((int) in[i]) - offset + VALID_MINUI4;
```
jessemapel: Should this be a qround call or an int cast? I'm not sure.

jessemapel (changes requested):

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: I store them as a QString because CubeAttributeOutput takes in a string of bands to copy over from the input cube. If I stored them in a list, I would have to create another for loop to store them in a QString.

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: I do this because I would have to create another for loop to grab the index of the tracking band because I need that number stored in a QString to pass it into CubeAttributeOutput.

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: This does strip the path from the Filename, so no worries.

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: We looked at qview, and they cast the pixel value to an int instead of using qround.

isis/src/base/apps/trackextract/trackextract.cpp
jessemapel: My main concern is that it's completely opaque that this may be a list of bands. Using a QStringList would fix that without much difficulty. You can just add the band indices as you get them, you wouldn't need line 60 anymore. Then, when you add them to the filename you can use QStringList::join(",") to put them together.

isis/src/base/apps/trackextract/trackextract.cpp
jessemapel: :+1:

isis/src/base/apps/trackextract/trackextract.cpp
jessemapel: That's probably the correct choice then.

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: If we are not using these values in other places, I thought it would be simpler to calculate these values in the same method they are being used in. Do you foresee us having to use these values in multiple places in the future?

isis/src/base/apps/trackextract/trackextract.cpp
kaitlyndlee: I did not realize, but CubeAttributeInput has a setBands method that takes in a vector of QStrings. I'll go ahead and change it to that. Thanks!

isis/src/base/apps/trackextract/trackextract.cpp
jessemapel: The main issue is wasting time in your process function. This method is going to be called for every line in every band in your cube. We're not aiming for huge efficiencies in this app and it's really just a couple comparisons and assignments, so it can probably be ignored.

kaitlyndlee: I'm sorry, I completely forgot that it would call this method every line.  I agree with you that it would be better to grab the values in createTrackCube instead.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  CubeAttributeOutput outAtt;
+  outAtt.setPixelType(UnsignedInteger);
+  outAtt.setMinimum(VALID_MINUI4);
+  outAtt.setMaximum(VALID_MAXUI4);
+
+  p.SetOutputCube(trackingName, outAtt, numSample, numLine);
+
+  p.StartProcess(copyTrackPixels);
+  p.EndProcess();
```
jessemapel: Is the new tracking table ever written to the tracking cube? This app needs to write an updated tracking table to the tracking cube.

jessemapel (changes requested):

isis/src/base/apps/trackextract/trackextract.cpp
jessemapel: I'm assuming that the existing table from the input cube is copied to the new tracking cube by ProcessByLine, but it'd be nice if it was in the new format with the indices as a 3rd field. You could initialize a TrackingTable object from the input cube's table and then immediately call TrackingTable::toTable and write that to the tracking cube.

isis/src/base/apps/trackextract/trackextract.cpp
```diff
@@ -0,0 +1,234 @@
...
+    int m_offset;
+    int m_defaultValue;
+
+  public:
+    CopyPixelsFunctor(int offset, int defaultValue) {
+      m_offset = offset;
+      m_defaultValue = defaultValue;
+    }
+
+    void operator()(Buffer &in, Buffer &out) const {
```
jessemapel: The operator and constructor will need method documentation too.

jessemapel (approved):

kaitlyndlee:
[Write an application to take a tracking band in a mosaic and convert it into a tracking cube.](https://trello.com/c/eH0dz8FY/43-write-an-application-to-take-a-tracking-band-in-a-mosaic-and-convert-it-into-a-tracking-cube)

PR #378, opened 2018-07-26
Changes radius used for scaling lat/lon sigmas from equatorial to local radius -- fixes #5457

dcookastro:


isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -279,6 +256,8 @@ namespace Isis {
     else {
       p_z = new Displacement(z);
     }
+    
+    p_localRadius = NULL;
```
jessemapel: This should check if there is an existing local radius and delete it before assigning to NULL. Current implementation is a potential memory leak

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -697,48 +693,8 @@ namespace Isis {
...
-    }
-
-    if(p_polarAxis) {
-      *p_polarAxis = polarRadius;
-    }
-    else {
-      p_polarAxis = new Distance(polarRadius);
-    }
+    
+    p_localRadius = NULL;
```
jessemapel: This also needs to delete any existing Distance object

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -905,11 +871,16 @@ namespace Isis {
...
+        return *p_localRadius;
+      }
+      else {
+        double x = p_x->meters();
+        double y = p_y->meters();
+        double z = p_z->meters();
 
-      return Distance(sqrt(x*x + y*y + z*z), Distance::Meters);
+        return Distance(sqrt(x*x + y*y + z*z), Distance::Meters);
+      }
```
jessemapel: Why does this not save the local radius object?

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -924,14 +895,10 @@ namespace Isis {
         Angle latSigma = GetLatSigma();
 
         if (latSigma.isValid()) {
-          if (!p_majorAxis || !p_majorAxis->isValid()) {
-            IString msg = "In order to calculate sigmas in meter units, the "
-              "equitorial radius must be set with a call to SetRadii.";
-            throw IException(IException::Programmer, msg, _FILEINFO_);
-          }
+          Distance scalingRadius = GetLocalRadius();
```
jessemapel: This local variable can be removed and line 901 can just be `latSigmaDistance = latSigma.radians() * GetLocalRadius();`

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -1038,18 +999,18 @@ namespace Isis {
...
+   * the local radii will be used.
    */
   Distance SurfacePoint::GetDistanceToPoint(const SurfacePoint &other) const {
-    if(p_majorAxis || p_minorAxis || p_polarAxis ||
-       other.p_majorAxis || other.p_minorAxis || other.p_polarAxis) {
-      IString msg = "SurfacePoint::GetDistanceToPoint not yet implemented for "
-          "ellipsoids";
-      throw IException(IException::Programmer, msg, _FILEINFO_);
-    }
+    // TODO What do we do for the check? We no longer have the ability to check.
```
jessemapel: I think we just have to accept that this is the best we can do without the actual surface shape. Getting the true distance between points on a complex shape is probably too complex for what this is used for.

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
```diff
@@ -1156,28 +1106,16 @@ namespace Isis {
...
-       !p_majorAxis && !other.p_majorAxis &&
-       !p_minorAxis && !other.p_minorAxis &&
-       !p_polarAxis && !other.p_polarAxis &&
        !p_rectCovar && !other.p_rectCovar &&
        !p_sphereCovar && !other.p_sphereCovar) {
       *p_x = *other.p_x;
       *p_y = *other.p_y;
       *p_z = *other.p_z;
+        // Finally initialize local radius to avoid r
+       p_localRadius = NULL;
```
jessemapel: Delete existing here too

celbal:
Can you please update the title of this pull request to reflect the changes you've made and end the title with Fixes #5457?

jessemapel (changes requested):
A couple minor fixes need to happen with SurfacePoint

dcookastro:
Done

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
dcookastro: Fixed

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
dcookastro: I can't change the object here, because of the const.  Every method that calls GetLocalRadius is also followed by const as well so I can't get around it taking the const off this method.  The best solution I can think of is to calculate the local radius and set it in the methods that set the coordinates.  It means we are setting this value whether we need it or not, but most of the time it will just be an assignment from the radius of the latitudinal coordinates.  I will only need to calculate it for the rectangular coordinates.  Do you have any suggestions?

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
dcookastro: Fixed.

isis/src/base/objs/SurfacePoint/SurfacePoint.cpp
dcookastro: Thanks.  I removed the TODO comment.

jessemapel (dismissed):

jessemapel:
Still needs to be re-reviewed, I'm just approving the previous changes

jessemapel:
Looks good this is ready to merge

jessemapel (approved):

jessemapel:
@dcookastro This is ready to go, once you're ready to merge test data go ahead and squash and merge this, then check in your test data

dcookastro:
Updated truth data are being checked in right now.

jessemapel:
:+1:

PR #379, opened 2018-07-27
Added an offset to TrackingTable to account for 0-2 being special pixels. References #971

jessemapel:
I also changed the following method names:

indexToFileName -> pixelToFileName
fileNameToIndex -> fileNameToPixel

because the values are not actually indices now, they are pixel values that have to get converted to indices. Same with the field in the tracking table.

I also tweaked some stuff in the unitTest and some of the error messages.


SgStapleton:
This all looks good to go.

PR #381, opened 2018-07-27
Modified Histogram Cube constructor to work with signed and unsigned word cubes

jessemapel:
I also added tests to cover the different bit types for the constructor.


krlberry (approved):
:+1: This all looks good to me.

PR #382, opened 2018-07-27
Fix to "points not showing up in qnet" problem

krlberry:
After merging the control net changes from ipce into dev, control measures no longer appeared in cubes opened in qnet. This PR fixes this issue. 

This was tested by Ethan.

### DIFF
--- a/isis/src/control/objs/ControlNet/ControlNet.cpp
+++ b/isis/src/control/objs/ControlNet/ControlNet.cpp
@@ -764,7 +764,6 @@ namespace Isis {
     // Conceptually, I think this belongs in measureIgnored, but it isn't done
     // for the old graph.
     m_controlGraph[m_vertexMap[serial]].measures.remove(measure->Parent());
-
   }
 
 
@@ -1819,7 +1818,6 @@ namespace Isis {
     std::swap(points, other.points);
     std::swap(pointIds, other.pointIds);
     m_controlGraph.swap(other.m_controlGraph);
-    std::swap(m_vertexMap, other.m_vertexMap);
     std::swap(m_mutex, other.m_mutex);
     std::swap(p_targetName, other.p_targetName);
     std::swap(p_networkId, other.p_networkId);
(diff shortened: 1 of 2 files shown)


isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1864,6 +1860,14 @@ namespace Isis {
...
     }
 
+    m_vertexMap.clear();
+    VertexIterator v, vend;
+    for (boost::tie(v, vend) = vertices(m_controlGraph); v != vend; ++v) {
+      ImageVertex imVertex = *v;
+      QString serialNum = m_controlGraph[*v].serial;
+      m_vertexMap[serialNum] = imVertex;
+    }
+
```
jessemapel: I'm worried that maybe this needs to be in swap(). If you just call swap on a ControlNet right now, the vertexmap will be completely broken. In fact I think the vertex maps in both ControlNets will be broken.

krlberry: :+1: I'll go ahead and move this in there and make sure everything still works. I don't remember why I moved this out of ControlNet::swap() in the first place.

jessemapel: I think you'll need to go through this process for both cubes that are being swapped, self and other.

krlberry: Good point, thank you.

isis/src/control/objs/ControlNet/ControlNet.cpp
```diff
@@ -1844,8 +1842,23 @@ namespace Isis {
...
+    other.m_vertexMap.clear();
+    VertexIterator v2, vend2;
+    for (boost::tie(v2, vend2) = vertices(other.m_controlGraph); v2 != vend2; ++v2) {
+      ImageVertex imVertex = *v2;
+      QString serialNum = other.m_controlGraph[*v2].serial;
+      other.m_vertexMap[serialNum] = imVertex;
+    }
 
+    emit networkModified(ControlNet::Swapped);
```
jessemapel, line 1861: We also need the other network to emit this signal, I'm not sure what the best way to do that is though? For right now, I don't think this is being used anywhere that it matters, but it may come up in the future.

jessemapel (approved):

PR #383, opened 2018-07-30
Fixed c++ compiler build warnings on prog28 and prog29

AgoinsUSGS:
Changes include:  Catching IException as a reference rather than value, 
                              Not using Qt::ControlModifier as a boolean reference, 
                              Removed deprecated dynamic throw() call.

There are special notes for the fixes for the Qt::ControlModifier and throw() warnings in the PR.

### DIFF
--- a/isis/src/mro/apps/hijitreg/Instrument.cpp
+++ b/isis/src/mro/apps/hijitreg/Instrument.cpp
@@ -390,7 +390,8 @@ namespace UA {
       unsigned int  CPMM,
       unsigned int  binning
     )
-    throw(Out_of_Range, Invalid_Argument) {
+    {
+
       if (CPMM >= CCDS) {
         ostringstream
         message;
@@ -414,7 +415,7 @@ namespace UA {
       return
         (int)rint(CCD_FOCAL_PLANE_X_OFFSETS_MM[CPMM]
                   / (CCD_PIXEL_SIZE_MM * binning));
-    }
 
+  }
   }   //  namespace HiRISE
 }   //  namespace UA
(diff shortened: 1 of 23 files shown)


isis/src/qisis/objs/CubeViewport/CubeViewport.cpp
```diff
@@ -1776,7 +1776,7 @@ namespace Isis {
       e->accept();
     }
     else if ((e->key() == Qt::Key_C) &&
-             QApplication::keyboardModifiers() &&
+             QApplication::keyboardModifiers() &
              Qt::ControlModifier) {
```
AgoinsUSGS, line 1780: warning: enum constant in boolean context [-Wint-in-bool-context] was being thrown by using the logical &&, I replaced it with the bitwise & (only difference as I understand it is that && can short-circuit and & will not, which I don't see is a huge difference in this context). If someone see's an issue with this fix let me know

isis/src/mro/apps/hijitreg/Instrument.cpp
```diff
@@ -414,7 +415,7 @@ namespace UA {
       return
         (int)rint(CCD_FOCAL_PLANE_X_OFFSETS_MM[CPMM]
                   / (CCD_PIXEL_SIZE_MM * binning));
-    }
```
AgoinsUSGS, line 417: warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]

For this, we were using a dynamic throw block that would throw either of the 2 exceptions in the block itself. 
These dynamic throws are deprecated in c++11 and removed from c++17 onward, so I moved the individual throws to be outside the throw() block and got rid of it. If anyone see's an issue with this please let me know.

SgStapleton: Confirmed that this should not be an issue.

isis/src/qisis/objs/CubeViewport/CubeViewport.cpp
SgStapleton: Confirmed that functionality has not changed. Holding cntrl-c copies the filename to the clipboard in both the nightly build as well as this version.

SgStapleton:
Everything looks good to go!

PR #388, opened 2018-07-31
Updated qview to use TrackingTable 

kaitlyndlee:


kdl222:
[Integrate tracking table class in qview](https://trello.com/c/NZwiAGNI/42-integrate-tracking-table-class-in-qview)

isis/src/qisis/objs/FileTool/FileTool.cpp
```diff
@@ -580,9 +580,9 @@ namespace Isis {
         else if ((ocube->pixelType() != Real) &&
                 (ocube->pixelType() != UnsignedByte) &&
                 (ocube->pixelType() != SignedWord) &&
-                (ocube->pixelType() != UnsignedWord)) &&
-                (cube->pixelType() != Isis::UnsignedInteger) &&
-                (cube->pixelType() != Isis::SignedInteger)) {
```
SgStapleton, line 585: How was this even compiling before!?

isis/src/base/objs/TrackingTable/TrackingTable.cpp
```diff
@@ -176,4 +176,53 @@ namespace Isis {
...
+    for (int i = 0; i < m_fileList.size(); i++) {
+      if (m_fileList[i].first == file) {
+        return i;
+      }
+    }
+
+    // At this point, the file is not in the internal file list so append it
+    // and return its new index.
+    m_fileList.append(QPair<FileName, QString>(file, serialNumber));
+    return m_fileList.size() - 1 + VALID_MINUI4;
```
SgStapleton: Do we need to add the VALID_MINUI4 offset here? Seems like we would not want that as the earlier logic is looking for the index in the m_fileList - which would not contain that offset...

isis/src/qisis/objs/FileTool/FileTool.cpp
kaitlyndlee: I added this in with my last PR, but it got fixed immediately after. These changes are already in the branch. Not sure why it thinks it's not.

isis/src/base/objs/TrackingTable/TrackingTable.cpp
kaitlyndlee: Oops, forgot to take that out.

SgStapleton:
The changes look good to go!

PR #398, opened 2018-08-14
Updated MosaicSceneWidgetTester and ProgramLauncher truth data

AgoinsUSGS:
Commented out the MosaicSceneWidgetTester because at this time we don't use QtTest for testing our GUI stuff. The code is commented out rather than removed incase we want to revisit it at some point.


kaitlyndlee:
Instead of commenting out the code, should we delete all the testing files in the directory, since we are not really testing anything anymore?

AgoinsUSGS:
We should double check with Stuart on that. My understanding was that we wanted the testing files there as a placeholder if we want to revisit this but I could be mistaken

AgoinsUSGS:
MosaicSceneWidgetTester and unitTest.cpp have been removed from MosaicSceneWidget

PR #400, opened 2018-08-14
Fixed three more files with messed up doxygen tags

twilson271828:
These have been throwing warnings in the nightly builds.


SgStapleton:
Simple change - looks good to go!

SgStapleton (approved):
Ran-over this with Tyler. Looks good to go!

PR #404, opened 2018-08-21
Updated cube tolerances in the photomet test useDem

kaitlyndlee:
Stuart and I had changed the dem the test was using because it was 32 gb and was not located in our test data area. Truth data for Mac was checked in and the output for prog25 is not very different. Just needed to update the tolerances.


SgStapleton (approved):
Confirmed truthdata checked in. Looks good!

PR #410, opened 2018-08-24
(DO NOT MERGE) Turn off useQuantizedAabbCompression in BulletDskShape

krlberry:
This fixes the segfaults that occur in the parts of the raytracing cat test and spiceinit app test that use the Bullet library. This may be a temporary fix as it would be much better to be able to continue using the compression, but I was unable to figure out how to do this for now.


krlberry:
Need to get the okay to actually do this.

krlberry:
Received external feedback that this should not be turned off.

PR #412, opened 2018-08-27
Removing SquishTests category tests

SgStapleton:
Removing unnecessary GUI category tests


celbal (approved):

celbal:
These tests were breaking the cmake tests and they are no longer used therefore they are being removed.

PR #413, opened 2018-08-27
Fixed cam2cam looking for the wrong ground point on irregular bodies. Fixes #5425

### LINKED ISSUE #5425: Supporting ShadowCam Image Pre-processing
Where can I see the process of pre-processing ShadowCam using ISIS software? Thanks!

jessemapel:
Currently, cam2cam tries to find the ground point in the output image that is closest to the output image observer. With irregular bodies this causes weird mappings when there are multiple ground points with the same lat/lon (ie comet 67P).

New test data for one noproj test and test logs for fedora and mac 10.11 can be found at `/work/projects/isis/latest/m05425/ISIS3`


krlberry (approved):
:+1: Looks good.

PR #415, opened 2018-08-28
re-add gold linker for linux only

ihumphrey:


SgStapleton:
Tested this. Looks good to go.

PR #418, opened 2018-08-31
Cmake tests

SgStapleton:
These are a few smaller changes for app and unit test to pass.


jlaura:
:+1: Getting CMake rolling along. The work you all are doing on this is super appreciated for improving the ability for internal and external developers to build the code and contribute to the project. Thank you!

A question with a lot of these tests: NaN is an IEEE defined value. Negative NaN is not a thing. Sure the value has a sign bit, but the IEEE spec says that it should be ignored. [Section 6.3](https://books.google.com/books?id=9NqiCgAAQBAJ&pg=PA49&lpg=PA49&dq=When+either+an+input+or+result+is+NaN,+this+standard+does+not+interpret+the+sign+of+a+NaN.&source=bl&ots=8N7MR0vbR7&sig=s3XSJl5QXOk7oEm6GF7O_ttDds4&hl=en&sa=X&ved=2ahUKEwjQrPOV2ZjdAhUNOa0KHYZsCdQQ6AEwCXoECAgQAQ#v=onepage&q=When%20either%20an%20input%20or%20result%20is%20NaN%2C%20this%20standard%20does%20not%20interpret%20the%20sign%20of%20a%20NaN.&f=false) As you read on in Section 6.3, the standard does offer the instances where a sign check (and therefore test) might be appropriate, but this should be quite infrequent.

Differentiating between NaN and -NaN is a great way to throw false positive failures with an OS or tooling change (as you seem to be finding).

PR #421, opened 2018-09-05
Update catoriglab XmlLabel test 

krlberry:
Update catoriglab XmlLabel test to strip out attributes from test data as they can occurr in any order. Fixes the failing catoriglab test in nightly


ihumphrey (approved):
looks like you are addressing all the labels tags that have multiple attributes, looks good :+1:

krlberry:
@SgStapleton Pointed out that the order of attributes is always the same for each OS. We can make OS-Specific truthdata instead and have a way to test the attributes, rather than removing them. @scsides Likes this approach better than I agree, so reverting this PR.

PR #422, opened 2018-09-05
For VIRTIS-M calibrated images, fixes the reading of SCETs into an ISIS3 table 

krlberry:
Fixes the reading of SCETs into the ISIS3 table in the ingested image for level-3, calibrated VIRTIS-M images.


jcwbacker (approved):
Looks good. I'll add more stuff after I merge.

PR #424, opened 2018-09-06
Revert "Update catoriglab XmlLabel test "

krlberry:
Reverts USGS-Astrogeology/ISIS3#421

@SgStapleton Pointed out that the order of attributes is always the same for each OS. We can make OS-Specific truthdata instead and have a way to test the attributes, rather than removing them. @scsides Likes this approach better than I agree, so reverting this PR.


SgStapleton (approved):
Revert looks good!

PR #426, opened 2018-09-07
Fixed environment

jessemapel:
Updated qt to 5.9.6 to avoid dependency error with qwt, also fixed armadillo version


SgStapleton:
Looks good!

PR #427, opened 2018-09-07
(DO NOT MERGE) Initial updates to RosettaVirtisCamera model to support VIRTIS-M calibrated images

krlberry:
These are initial updates to the RosettaVirtisCamera model to support VIRTIS-M calibrated (level 3) images. With this code, spiceinit runs without error, we have intersection with test images, but the pointing is not great. At least part of this is that the timing format mismatch between Rosetta and ISIS3 still needs to be fixed.


krlberry:
Do not merge for now, accidentally updated with incomplete code.

PR #431, opened 2018-09-11
Added initial updates to the RosettaVirtisCamera model to support VIRTIS-M calibrated Images

krlberry:
These are initial updates to the RosettaVirtisCamera model to support VIRTIS-M calibrated (level 3) images. With this code, spiceinit runs without error, we have intersection with test images, but the pointing is not great. At least part of this is that the timing format mismatch between Rosetta and ISIS3 still needs to be fixed.


scsides (approved):
These changes look good to go

PR #432, opened 2018-09-13
Fixed qview to accept cubes and cube lists under any extension. Fixes m05439 and m05476.

AgoinsUSGS:
qview will now attempt to open the file it receives as a cube first (including a cube labels) and if that fails, attempt to open it as a cubelist and if that fails, then notify the user of an error.

This allows all cubes, detached cubes, and cube lists to be accepted under any extension.

Changes can be found in /work/projects/isis/latest/m05439_m05476.


isis/src/qisis/objs/Workspace/Workspace.h
```diff
@@ -157,41 +162,45 @@ namespace Isis {
...
       /**
        * Method adds the name of a cube into Workspace as a CubeViewport
-       * 
+       *
        * @param cubename The cube to be added to the Workspace
        */
       void addCubeViewport(QString cubename);
-      
+
+
```
SgStapleton: Want to add a little method header here (like the other methods have)?

isis/src/qisis/objs/Workspace/Workspace.cpp
```diff
@@ -330,44 +330,89 @@ namespace Isis {
...
-      // If the file received isn't a cube or label, it has to be a cubelist. We read every cube in
-      // the cubelist and append it to the cubesToOpen QList so that we can open them.
-      QFile file(cubename);
-      file.open(QIODevice::ReadOnly);
 
-      QTextStream in(&file);
+    catch (IException &e) {
+
+      QString message("Error opening cube [" + cubename + "]...\n");
+      message += "Attempting to open [" + cubename + "] as a cube list...\n";
```
SgStapleton, line 373: Are we sure that we want an "error" message to display if the user enters a cube list? Since a cube list is a valid option, it does not seem appropriate to ever display any kind of "error" message to the user...

SgStapleton: ...Plus I guess this is never being displayed anyhow...

AgoinsUSGS: This error message is displayed only if attempting to open it as a cube list fails and serves the purpose of alerting the user that first it was attempted to open as a cube and when that failed it attempted to open it as a cube list. If the cube list succeeds they never see it failed when it tried to open it as a cube first

ihumphrey (approved):
Looks good, looks like Lynn tested this as well (see 5439 for comments)

PR #433, opened 2018-09-20
Fix StopTime and SpacecraftClockStopCount in VIRTIS-M Label

krlberry:
Updates the SpacecraftStartClockCount, SpaceCraftStopClockCount, StartTime, and StopTime to be accurate in the labels of ingested Rosetta VIRTIS-M calibrated images. 

Also includes some small updates to the camera model. 

UPDATE: 
In addition to fixing the incorrect scet calculation, now the table attached to the ingested cube contains the (hopefully!) correctly-formatted sclk-strings rather than the scet values right out of idl, as these are more useful to the camera model.

### DIFF
--- a/isis/src/rosetta/apps/rosvirtis2isis/rosvirtis2isis.cpp
+++ b/isis/src/rosetta/apps/rosvirtis2isis/rosvirtis2isis.cpp
@@ -30,6 +30,7 @@ int word(const char byte1, const char byte2);
 int swapb(const unsigned short int word);
 double translateScet(int word1, int word2, int word3);
 bool isValid(int word);
+QString convertSCET(QString scetOriginal); 
 
 void IsisMain ()
 {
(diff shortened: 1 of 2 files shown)


isis/src/rosetta/apps/rosvirtis2isis/rosvirtis2isis.cpp
```diff
@@ -421,6 +477,23 @@ void IsisMain ()
...
+QString convertSCET(QString scetOriginal) {
+  // whole seconds stay the same
+  QStringList scetStringList = scetOriginal.split('.');
+  
+  //  SSSSSSSSSS:
+  QString scetFinal = scetStringList[0];
+  scetFinal.append(":");
+
+  // FFFFF
+  double fractValue = toDouble(scetStringList[1])/65536.0;
```
jessemapel: This should be a multiplication.

jessemapel (changes requested):
Converting the scet to a double doesn't make sense, we should just keep it as a string in the table, then convert to a double as part of the scet2et process.

PR #434, opened 2018-09-20
Removed absolute path in csv2table errors test

jessemapel:
the errors test for csv2table is failing because the error messages have absolute file paths in them. Added a regex to strip out everything except the final file.


krlberry (approved):
:+1: Looks good!

PR #438, opened 2018-09-25
isis3.6.0 version update

ihumphrey:


isis/version
```diff
@@ -1,5 +1,2 @@
-3.5.3.0      # Public version number
-2018-08-01   # Version date
```
celbal, line 2: Do we want the version date any more? If we don't know what the date is going to be we can always put the month and year rather than the actual day.

ihumphrey: good idea, then we can change the date when the release candidate and stable release go out

celbal (approved):

Issue #446, opened 2018-09-27
ISIS3/cmake curl==7.55.1=0 broken


ihumphrey:
When trying to set up the isis3 dependencies environment from our ```environment.yml``` file, the following error is issued:

```bash
ResolvePackageNotFound: 
  - curl==7.55.1=0
```

Looking at [conda-forge's curl packages](https://anaconda.org/conda-forge/curl/files?page=2), the ```linux-64/curl-7.55.1-0.tar.bz2``` file has the ```broken``` label. All of the linux 7.55.1-0 versions have the ```broken``` label, so it will not be found by conda using our environment specification (conda looks for ```main``` label packages).

_This can be fixed quickly by changing the version to a higher package version in the main label hosted by conda-forge._

However, this potentially brings up some questions. I'll post these in a separate issue.

ihumphrey:
This was fixed in #456

PR #449, opened 2018-09-28
Fix for OpenGL error on Conda Build

Kelvinrr:
It seems the issue was in the CMake scripts that ship with Qt. They find the incorrect version of OpenGL, most likely a bug on Qt's end as the version found isn't the library but rather the directory of the framework, hence the error. The solution is to add the PATH option with the default system location for OpenGL on MacOS machines. 

The package is uploaded to Anaconda under the USGS-Astro channel, probably worth having someone other than myself test it. 

~Also this should be squashed and merged for obvious reasons.~ Squashed locally and pushed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,7 +13,6 @@ dependencies:
   - bzip2==1.0.6=1
   - cmake==3.9.1=0
   - cspice==66-0
-  - curl==7.55.1=0
   - doxygen==1.8.14=0
   - eigen==3.3.3=0
   - embree==2.14.0=0
--- a/isis/CMakeLists.txt
+++ b/isis/CMakeLists.txt
@@ -209,7 +209,9 @@ find_package(Qt5 COMPONENTS
                 WebChannel
                 Widgets
                 Xml
-                XmlPatterns REQUIRED)
+                XmlPatterns
+                # Search this path explicitly for MacOS OpenGL Framework
+                PATHS /System/Library/Frameworks/ REQUIRED)
 
 # Some of these will have non-traditional installs with version numbers in the paths in v007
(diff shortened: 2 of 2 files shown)


jlaura:
:+1:

jlaura:
:+1:

jessemapel:
I'll test this out today

environment.yml
```diff
@@ -13,7 +13,6 @@ dependencies:
   - bzip2==1.0.6=1
   - cmake==3.9.1=0
   - cspice==66-0
-  - curl==7.55.1=0
```
jessemapel, line 16: Is this not breaking? see #446

Kelvinrr: I pretty sure this is a dependency of a dependency so we shouldn’t have to manage it separately. So if the version removed where is broken, removing the line has anaconda implicitly search for a version that isn’t broken but still satisfies the requirements. From using it personally (albeit briefly) didn’t seem to have any problem.

chrisryancombs: This specific build of curl is broken. I think it would be better to remove the version number.

Kelvinrr: I think it comes down to whether or not we use it directly, and if we care about the version number. If the answer to both of these questions are no, remove it altogether and allow another package to handle it itself. If we do use it directly and don’t care too much about version (which sounds like the case) then remove it and have no version number. I think we should stay away from build numbers from anything not built by us and very exceptional cases in general. The more flexible the environment file the better.

chrisryancombs (approved):
Looks good. Builds and tests as expected on linux and mac.

environment.yml
jessemapel: @Kelvinrr that fix sounds good to me

PR #468, opened 2018-10-03
Fix problems exporting tgo cassis mosaics

krlberry:
This fixes a problem that comes up when trying to export a mosaic generated using `tgocassismos` using `tgocassisrdrgen`. (`tgocassisrdrgen` was not looking for the `ObservationId` in the `Mosaic` group in the label of the ISIS3 cube. It was looking for it in the `Archive` group, which does not exist for mosaics.)


SgStapleton:
Simple change. Looks good!

SgStapleton (approved):
Everything looks good!

PR #469, opened 2018-10-04
Update tgo's singleFrameletProjection test to fix error. 

krlberry:
Update tgo's `singleFrameletProjection` category test to not include the modification date or ISIS3 version in the test, as these can change depending on when the test is run. This fixes the failure in nightly.


ihumphrey (approved):
Seems like a reasonable change to me, thanks Kristin :+1:

krlberry:
@ihumphrey-usgs Thanks for approving this! :+1:

PR #476, opened 2018-10-11
OREX socetframsettings changes Fixes #5235

### LINKED ISSUE #5235: campt reports PixelValue from the first band always
**ISIS version(s) affected**: 7.0.0

**Description**  

campt reports PixelValue from the first band even when asking for band != 1

**How to reproduce**  

download 10 by 10 pixel Chandrayaan-1 M3 sub-cube [M3_10x10.cub.zip](https://github.com/DOI-USGS/ISIS3/files/11900429/M3_10x10.cub.zip)

```
unzip M3_10x10.cub.zip
campt from=M3_10x10.cub+1 sample=1 line=1  | grep PixelValue
PixelValue                 = -18.618536
```

```
campt from=M3_10x10.cub+25 sample=1 line=1  | grep PixelValue
PixelValue                 = -18.618536
```
should be: 16.72398

**Possible Solution**  
**explode**-ing the cube then reassemble it, as a workaround.

**Additional context**

jessemapel:
This PR contains changes made by @drbitboy under #94. It also contains my own changes from the review process. An internal build can be found at /work/users/jmapel/ISIS3/isis with the new test data for the two additional tests.

Until the Clementine camera model errors are fixed, this cannot be merged because one of the socetframesettings tests depends on it.

### DIFF
--- a/isis/src/socet/apps/socetframesettings/socetframesettings.cpp
+++ b/isis/src/socet/apps/socetframesettings/socetframesettings.cpp
@@ -30,10 +30,10 @@
 using namespace std;
 using namespace Isis;
 
-void getCamPosOPK(Spice &spice, QString spacecraftName, double et, Camera *cam,
-                  double ographicCamPos[3], double omegaPhiKappa[3],
-                  double isisFocalPlane2SocetPlateTranspose[3][3]);
-      
+void getCamPosOPK(Spice &spice, QString spacecraftName, SpiceDouble et, Camera *cam,
+                  SpiceDouble ographicCamPos[3], SpiceDouble omegaPhiKappa[3],
+                  SpiceDouble isisFocalPlane2SocetPlateTranspose[3][3]);
+
 void IsisMain() {
 
   // Use a regular Process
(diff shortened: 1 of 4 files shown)


jessemapel:
@thareUSGS I'm building for CentOS7 is at `/work/projects/isis/latest/m05235` I also have the test data from Sarah Sutton pulled down there. It should be read within in an hour.

thareUSGS:
I tested three instruments MDIS NAC, Viking, and Galileo ISS between isis3.5.2 and isis3.6. Galileo ISS was identical. And for MDIS and Viking they were essentially identical.  Only a couple parameters were slightly different and they only differed past 11 decimal places. I am fine with the update. I also tested back to isis3.5.0 (same results).

jessemapel:
@thareUSGS Can you submit an approving review for this PR based on your testing?

thareUSGS (approved):
New code appears not to impact existing instruments. I have not reviewed the new code in terms of correctly representing the O-Rex instrument. Tests are being added to help with this. As a note, there are very verbose comments. This perhaps could be pulled out into a different document (for example describing and graphing the difference between FITS and ISIS3).

drbitboy:
Those contents do not apply only to O-REx; they explain the relationship between ISIS and SOCET, so future instruments can be added.  Stepping back a bit, there are only eight cases, equivalent to the 8 rotations in the IDL ROTATE routine.  Very few missions will not fit into one of those cases, so setting up code to run each possible case, and then assigning the chosen ISIS convention for each instrument to one of those cases, will make adding future instrument a one-line modification.

Part of the problem going in before O-REx was that there was neither documentation of how SOCETSET defines its frames nor an explanation of its requirements.  Now we have that.

PR #496, opened 2018-10-23
Update for new Clementine camera model

scsides:


scsides:
@SgStapleton approved this merge

PR #503, opened 2018-10-25
Fixing Pipeline failing test

SgStapleton:
This was due to profile IsisPreferences file being executed for nested app calls


SgStapleton:
OK per @ihumphrey-usgs

PR #506, opened 2018-10-26
Modified qnet to not close if the 'cancel' option was selected after closing a modified cnet

AgoinsUSGS:
After modifying a control network and clicking the close button, the user would receive a close confirmation popup. If the user selected 'cancel' qnet would close anyway.

Qnet was modified to intercept the QCloseEvent that was triggered when this would happen to ensure that the window was only closed if the cancel button wasn't selected.


kaitlyndlee (approved):
Everything looks good!

PR #529, opened 2018-11-07
Angle test refactor to use gtest

### LINKED ISSUE #555: Rewrite Angle test in gtest

krlberry:
Adds a gtest unit test for Angle. Fixes #555

### DIFF
--- a/isis/tests/AngleTest.cpp
+++ b/isis/tests/AngleTest.cpp
@@ -4,6 +4,7 @@
 #include "IException.h"
 
 #include <QString>
+#include <QDebug> 
 
 #include <gtest/gtest.h>
 
@@ -56,9 +57,13 @@ TEST(AngleTest, Mutators) {
   EXPECT_NEAR(angle.radians(), Isis::PI, TEST_EPSILON); 
 }
 
-// How do we test qDebug() << angle
-TEST(AngleOperators, DISABLED_QDebug) {
-  EXPECT_TRUE(true); 
+TEST(AngleOperators, QDebug) {
+  // Stream qDebug output to a string and then do a string comparison?
+  Isis::Angle angle(45.0, Isis::Angle::Degrees);
+  QString qdebug_output;
+  QDebug debug(&qdebug_output);
+  debug << angle;
+  EXPECT_STREQ(qdebug_output.toLatin1(), "0.785398 <radians> (45 <degrees>) "); 
 }
 
 
@@ -133,11 +138,20 @@ TEST(AngleOperators, LessThanOrEqualTo) {
   EXPECT_FALSE(angle2 <= angle1); 
 }
 
-// Can we test error messages with an EXPECT_THROW?? 
+
 TEST(AngleExceptions, NullAngle){
   Isis::Angle angle1(30., Isis::Angle::Degrees );
-  EXPECT_THROW(angle1 < Isis::Angle(), Isis::IException);
-  // TODO: use Jesse's code to test error message. 
+  
+  try {
+    angle1 < Isis::Angle();
+    FAIL() << "Expected an error";
+   }
+   catch(Isis::IException &e) {
+     EXPECT_EQ(e.toString().toLatin1(), "**PROGRAMMER ERROR** Cannot compare a invalid angles with the < operator." );
+   }
+   catch(...) {
+     FAIL() << "Expected IException: **PROGRAMMER ERROR** Cannot compare a invalid angles with the < operator, but got something else "; 
+   }
 }
     
 
--- a/isis/tests/AngleTest.cpp
+++ b/isis/tests/AngleTest.cpp
@@ -8,8 +8,6 @@
 
 #include <gtest/gtest.h>
 
-#define TEST_EPSILON 1e-8
-
 TEST(AngleTest, DefaultConstructor) {
   Isis::Angle angle;
   EXPECT_FALSE(angle.isValid());
@@ -21,7 +19,7 @@ TEST(AngleTest, DefaultConstructor) {
 
 TEST(AngleTest, NullInput_DegreeOutput) {
   Isis::Angle angle;
-  EXPECT_NEAR(angle.degrees(), Isis::Null, TEST_EPSILON);
+  EXPECT_DOUBLE_EQ(angle.degrees(), Isis::Null);
   EXPECT_FALSE(angle.isValid());
   EXPECT_STREQ(angle.toString().toLatin1(), ""); 
 }
@@ -29,7 +27,7 @@ TEST(AngleTest, NullInput_DegreeOutput) {
 
 TEST(AngleTest, DegreeInput_RadianOutput) {
   Isis::Angle angle(30., Isis::Angle::Degrees );
-  EXPECT_NEAR(angle.radians(), 30 * Isis::PI/180.0, TEST_EPSILON);
+  EXPECT_DOUBLE_EQ(angle.radians(), 30 * Isis::PI/180.0);
   EXPECT_TRUE(angle.isValid()); 
   EXPECT_STREQ(angle.toString().toLatin1(), "30.0 degrees");  
 }
@@ -37,7 +35,7 @@ TEST(AngleTest, DegreeInput_RadianOutput) {
 
 TEST(AngleTest, RadianInput_DegreeOutput) {
   Isis::Angle angle(30. * Isis::PI / 180.0, Isis::Angle::Radians );
-  EXPECT_NEAR(angle.degrees(), 30.0, TEST_EPSILON);
+  EXPECT_DOUBLE_EQ(angle.degrees(), 30.0);
   EXPECT_STREQ(angle.toString(false).toLatin1(), "30.0");  
 }
 
@@ -45,32 +43,31 @@ TEST(AngleTest, RadianInput_DegreeOutput) {
 TEST(AngleTest, CopyConstructor) {
   Isis::Angle angle(30., Isis::Angle::Degrees );
   Isis::Angle angleCopy(angle);
-  EXPECT_NEAR(angleCopy.degrees(), 30.0,TEST_EPSILON); 
+  EXPECT_DOUBLE_EQ(angleCopy.degrees(), 30.0);
 }
 
 
 TEST(AngleTest, Mutators) {
   Isis::Angle angle(30.0, Isis::Angle::Degrees );
   angle.setDegrees(180);
-  EXPECT_NEAR(angle.degrees(), 180.0, TEST_EPSILON);
+  EXPECT_DOUBLE_EQ(angle.degrees(), 180.0);
   angle.setRadians(Isis::PI);
-  EXPECT_NEAR(angle.radians(), Isis::PI, TEST_EPSILON); 
+  EXPECT_DOUBLE_EQ(angle.radians(), Isis::PI);
 }
 
 TEST(AngleOperators, QDebug) {
-  // Stream qDebug output to a string and then do a string comparison?
-  Isis::Angle angle(45.0, Isis::Angle::Degrees);
+  Isis::Angle angle(0.0, Isis::Angle::Degrees);
   QString qdebug_output;
   QDebug debug(&qdebug_output);
   debug << angle;
-  EXPECT_STREQ(qdebug_output.toLatin1(), "0.785398 <radians> (45 <degrees>) "); 
+  EXPECT_STREQ(qdebug_output.toLatin1(), "0 <radians> (0 <degrees>) "); 
 }
 
 
 TEST(AngleOperators, Assignment) {
   Isis::Angle angle(30.0, Isis::Angle::Degrees );
   angle = Isis::Angle(45.0, Isis::Angle::Degrees);
-  EXPECT_NEAR(angle.degrees(), 45.0, TEST_EPSILON); 
+  EXPECT_DOUBLE_EQ(angle.degrees(), 45.0);
 }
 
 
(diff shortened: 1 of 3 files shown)


isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+  EXPECT_NEAR(angle.radians(), Isis::PI, TEST_EPSILON); 
+}
+
+TEST(AngleOperators, QDebug) {
+  // Stream qDebug output to a string and then do a string comparison?
+  Isis::Angle angle(45.0, Isis::Angle::Degrees);
+  QString qdebug_output;
+  QDebug debug(&qdebug_output);
+  debug << angle;
+  EXPECT_STREQ(qdebug_output.toLatin1(), "0.785398 <radians> (45 <degrees>) ");
```
jessemapel: Comparisons of strings with decimal places makes me feel un-easy. Could this be tested by just setting a 1 radian angle and outputing it?

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+
+TEST(AngleOperators, LessThan) {
+  Isis::Angle angle1(30.0, Isis::Angle::Degrees );
+  Isis::Angle angle2(45.0, Isis::Angle::Degrees );
+  EXPECT_TRUE(angle1 < angle2);
+  EXPECT_FALSE(angle2 < angle1);
+  Isis::Angle epsilon(1.1920929e-10, Isis::Angle::Degrees);
+  EXPECT_TRUE(angle1 < angle1 + epsilon);
```
jessemapel: I don't think this epsilon part is required. The third case would be angle1 < angle1 == false

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// Added good and consistent test for "greater than" 
+TEST(AngleOperators, GreaterThan) {
+  Isis::Angle angle1(30.0, Isis::Angle::Degrees );
+  Isis::Angle angle2(45.0, Isis::Angle::Degrees );
+  EXPECT_TRUE(angle2 > angle1);
+  EXPECT_FALSE(angle1 > angle2); 
+  Isis::Angle epsilon(1.1920929e-10, Isis::Angle::Degrees);
+  EXPECT_TRUE(angle2 > angle2 - epsilon);
```
jessemapel: Again, I think the epsilon bit can be removed and just test angle1 > angle1

isis/tests/AngleTest.cpp
jessemapel: Similar to the LessThanOrEqualTo test

isis/tests/AngleTest.cpp
jessemapel: Similar to the LessThanOrEqualTo test

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+TEST(AngleExceptions, LessThanNullAngle){
+  Isis::Angle angle1(30., Isis::Angle::Degrees );
+  
+  try {
+    angle1 < Isis::Angle();
+    FAIL() << "Expected an error";
+   }
+   catch(Isis::IException &e) {
+     EXPECT_EQ(e.toString().toLatin1(), "**PROGRAMMER ERROR** Cannot compare a invalid angles with the < operator." );
```
jessemapel: I've had success using the following to test exception messages:

```
EXPECT_THAT(
      e.toString().toStdString(),
      ::testing::HasSubstr("Part of the error message string")
);
```

jessemapel: Make sure you add gmock as an include if you use this! `#include "gmock/gmock.h"`

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+  Isis::Angle angle(30.0, Isis::Angle::Degrees );
+  angle = 2.0 * angle; 
+  EXPECT_NEAR(angle.degrees(), 60.0, TEST_EPSILON); 
+}
+
+
+TEST(Angle, wrapValue){
+  Isis::Angle angle(30., Isis::Angle::Degrees);
+  EXPECT_EQ(angle.unitWrapValue(Isis::Angle::Degrees), 360);
+  EXPECT_NEAR(angle.unitWrapValue(Isis::Angle::Radians), 6.28318531, TEST_EPSILON);
```
jessemapel: Is there a reason why this doesn't use `2*Isis::PI`?

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+TEST(Angle, setAngleRadians){
+  Isis::Angle angle(0.0, Isis::Angle::Radians);
+  angle.setAngle(0.5, Isis::Angle::Radians);
+  EXPECT_NEAR(angle.radians(), 0.5, TEST_EPSILON);
+}
+
+
+TEST(Angle, QStringConstructor){
+  Isis::Angle angle(QString("-70 15 30.125"));
+  EXPECT_STREQ(angle.toString().toLatin1(), "-70.258368055556 degrees");
```
jessemapel: Can we do this comparison with doubles and actually compute what the value should be?

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+  Isis::Angle angle2(QString("  +70  30 11     "));
+  EXPECT_STREQ(angle2.toString().toLatin1(), "70.503055555556 degrees");
```
jessemapel: Same here

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+  Isis::Angle angle3(QString("100 00 00"));
+  EXPECT_STREQ(angle3.toString().toLatin1(), "100.0 degrees");
```
jessemapel: And here

jessemapel (commented):
Lots of small comments. One big whole picture question is can we use `EXPECT_DOUBLE_EQ` instead of all of the EXPECT_NEAR calls (https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#floating-point-comparison)? There is nothing in this test that is going to vary by more than a few bits and `EXPECT_DOUBLE_EQ` accounts for that.

isis/tests/AngleTest.cpp
jlaura: :+1:  to the idea of doing a fixed. no FP tests. As a follow on, are these debugs something that is valuable to test?

isis/tests/AngleTest.cpp
jlaura: Depending on how @jessemapel  and @kberryUSGS resolve the string containing FP comments, it might be good to update documentation to describe a consistent method for testing this.  I err on the side of wanting to avoid any kind of string comparison that contains a FP number due to the challenges it introduces.

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    FAIL() << "Expected an error";
+   }
+   catch(Isis::IException &e) {
+     EXPECT_EQ(e.toString().toLatin1(), "**PROGRAMMER ERROR** [100] is not a vaid input to Angle. It needs to be of the form: \"dd mm ss.ss\".");
+   }
+   catch(...) {
+     FAIL() << "Expected IException."; 
+   }
+}
```
jlaura, line 273: Can this be handled using the SUBSTR technique as above?

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+TEST(AngleExceptions, InvalidInput2){
+  try {
+    Isis::Angle angle(QString("70 11"));
+    angle.toString(); 
+
+    FAIL() << "Expected an error";
+   }
+   catch(Isis::IException &e) {
+     EXPECT_EQ(e.toString().toLatin1(), "**PROGRAMMER ERROR** [70 11] is not a vaid input to Angle. It needs to be of the form: \"dd mm ss.ss\".");
+   }
```
jlaura, line 287: Can this be handled using the SUBSTR technique as above?

isis/tests/AngleTest.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+TEST(AngleExceptions, InvalidInput3){
+  try {
+    Isis::Angle angle(QString("this 79 should 00 fail 0.111"));
+    angle.toString(); 
+
+    FAIL() << "Expected an error";
+   }
+   catch(Isis::IException &e) {
+     EXPECT_EQ(e.toString().toLatin1(), "**PROGRAMMER ERROR** [this 79 should 00 fail 0.111] is not a vaid input to Angle. It needs to be of the form: \"dd mm ss.ss\".");
```
jlaura: Can this be handled using the SUBSTR technique as above?

isis/tests/AngleTest.cpp
jessemapel: @jlaura They are used for some file outputs, so it's worth testing that they properly output the angle as a string.

krlberry: @jessemapel :+1: This can definitely be changed. My initial approach was to convert the whole existing unit test, as is (more-or-less) to gtest without changing what it tested. I do think it makes sense to change this test.

isis/tests/AngleTest.cpp
krlberry: @jessemapel Stuart and I were _very_ confused by why the existing unit test made this particular comparison. I would be more than happy to get rid of this if you agree that it's unneeded.

isis/tests/AngleTest.cpp
krlberry: See above, + yes.

isis/tests/AngleTest.cpp
krlberry: Ah, very good! I knew we'd talked about testing substrings, but couldn't find that under gtest -- I will go ahead and change all of these.

isis/tests/AngleTest.cpp
krlberry: Just because that's what the existing test did :) 

I'll change that, thanks for catching it!

isis/tests/AngleTest.cpp
krlberry: Yes, definitely!

isis/tests/AngleTest.cpp
krlberry: :+1:

isis/tests/AngleTest.cpp
krlberry: Just another thought -- this was testing two things before: (1) that the numerical value is correct and (2) that it's being converted to a string with the correct units (degrees) at the end. If I strip off the "degrees" and do the test with doubles, I may add another test to make sure that the correct units are printed out when it is converted to a string.

isis/tests/AngleTest.cpp
krlberry: Yes

isis/tests/AngleTest.cpp
krlberry: Yes

isis/tests/AngleTest.cpp
krlberry: Yes

isis/tests/AngleTest.cpp
krlberry: @jlaura @jessemapel If possible, I like the idea of testing numerical values with many decimal places and string and/or debug output separately. In cases where we are testing the string output, we could use a very simple number like "1".

isis/tests/AngleTest.cpp
jessemapel: Here's the docs on these matchers https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#more-string-assertions

isis/tests/AngleTest.cpp
jessemapel: :+1:

isis/tests/AngleTest.cpp
jessemapel: :+1:

isis/tests/AngleTest.cpp
krlberry: @jessemapel One complication with the "change it to 1 radian" here is that the angle's value is output in both radians and degrees. I don't think I can avoid something with decimal places unless I use 0.

krlberry: (currently updating this to use 0)

isis/tests/AngleTest.cpp
krlberry: I did get rid of this test in the updated version, but I think I figured out why this was in there in the first place. The original test actually did something like: 

```
  EXPECT_FALSE(angle1 < angle1 + epsilon); 
```

So perhaps the original test was testing something like: "this epilson is so small that when added to `angle1`, it doesn't increase its value to measurably more than `angle1`... I'm not sure why were were testing this, but it looks like that was what we were doing.

SgStapleton:
I'm a little confused. Which version of the AngleTest is going in? (vs. https://github.com/USGS-Astrogeology/ISIS3/pull/541)

isis/tests/AngleTest.cpp
krlberry: And just to document this -- I believe we decided to use something like: 

```
EXPECT_TRUE(Qstring.contains(substring from error message)
```

krlberry:
@SgStapleton This version of the Angle code is correct. The code in the `Spectel` PR is not up-to-date.

isis/tests/AngleTest.cpp
krlberry: Ended up using QString.contains() with `EXPECT_TRUE`

isis/tests/AngleTest.cpp
jessemapel: Sounds good :+1:

isis/tests/AngleTest.cpp
jessemapel: I really don't think that's worth testing, unless we are doing some weird internal tolerances in the comparison.

krlberry: Glad to hear it, as I did just get rid of this.

jessemapel (approved):

PR #533, opened 2018-11-07
Converted crop to a threaded object which can be called.

twilson271828:
I converted the ISIS3 crop application to an ISIS3 CropApp class. The functionality of the application is moved to a single thread.  The purpose of the original application is now to validate parameters that are passed into the class encapsulating the logic of the application.  This will make it easier for testing/debugging because now they are separate modules of logic instead of being co-mingled.  Validation of parameters in some ISIS3 applications can also be quite complex and non-trivial, so this is a good thing.

I have kept the code logic as simple as possible so that it can be easily replicated across applications.  Hopefully following this blueprint we can finally get rid of ProgramLauncher and pipeline (two of the touchiest applications within ISIS3).

### DIFF
--- a/isis/src/base/objs/CropApp/CropApp.cpp
+++ b/isis/src/base/objs/CropApp/CropApp.cpp
@@ -1,0 +1,229 @@
+#include "CropApp.h"
+
+#include <QString>
+#include <QThread>
+
+#include <cmath>
+
+#include "Cube.h"
+#include "ProcessByLine.h"
+#include "SpecialPixel.h"
+#include "LineManager.h"
+#include "FileName.h"
+#include "IException.h"
+#include "Projection.h"
+#include "AlphaCube.h"
+#include "Table.h"
+#include "SubArea.h"
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+
+
+int cropper::m_sline= -1;
+int cropper::m_linc=-1;
+int cropper::m_sband=-1;
+int cropper::m_ssample=-1;
+int cropper::m_sinc=-1;
+Cube * cropper::m_cube=NULL;
+LineManager * cropper::m_in = NULL;
+
+int cropper::m_osamples =-1;
+int cropper::m_olines=-1;
+int cropper::m_obands=-1;
+
+QString cropper::m_outputCubeName="";
+bool cropper::m_propspice = false;
+
+
+         CropApp::CropApp(QString &from, QString &to, int ssample,int nsamples,
+                          int sinc, int sline,int nlines,int linc,bool propspice){
+
+            m_fromCube =from;
+            m_toCube = to;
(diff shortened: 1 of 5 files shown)


isis/src/base/objs/CropApp/CropApp.h
```diff
@@ -0,0 +1,97 @@
...
+  static int m_sband;
+  static int m_ssample;
+  static int m_sinc;
+  static int m_osamples;
+  static int m_olines;
+  static int m_obands;
+  static Cube *m_cube;
+  static QString m_outputCubeName;
+  static bool m_propspice;
+  static LineManager *m_in;
```
jessemapel, line 69: Why are all of these static? If there are ever multiple cropper objects working in parallel, they will conflict over the static variables. These should just be public member variables.

isis/src/base/objs/CropApp/CropApp.cpp
```diff
@@ -0,0 +1,392 @@
...
+  cropper::cropper(int ssample,int nsamples,
+          int sinc, int sline,int nlines,int linc,bool propspice,QString to,QString from){
+
+
+
+
+    CubeAttributeInput inAtt(from);
+    m_cube = new Cube();
+    m_cube->setVirtualBands(inAtt.bands());
+    m_cube->open(from);
```
jessemapel, line 104: You already had to open this cube, It makes more sense to open it once and then just pass it into the cropper object. This also makes this app way more testable.

You can do the same thing with the output cube.

isis/src/base/objs/ProcessByBrick/ProcessByBrick.cpp
```diff
@@ -456,8 +456,8 @@ void ProcessByBrick::SetOutputRequirements(int outputRequirements) {
     bool haveInput = PrepProcessCubeInPlace(&cube, &brick);
 
     // Loop and let the app programmer work with the bricks
-    p_progress->SetMaximumSteps(brick->Bricks());
-    p_progress->CheckStatus();
+    //p_progress->SetMaximumSteps(brick->Bricks());
+    //p_progress->CheckStatus();
```
jessemapel, line 460: Why are these commented out?

twilson271828: The progress bar is instantiated on the main application thread.  The processing takes place on the object thread, and when it tries to access the main application thread it will cause seg faults or other weird behavior.  So I commented these out just to get things running, realizing that if we use threads in the future, functionality needs to be added to the ProcessByBrick (or Process class) to facilitate communication between threads.

isis/src/base/objs/CropApp/CropApp.cpp
twilson271828: Good point.

PR #536, opened 2018-11-07
PixelTest

ihumphrey:


isis/tests/PixelTests.cpp
```diff
@@ -0,0 +1,481 @@
...
+  // "Null" pixel
+  EXPECT_FLOAT_EQ(Isis::NULL4, Pixel::To32Bit(Isis::Null));
+  // HRS
+  EXPECT_FLOAT_EQ(Isis::HIGH_REPR_SAT4, Pixel::To32Bit(Isis::Hrs));
+  // HIS
+  EXPECT_FLOAT_EQ(Isis::HIGH_INSTR_SAT4, Pixel::To32Bit(Isis::His));
+  // LRS
+  EXPECT_FLOAT_EQ(Isis::LOW_REPR_SAT4, Pixel::To32Bit(Isis::Lrs));
+  // LIS
+  EXPECT_FLOAT_EQ(Isis::LOW_INSTR_SAT4, Pixel::To32Bit(Isis::Lis));
```
jessemapel: I feel like these should use just EXPECT_EQ because they should be converted to exactly this value. If it's off by a byte it's a different special pixel or not a special pixel at all!

isis/tests/PixelTests.cpp
```diff
@@ -0,0 +1,481 @@
...
+  // "Null" pixel
+  EXPECT_FLOAT_EQ(Isis::NULL4, Pixel(1, 2, 3, Isis::Null).To32Bit());
+  // HRS
+  EXPECT_FLOAT_EQ(Isis::HIGH_REPR_SAT4, Pixel(1, 2, 3, Isis::Hrs).To32Bit());
+  // HIS
+  EXPECT_FLOAT_EQ(Isis::HIGH_INSTR_SAT4, Pixel(1, 2, 3, Isis::His).To32Bit());
+  // LRS
+  EXPECT_FLOAT_EQ(Isis::LOW_REPR_SAT4, Pixel(1, 2, 3, Isis::Lrs).To32Bit());
+  // LIS
+  EXPECT_FLOAT_EQ(Isis::LOW_INSTR_SAT4, Pixel(1, 2, 3, Isis::Lis).To32Bit());
```
jessemapel: Should also use EXPECT_EQ

isis/tests/PixelTests.cpp
```diff
@@ -0,0 +1,481 @@
...
+  short s = 0.0;
+  EXPECT_DOUBLE_EQ(0.0, Pixel::ToFloat(s));
+
+  // unsigned short
+  unsigned short us = 0.0;
+  EXPECT_DOUBLE_EQ(Isis::NULL4, Pixel::ToFloat(us));
+
+  // float
+  float f = 0.0;
+  EXPECT_DOUBLE_EQ(0.0, Pixel::ToFloat(f));
```
jessemapel: These should use either EXPECT_EQ or EXPECT_FLOAT_EQ instead of EXPECT_DOUBLE_EQ because these are actually 32 bit floats.

isis/tests/PixelTests.cpp
ihumphrey: good point!

isis/tests/PixelTests.cpp
ihumphrey: done

isis/tests/PixelTests.cpp
ihumphrey: right, i went ahead and added EXPECT_FLOAT_EQ

PR #530, opened 2018-11-07
App to callable function conversion rough draft

Kelvinrr:
Changes:

* App directory source files containing `int main` are renamed to `main.cpp` so it's easier to distinguish the app code from library code. 
* CMake scripts updated to add apps as part of the ISIS3 library 
* closes #582 
* closes #581 
* Intermediate solution to #576, library code for those was moved into the objs folder for now 
* addresses #572, the two importfits were renamed to Nirs and Amica importfits since they are different enough and there is no obvious way to merge the two. 
* closes #571

PR #539, opened 2018-11-08
Refactoring stats application into a function

jessemapel:
This refactors it to the style from #530 

This has kelvin's changes in it, my changes are in the StatsFunc class and the stats.ccp file.


celbal:
Why are there changes to BundleSolutionInfo?

jessemapel:
@makaylas Those are from @Kelvinrr I'm not sure why they are there.

PR #542, opened 2018-11-08
Stats refactor 

jessemapel:
This refactors the stats application into a function with a c++ api and a separate function for writing out to a file.


isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+      // Write the results to the log
+      Application::Log(results);
+    }
+
+    return statsPvl;
+  }
+
+
+  void writeStatsStream(Pvl statsPvl, bool writeHeader, ostream *stream) {
```
jessemapel, line 136: Changed from us an ofstream to injecting and ostream. This way we can use a ostringstream in testing to get what would be written to the file.

isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,151 @@
...
+        else {
+          os->open(outFile.toLatin1().data(), ios::out);
+          writeHeader = true;
+        }
+        writeStatsStream(statsPvl, writeHeader, os);
+        delete os;
+        os = NULL;
+      }
+    }
+  }
```
jessemapel, line 81: This function still has a lot of logic in it that will be hard to test. One possible solution is using a worker object and then injecting it into this function. That way we can mock it out for testing purposes.

jessemapel:
This is part 2 of #539

isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    if ( ui.WasEntered("VALIDMAX") ) {
+      validMax = ui.GetDouble("VALIDMAX");
+    }
+
+    Pvl statsPvl = stats(inputCube, validMin, validMax);
+
+    for (int resultIndex = 0; resultIndex < statsPvl.groups(); resultIndex++) {
+      if (statsPvl.group(resultIndex).name() == "Results") {
+        Application::Log(statsPvl.group(resultIndex));
```
jessemapel, line 43: Create a dummy application object at the top of this function to make it so it doesn't segfault if run outside of an app

jessemapel:
I've rebased onto testing in #560

PR #545, opened 2018-11-09
Fixed two bug with AutoReg. One occurs with one dimensional pattern c…

celbal:
…ubes and the other causes the revious registration sample/line to be returned on sub-pixel registration failure. Fixes #5247 #5248


scsides (approved):
These all look good, and should make future autoregs more accurate

PR #546, opened 2018-11-09
Added Endian unit test

kaitlyndlee:
So far, I am only testing two methods. There are four more methods to test, but the output depends on the machine's byte order. I am not sure what the best way to test this is.


ihumphrey (approved):
Looks good!

kaitlyndlee:
We decided that it is not worth it to test the methods dependent on the architecture of the machine.

PR #560, opened 2018-11-14
Clean stats refactor branch

jessemapel:


isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+      delete stats;
+      stats = NULL;
+    }
+
+    return statsPvl;
+  }
+
+
+  void writeStatsStream(Pvl statsPvl, bool writeHeader, ostream *stream) {
```
ihumphrey: Could you potentially use reference parameters for Pvl and ostream? (would no longer have to allocate and manage heap memory for the ostream created around line 70). I'm not sure if the Pvl param could be a const ref, depending on the signatures of its accessor methods...

isis/src/base/objs/StatsFunc/StatsFunc.h
```diff
@@ -0,0 +1,18 @@
+#include <iostream>
+#include <vector>
+
+#include <QString>
+
+#include "Cube.h"
+
+namespace Isis {
+  extern void stats(std::vector<char*>&);
```
ihumphrey: Is there anywhere we should document these functions (not a priority, but a consideration we should have if we are providing a callable interface to our applications)?

isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "Pvl.h"
+#include "UserInterface.h"
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+  void stats(std::vector<char*>& args) {
+    // TODO is this line bad?
+    UserInterface ui("/work/users/jmapel/ISIS3/isis/src/base/apps/stats/stats.xml", args);
```
ihumphrey: Do you know if we decided to pass in the `vector<char*>` or the `UserInterface &`? Does having this UserInterface instantiation pose any problems?

isis/src/base/objs/StatsFunc/StatsFunc.cpp
```diff
@@ -0,0 +1,158 @@
...
+    Pvl statsPvl = stats(inputCube, validMin, validMax);
+
+    for (int resultIndex = 0; resultIndex < statsPvl.groups(); resultIndex++) {
+      if (statsPvl.group(resultIndex).name() == "Results") {
+        Application::Log(statsPvl.group(resultIndex));
+      }
+    }
+
+    delete inputCube;
+    inputCube = NULL;
```
ihumphrey: This is just a small suggestion, but we should probably be converting any old `NULL` macros into the c++11 `nullptr` keyword.

isis/src/base/objs/StatsFunc/StatsFunc.cpp
jessemapel: The Pvl can become a const ref, the ostream needs to be a pointer for polymorphism. For testing, we want it to be an ostringstream and for actual use we want it to be an ofstream.

isis/src/base/objs/StatsFunc/StatsFunc.h
jessemapel: Since these are part of the API I do need to document them!

isis/src/base/objs/StatsFunc/StatsFunc.cpp
jessemapel: I think I have the wrong commit pushed up here. I'll get that fixed.

isis/src/base/objs/StatsFunc/StatsFunc.cpp
jessemapel: :+1:

jessemapel:
Closing because this somehow got all messed up?

PR #563, opened 2018-11-14
Update isis3VarInit.py

rbeyer:
Unnecessary 'touch'ing, the '>' redirect will create the file if it doesn't exist.
Also, the first redirect should be the single '>'.  Otherwise, multiple runs of this program will continue appending to these files.  They still work, its just a lot of setting and unsetting for no good reason.


jessemapel (approved):
Looks good to me

PR #568, opened 2018-11-15
FileName gtest conversion

celbal:


isis/tests/FileNameTests.cpp
```diff
@@ -2,7 +2,12 @@
 
 #include <QString>
 
+#include <iostream>
+#include <fstream>
+#include <stdlib.h>
+
 #include <gtest/gtest.h>
```
jessemapel, line 9: You can remove the gtest include once you have gmock included

isis/tests/FileNameTests.cpp
```diff
@@ -74,10 +86,12 @@ TEST(FileName, Extension) {
...
-//TODO Waiting for GMock
-
-// TEST(FileName, Expanded) {
-// }
+TEST(FileName, Expanded) {
+  putenv("ISISTMPROOT=/testy/mc/test/face");
+  FileName file("$ISISTMPROOT/test.cub");
+  
+  EXPECT_EQ("/testy/mc/test/face/test.cub", file.expanded());
+}
```
jessemapel, line 92: Can you just use the ISISROOT environment variable here? You can just check that the expanded filename contains the ISISROOT variable and test.cub

isis/tests/FileNameTests.cpp
```diff
@@ -111,47 +124,117 @@ TEST(FileName, isQuestionMarksNoExtensionVersioned) {
...
+  fileStream.open("test000002.cub");
+  fileStream << "test";
+  fileStream.close();
+  
+  FileName file("test??????.cub");
+  EXPECT_EQ("./test000002.cub", file.highestVersion().expanded());
+  remove("test000001.cub");
+  remove("test000002.cub");
+  
+}
```
jessemapel, line 156: If you run into an exception during this test, the files won't get cleaned up. You can use a fixture's setup and teardown methods to setup the files and ensure that they get removed even if the test doesn't run to completion.

isis/tests/FileNameTests.cpp
```diff
@@ -111,47 +124,117 @@ TEST(FileName, isQuestionMarksNoExtensionVersioned) {
...
+  FileName fakeFile("test.cub");
+  EXPECT_FALSE(fakeFile.fileExists());
+  
+  remove("test000001.cub");
+}
 
-// TEST(FileName, Version) {
+//TODO Do we need to test this? All it does is call expanded, makes a QFileInfo and then make a QDir
+// We shouldn't be testing Qt
+// TEST(FileName, Dir) {
```
jessemapel, line 191: I agree don't test it

jessemapel (approved):
Just have a few comments, I think this is ready to go though.

isis/tests/FileNameTests.cpp
celbal: No because someone's ISISROOT may be set to s different directory and I don't want to override the current ISISROOT

isis/tests/FileNameTests.cpp
celbal: Honestly I don't use gmock so I shouldn't be including it. I'll remove gmock

isis/tests/FileNameTests.cpp
jessemapel: I mean you can just use the already set ISISROOT environment variable. That way you don't need to set another global environment variable.

celbal: Except that ISISROOT is going to expanded differently whenever someone has it set to something else

PR #570, opened 2018-11-16
Fixed isis3Startup scripts

### LINKED ISSUE #569: Incorrect default for $ISISROOT in 3.6.0 Startup Scripts
It looks like /usgs/pkgs/isis3.6.0/isis/scripts/isis3Startup.sh is using the incorrect default for `$ISISROOT` and this is causing the wrong bin directory to be set.  As I understand it, the correct default for $ISISROOT should be `/usgs/pkgs/isis3.6.0/install`

However, when I run `source /usgs/pkgs/isis3.6.0/isis/scripts/isis3Startup.sh`, my `$PATH` gets updated to include `/usgs/pkgs/isis3/isis/bin`, but this directory no longer exists in 3.6.0. As a result, I can't run any ISIS3 programs.

I haven't tested the analogous startup script for CSH, so I don't know if that script has the same issue. Either way, this should only affect internal USGS users because external users are expected to install 3.6.0 through `conda`. I have confirmed that this has broken PDS Services (MAP2 and POW).

jessemapel:
Fixes #569

### DIFF
--- a/isis/scripts/isis3Startup.csh
+++ b/isis/scripts/isis3Startup.csh
@@ -24,6 +24,8 @@
 #                       to "/usgs/cpkgs/isis3/testData"
 #       SEP 08 2017 - Christopher Combs - Removed setting of QT_PLUGIN_PATH as
 #                       an environment variable. Fixes #4617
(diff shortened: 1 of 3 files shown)


krlberry (approved):
:+1: Looks good.

Issue #575, opened 2018-11-16
Parabola class not used in ISIS3?


krlberry:
It looks like it was replaced with `PolynomialUnivariate`. I'm going to remove this and see if it breaks anything.

PR #577, opened 2018-11-16
Remove unused Parabola class

krlberry:
While looking for good classes for test refactoring, I noticed that the Parabola class is no longer used. It is removed in this PR.


krlberry:
Fixes #575

kaitlyndlee:
Should this also go into dev instead?

krlberry:
Yes, good point @kdl222 !

krlberry:
@kdl222 See PR #586

PR #578, opened 2018-11-16
Removed unneeded LineEquation includes

krlberry:
Removed unneeded LineEquation.h includes. Noticed that they were no longer used while looking for good unit tests to convert.


jessemapel:
Can this go onto dev?

krlberry:
Sure.

krlberry:
Oops, let me fix this so it can actually go into dev.

krlberry:
@jessemapel See: PR #580

PR #584, opened 2018-11-16
Refactored stats to a functions

jessemapel:


jlaura:
@jessemapel Looks like this needs to be updated with a pull from upstream. @kberryUSGS @SgStapleton @Kelvinrr @kdl222 Is it possible to get another review on this so we can merge? I believe this is a PR that has been reviewed one by @ihumphrey.

Kelvinrr (dismissed):

Kelvinrr:
The code itself looks fine, once updated from upstream, and I believe the base branch also needs to be changed, it can be merged.

Kelvinrr:
Can one of the admins verify this patch?

jlaura:
@jessemapel Can you merge in the upstream and then aim into the dev branch?

jessemapel:
Yeah, I'll get this done today

jessemapel:
Header files from the app directories are not being copied over. I've fixed that but now I'm getting errors in the standard libraries. It looks like something has gone very very wrong. So It'll take me a bit to debug.

twilson271828:
I'm getting the following compilation error:

`/scratch/pr584/isis/tests/StatsFuncTests.cpp:1:10: fatal error: StatsFunc.h: No such file or directory
 #include "StatsFunc.h"
`
I've been using this PR as a basis for my own app tests for hyb2onc2isis, and received an identical error.  What fix did you apply?  I think the CMakeLists.txt file needs to be modified, or the header file placed in a location where cmake is looking.

twilson271828:
Ooops...just read your last comment.  Ok good, now we're in the same place!

twilson271828:
Also, is the plan with this PR to functionalize every application within ISIS3 for the purposes of testing?   I think I missed out on some discussions when I was in Dallas.

jessemapel:
@twilson271828 I see two paths forward that are potentially complimentary.
1. We functionalize everything and rewrite all of the old app tests using gtest.
1. We work through high-priority apps and do more thorough refactor and testing.

jessemapel:
I think I've got this nailed down. For some weird reason, Maalin Space Science Systems made a file called limits.h in the mocuncompress code. This was clashing with the system limits.h file that's kind of important.

Why anyone would name a file the same as an important system file I do not know. Maybe it's some significantly older code.

jessemapel:
This is now ready to go. There are a handful of changes in odd places to account for adding all of the app includes.

1. The Stereo class in smtk was updated to coding standards and integrated into the ISIS3 library a several years ago, but the original was never removed. So, this removes that to avoid conflicts with two Stereo.h files.
1. mocuncompress was using some troublesome file names so I added mocuncompress to the start of the file names to avoid conflicts.
1. I added lots of include guards to headers that are now included in our general library headers that didn't have them before.

Kelvinrr (approved):

PR #585, opened 2018-11-16
Added Distance unit tests

kaitlyndlee:
I will be refactoring this test since there are areas where I can use parameters. I also made some comments about some questions I had.

### DIFF
--- a/isis/tests/DistanceTests.cpp
+++ b/isis/tests/DistanceTests.cpp
@@ -112,10 +112,10 @@ TEST(DistanceTests, SetNegativeDistance) {
 }
 
 
-// TEST(DistanceTests, ToString) {
-// 	Distance dist(1500500, Distance::Meters);
-// 	ASSERT_STREQ(dist.toString(), "1500500 meters");
-// }
+TEST(DistanceTests, ToString) {
+	Distance dist(1500500, Distance::Meters);
+	ASSERT_EQ(dist.toString(), "1500500.0 meters");
+}
 
 
 TEST(DistanceTests, IsValidTrue) {
--- a/isis/tests/DistanceTests.cpp
+++ b/isis/tests/DistanceTests.cpp
@@ -19,7 +19,7 @@ TEST(DistanceTests, DefaultConstructor) {
 TEST(DistanceTests, MetersConstructor) {
 	Distance dist(1500500, Distance::Meters);
 	EXPECT_EQ(dist.meters(), 1500500);
-	EXPECT_DOUBLE_EQ(dist.kilometers(), 1500.5);
+	EXPECT_EQ(dist.kilometers(), 1500.5);
 	EXPECT_DOUBLE_EQ(dist.solarRadii(), 1500500 / 6.9599e8);
 	EXPECT_EQ(dist.pixels(1), 1500500);
 }
@@ -27,7 +27,7 @@ TEST(DistanceTests, MetersConstructor) {
 
 TEST(DistanceTests, KilometersConstructor) {
 	Distance dist(1500.5, Distance::Kilometers);
-	EXPECT_DOUBLE_EQ(dist.kilometers(), 1500.5);
+	EXPECT_EQ(dist.kilometers(), 1500.5);
 	EXPECT_EQ(dist.meters(), 1500500);
 	EXPECT_DOUBLE_EQ(dist.solarRadii(), 1500500 / 6.9599e8);
 	EXPECT_EQ(dist.pixels(1), 1500500);
@@ -36,7 +36,7 @@ TEST(DistanceTests, KilometersConstructor) {
 
 TEST(DistanceTests, SolarRadiiConstructor) {
 	Distance dist(1, Distance::SolarRadii);
-	EXPECT_DOUBLE_EQ(dist.solarRadii(), 1);
+	EXPECT_EQ(dist.solarRadii(), 1);
 	EXPECT_DOUBLE_EQ(dist.meters(), 6.9599e8);
 	EXPECT_DOUBLE_EQ(dist.kilometers(), 6.9599e5);
 	EXPECT_DOUBLE_EQ(dist.pixels(1), 6.9599e8);
@@ -47,7 +47,7 @@ TEST(DistanceTests, PixelsConstructor) {
 	Distance dist(1500500, Distance::Pixels);
 	EXPECT_EQ(dist.pixels(1), 1500500);
 	EXPECT_EQ(dist.meters(), 1500500);
-	EXPECT_DOUBLE_EQ(dist.kilometers(), 1500.5);
+	EXPECT_EQ(dist.kilometers(), 1500.5);
 	EXPECT_DOUBLE_EQ(dist.solarRadii(), 1500500 / 6.9599e8);
 }
 
@@ -56,7 +56,7 @@ TEST(DistanceTests, PixelsPerMeterConstructor) {
 	Distance dist(1500500, 2);
 	EXPECT_EQ(dist.pixels(2), 1500500);
 	EXPECT_EQ(dist.meters(), 750250);
-	EXPECT_DOUBLE_EQ(dist.kilometers(), 750.25);
+	EXPECT_EQ(dist.kilometers(), 750.25);
 	EXPECT_DOUBLE_EQ(dist.solarRadii(), 750250 / 6.9599e8);
 }
 
@@ -64,7 +64,7 @@ TEST(DistanceTests, PixelsPerMeterConstructor) {
 TEST(DistanceTests, CopyConstructor) {
 	Distance origDist(1500.5, Distance::Meters);
 	Distance copiedDist(origDist);
-	ASSERT_DOUBLE_EQ(copiedDist.meters(), 1500.5);
+	ASSERT_EQ(copiedDist.meters(), 1500.5);
 }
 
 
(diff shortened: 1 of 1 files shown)


isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+TEST(DistanceTests, DefaultConstructor) {
+	Distance dist;
+	EXPECT_EQ(dist.meters(), Null);
+	EXPECT_EQ(dist.kilometers(), Null);
+	EXPECT_EQ(dist.pixels(), Null);
+	EXPECT_EQ(dist.solarRadii(), Null);
+}
+
+
+TEST(DistanceTests, MetersConstructor) {
```
kaitlyndlee, line 19: I am not sure if using parameters would make things nicer here. My only idea would be to use a Qlist of QLists where each QList element is made up of Distance::<unit_type> and every value the original distance gets converted to in all units. Each test has the conversions in different orders, so each test would need a different ordered list.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+
+TEST(DistanceTests, IsValidFalse) {
+	Distance dist;
+	ASSERT_FALSE(dist.isValid());
+}
+
+
+//should we do all units?
```
kaitlyndlee: Since we have checked above that we can convert other units to miles, and each "mathy" test converts all units into miles before calculating, is it safe to say that we don't need to test every combination of units in the tests below? I just used meters for everything.

kaitlyndlee:
There are two helper methods, distance() and setDistance(), that are used in the constructor and the getters. I skipped testing these, but there are exceptions that can be thrown if we actually call either one of these methods directly. Is it worth testing these methods?

isis/tests/DistanceTests.cpp
krlberry: This seems okay to me!

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+	Distance dist2(10, Distance::Meters);
+	EXPECT_FALSE(dist1 < dist2);
+	EXPECT_FALSE(dist2 < dist1);
+}
+
+
+TEST(DistanceTests, LessThanNull) {
+	try {
+		bool value = Distance() < Distance();
+		ASSERT_TRUE(value);
```
krlberry: What is this ASSERT_TRUE for if the comparison in the above line is supposed to throw an exception?

krlberry (approved):
@kdl222 If you want to test the exceptions thrown by `setDistance()`, you could call the `Distance::Distance(double distance, Units distanceUnit)` constructor with some kind of invalid units. It looks to me like you are already testing the "negative distance" case. I don't think there's a need to test `setDistance()` more directly, given how it is used. I don't think you need to test `distance()` any more unless you'd like to add an exception test, as it is already being tested via other functions.

krlberry:
One more comment -- since the values used are doubles, could you use `EXPECT_DOUBLE_EQ`, rather than `EXPECT_FLOAT_EQ` ?

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+	EXPECT_EQ(dist.pixels(), Null);
+	EXPECT_EQ(dist.solarRadii(), Null);
+}
+
+
+TEST(DistanceTests, MetersConstructor) {
+	Distance dist(1500500, Distance::Meters);
+	EXPECT_EQ(dist.meters(), 1500500);
+	EXPECT_FLOAT_EQ(dist.kilometers(), 1500.5);
+	EXPECT_FLOAT_EQ(dist.solarRadii(), 0.002155922);
```
jessemapel: Use EXPECT_DBL_EQ for these comparisons. For the solar radii test, use the actual calculation for the comparison `1500500 / 6.9599e8`. Whenever comparing double values, try to avoid comparing against literals that are not 100% exact.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+	EXPECT_EQ(dist.pixels(1), 1500500);
+}
+
+
+TEST(DistanceTests, KilometersConstructor) {
+	Distance dist(1500.5, Distance::Kilometers);
+	EXPECT_FLOAT_EQ(dist.kilometers(), 1500.5);
+	EXPECT_EQ(dist.meters(), 1500500);
+	EXPECT_FLOAT_EQ(dist.solarRadii(), 0.002155922);
```
jessemapel: same here, use the exact value to compare against.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+	EXPECT_EQ(dist.pixels(1), 1500500);
+}
+
+
+TEST(DistanceTests, SolarRadiiConstructor) {
+	Distance dist(0.002155922, Distance::SolarRadii);
+	EXPECT_FLOAT_EQ(dist.solarRadii(), 0.002155922);
+	EXPECT_FLOAT_EQ(dist.meters(), 1.5005e+06);
+	EXPECT_FLOAT_EQ(dist.kilometers(), 1500.5);
+	EXPECT_FLOAT_EQ(dist.pixels(1), 1.5005e+06);
```
jessemapel: For this, you should use something simple like 1 solar radii. That will make all of the numbers exact. 1 solar radii = 6.9599e8 meters = 6.9599e5 km = 6.9599e8 pixels (with 1m per pixel).

isis/tests/DistanceTests.cpp
jessemapel: See my other comment and use 1 solar radii = 6.9599e8 meters = 6.9599e5 km = 6.9599e8 pixels (with 1m per pixel) because all of the math will work out exactly.

isis/tests/DistanceTests.cpp
jessemapel: See my other comment and use 1 solar radii = 6.9599e8 meters = 6.9599e5 km = 6.9599e8 pixels (with 1m per pixel) because all of the math will work out exactly.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+
+TEST(DistanceTests, PixelsConstructor) {
+	Distance dist(1500500, Distance::Pixels);
+	EXPECT_EQ(dist.pixels(1), 1500500);
+	EXPECT_EQ(dist.meters(), 1500500);
+	EXPECT_FLOAT_EQ(dist.kilometers(), 1500.5);
+	EXPECT_FLOAT_EQ(dist.solarRadii(), 0.002155922);
+}
```
jessemapel, line 52: See my other comment and use 1 solar radii = 6.9599e8 meters = 6.9599e5 km = 6.9599e8 pixels (with 1m per pixel) because all of the math will work out exactly.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+
+TEST(DistanceTests, PixelsPerMeterConstructor) {
+	Distance dist(1500500, 2);
+	EXPECT_EQ(dist.pixels(2), 1500500);
+	EXPECT_EQ(dist.meters(), 750250);
+	EXPECT_FLOAT_EQ(dist.kilometers(), 750.25);
+	EXPECT_FLOAT_EQ(dist.solarRadii(), 0.0010779609);
```
jessemapel: See my other comment and use 1 solar radii = 6.9599e8 meters = 6.9599e5 km = 1.39198e9 pixels (with 2m per pixel) because all of the math will work out exactly.

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+
+TEST(DistanceTests, CopyConstructor) {
+	Distance origDist(1500.5, Distance::Meters);
+	Distance copiedDist(origDist);
+	ASSERT_FLOAT_EQ(copiedDist.meters(), 1500.5);
```
jessemapel: This can be ASSERT_EQ or EXPECT_EQ

isis/tests/DistanceTests.cpp
```diff
@@ -0,0 +1,295 @@
...
+	catch(...) {
+	   	FAIL() << "Expected error message: Negative distances are not supported";
+	}
+}
+
+
+// TEST(DistanceTests, ToString) {
+// 	Distance dist(1500500, Distance::Meters);
+// 	ASSERT_EQ(dist.toString(), "1500500 meters");
+// }
```
jessemapel: Why is this untested?

isis/tests/DistanceTests.cpp
jessemapel: :+1:

isis/tests/DistanceTests.cpp
jessemapel: I agree, this should be removed. We shouldn't be testing what that expression outputs because it throws an exception. If we are throwing an exception we should not also be honoring a specific return value.

jessemapel (changes requested):
I think better numbers for the test can result in a better test. There is also a weird inconsistency between using EXPECT and ASSERT. You also should be using EXPECT_DOUBLE_EQ instead of EXPECT_FLOAT_EQ.

isis/tests/DistanceTests.cpp
kaitlyndlee: I forgot to make a comment about that when writing the test. The only reason why I instantiate a bool and use ASSERT_TRUE is so we do not get a build warning saying that it is unused. I would not have put it in otherwise. Would it be better to take it out and have the build warning?

jessemapel: You can just remove the assignment and throw away the result of the comparison.

```
try {
  Distance() < Distance();
  FAIL() << "Failure Message";
}
```

kaitlyndlee: That is what I did at first and I get an "unused comparison" warning. We could turn this type of warning off.

jessemapel: Oh! This is fine then.

jessemapel (approved):

Issue #588, opened 2018-11-17
Copyright status of contributed material


rbeyer:
I can't find any statements about what copyright terms you accept material under.  Since ISIS is in the public domain, anything other than public domain contributions (with the exception of isolated 3rd party libraries, etc.) would complicate things.

Specifically, when I initiate a pull request, and you accept it, what's going on?  If I don't explicitly state that my contribution is in the public domain, it isn't (thanks Berne Convention Implementation Act of 1988).  That means that if a file or script was in the public domain (because it was written by USGS civil servants), and then I modified it (without attaching a copyright waiver statement), submitted a pull request, and you accepted it and merged it into your codebase, that file is now copyright me.

I think this should be addressed in two ways:

(1) There is currently no boilerplate public domain statement at the top of each file in the codebase, there needs to be.  Doesn't have to be long, but it needs to be there.

(2) Don't accept pull requests that don't have a copyright waiver.  You already have a history of having a comment block at the top of your files, which includes a description and then text blocks for each author.  In order to accept a pull request on a file, the submitter *MUST* add their name, contribution date, and description (just like your in-house authors do now) but should also add a copyright waiver that the Creative Commons CC0 public domain statement suggests, like so:

`Author:  Bob the Builder`
`Date:    2018-11-17`
`Description: Streamlined the program, improved documentation, etc.`
`To the extent possible under law, Bob the Builder has waived all copyright and related or `
` neighboring rights to their contribution to this file.  This work is published from`
`the United States.`

jessemapel:
I think that #591 will resolve this.

```
In jurisdictions that recognize copyright laws, the author or authors
of this software dedicate any and all copyright interest in the
software to the public domain.
```

jlaura:
I have put in a PR that adds PR templates to the repository.  Please see and comment on #596.

jessemapel:
Fixed by #596

PR #589, opened 2018-11-18
Make this program more pythonic

rbeyer:
Okay, there were two things that I wanted to do here, the first was that the direct system calls seemed unnecessary, when all of that can be accomplished via Python function calls.  I appreciate that this program was meant to be quick and dirty, I get that.  However, this program is one of the first pieces of USGS software that a user runs after installing conda.  The fact that it provided absolutely no feedback about what it was doing, and that its usage statement was actually factually incorrect (it doesn't actually set the environment variables, it just builds the files that conda uses to set them), made me want to fix it.

Also, I wanted to use this to test an idea I had about how to clearly signify that (1) the program itself is in the public domain (because of its USGS origin), and (2) that my updated contribution was also released under the public domain.

### DIFF
--- a/isis/scripts/isis3VarInit.py
+++ b/isis/scripts/isis3VarInit.py
@@ -5,56 +5,92 @@ import os
 import sys
 
 ##########################################################################################################
-#   Author:  Tyler Wilson
-#   Date  :  2018-10-05
-#   Description:  This script sets ISISROOT/ISIS3DATA/ISIS3TESTDATA for the user and is executed
-#   within the conda environment created for the ISIS3 installation.
-#   The data directory and test directory are optional command line arguments.  If the user chooses
-#   not to set them, they will both be placed created on the same level as the $ISISROOT directory
-#   within the conda environment.
+#
+#  This work is free of known copyright restrictions.  USGS-authored or produced data, 
+#  information, and software are in the public domain.
+#
+#
+#   Description:  This program builds the shell scripts that define the 
+#       ISISROOT/ISIS3DATA/ISIS3TESTDATA environment variables for the user 
+#       when the ISIS3 conda environment is activated, and clean up when it is
+#       deactivated.
+#
+#       The data directory and test directory are optional command line arguments.  
+#       If the user chooses not to set them, they will both be created in the
+#       $ISISROOT directory.
+#
 #   History:
-#       Author:  Tyler Wilson
+#       Author:  Tyler Wilson, USGS
+#       Date:    2018-10-05
+#       Description: Initial commit.
+#
+#       Author:  Tyler Wilson, USGS
 #       Date:    2018-11-01
 #       Description:  Removed a pair of lines which were causing output errors on Mac OS X and were not
 #                     required anyway.
 #
+#       Author:  Ross Beyer
+#       Date:    2018-11-
+#       Description: Streamlined the program, improved documentation, and made the directory and
+#                    file creation more `pythonic' rather than using system calls.
+#       To the extent possible under law, Ross Beyer has waived all copyright and related or 
+#       neighboring rights to his contribution to this file.  This work is published from
+#       the United States.
+#   
+#
 ##########################################################################################################
 
-parser = argparse.ArgumentParser(description='Usage:  ./isis3VarInit --data_dir <data dir path> --test_dir <test dir path')
-   
-isisroot = '$CONDA_PREFIX'
-data_dir='$CONDA_PREFIX/data'
-testdata_dir='$CONDA_PREFIX/testData'
+# There are still a lot of Python 2 installations out there, and if people don't have
+# their conda environment set up properly, the error message they'll get will be hard
+# to decipher.  This might help:
+assert( sys.version_info >= (3,2) ) # Must be using Python 3.2 or later, is conda set up?
 
-parser.add_argument("--data-dir",default= data_dir,help="ISIS3 Mission Data Directory")
-parser.add_argument("--test-dir",default=testdata_dir,help="ISIS3 Mission Test Data Directory")
 
+# This just wraps and reports on the directory creation:
+def mkdir( p ):
+    if os.path.exists( p ): print( 'Tried to create '+p+', but it already exists.' )
+    else:
+        os.makedirs( p )
+        print( 'Created '+p )
+    return
+
+
+# Set up and then parse the command line:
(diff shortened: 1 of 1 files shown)


isis/scripts/isis3VarInit.py
```diff
@@ -5,31 +5,68 @@
...
+def mkdir( p ):
+    if os.path.exists( p ): print( 'Tried to create '+p+', but it already exists.' )
+    else:
+        os.makedirs( p )
+        print( 'Created '+p )
+    return
 
+
+# Set up and then parse the command line:
+parser = argparse.ArgumentParser( description='This program builds shell scripts that define ISIS3 environment variables during conda environment activation and deactivation, and creates some directories.' )
```
jessemapel: I think that the description can be converted into a doc string for the script and then just referenced here.

isis/scripts/isis3VarInit.py
```diff
@@ -5,31 +5,68 @@
...
+
+parser.add_argument('-d','--data-dir', 
+                    default=os.environ['CONDA_PREFIX']+'/data',
+                    help='ISIS3 Data Directory, default: %(default)s' )
+parser.add_argument('-t','--test-dir',
+                    default=os.environ['CONDA_PREFIX']+'/testData',
+                    help='ISIS3 Test Data Directory, default: %(default)s')
 args=parser.parse_args()
+<<<<<<< HEAD
+=======
```
jessemapel: You have merge markers in here

isis/scripts/isis3VarInit.py
```diff
@@ -54,4 +91,33 @@
 os.popen("echo 'unset ISISROOT' >>"+isisroot+"/etc/conda/deactivate.d/env_vars.sh")
 os.popen("echo 'unset ISIS3DATA' >>"+isisroot+"/etc/conda/deactivate.d/env_vars.sh")
 os.popen("echo 'unset ISIS3TESTDATA' >>"+isisroot+"/etc/conda/deactivate.d/env_vars.sh")
+>>>>>>> upstream/dev
```
jessemapel: Another merge marker

jessemapel: It looks like this block was intended to be deleted

isis/scripts/isis3VarInit.py
```diff
@@ -5,31 +5,68 @@
...
 
-parser.add_argument("--data-dir",default= data_dir,help="ISIS3 Mission Data Directory")
-parser.add_argument("--test-dir",default=testdata_dir,help="ISIS3 Mission Test Data Directory")
+# This just wraps and reports on the directory creation:
+def mkdir( p ):
+    if os.path.exists( p ): print( 'Tried to create '+p+', but it already exists.' )
+    else:
+        os.makedirs( p )
+        print( 'Created '+p )
+    return
```
jessemapel, line 55: I am confused about why this is needed. You can just use `exist_ok=true` on the [makedirs](https://docs.python.org/3/library/os.html?highlight=makedirs#os.makedirs) calls. Logging if the file exists already or not seems kind of pointless.

isis/scripts/isis3VarInit.py
```diff
@@ -5,31 +5,68 @@
...
 #       Description:  Removed a pair of lines which were causing output errors on Mac OS X and were not
 #                     required anyway.
 #
+#       Author:  Ross Beyer
+#       Date:    2018-11-
+#       Description: Streamlined the program, improved documentation, and made the directory and
+#                    file creation more `pythonic' rather than using system calls.
+#       To the extent possible under law, Ross Beyer has waived all copyright and related or 
+#       neighboring rights to his contribution to this file.  This work is published from
+#       the United States.
```
jessemapel: Is this not covered by our [license](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/license.txt)? If it is not, then we need to update our license. @jlaura

jessemapel (changes requested):
I'd been thinking about removing the popen calls for a little while on this. I'm glad someone got to it! Just needs some cleanup and I have a couple questions.

isis/scripts/isis3VarInit.py
rbeyer: Ah, that's an even better solution.  I'll make it in the next pull.

isis/scripts/isis3VarInit.py
rbeyer: Do'h!  The danger of doing software development on the weekend, and being interrupted by the kids.  So let's nix this pull request, and I'll clean out the merge markers and make the other changes, and submit a new pull request.

isis/scripts/isis3VarInit.py
rbeyer: Yeah.

isis/scripts/isis3VarInit.py
rbeyer: You absolutely can, and that's the way I first wrote it (and why the assert  at the top is for 3.2, where exist_ok first shows up for makedirs).  However, after I did that, I realized that the program wasn't providing any feedback to the user about what the heck it was doing.  You just ran the program, and then it gave you a prompt, zero information for the user.  Did it work, did it quietly fail somehow?  Who knows?  So when I decided I wanted to provide some reporting about what it was doing, and I wanted to provide the user information about what already existed or not, that's when I made this little wrapper.  If you prefer the silent treatment, I can take it out.

isis/scripts/isis3VarInit.py
rbeyer: I don't think it is.  So let's resolve #588 and #591, and then I'll do that.

isis/scripts/isis3VarInit.py
jessemapel: I'm fine with keeping this wrapper then

isis/scripts/isis3VarInit.py
jessemapel: You can keep pushing to the dev branch on your fork and this PR will automatically update. No need to put in a new PR.

jessemapel (approved):

isis/scripts/isis3VarInit.py
rbeyer: So I used the unlicense model in a recent commit.

isis/scripts/isis3VarInit.py
twilson271828: Nice changes!  Yes, please do not put in another PR.

PR #594, opened 2018-11-19
Removed old license, added new unlicense

jessemapel:
Fixes #591

PR #599, opened 2018-11-20
Fixed issue with the License change

jessemapel:
#594 broke the build because cmake was looking for the license file in the old location.


krlberry (approved):
:+1:

SgStapleton:
Looks good to go!

PR #602, opened 2018-11-21
Statistics unit test

twilson271828:


isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,143 @@
...
+    s.AddData(20.0);
+    s.AddData(30.0);
+    
+
+    //Values were computed in R
+    double mu =13.3333333333333;
+    double std = 20.8166599946613;
+    double var = 433.333333333333;
+    
+    double rms = 21.6024689946929;
```
jessemapel: There's only three data points. I think it would be better to just do the math right here.

isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+
+
+TEST(Statistics, HandCalculations) {
+
+    Statistics s;
+    
+    s.AddData(-10.0);
+    s.AddData(20.0);
+    s.AddData(30.0);
```
jessemapel, line 74: These values are going to result in messy output values. Why not do something simple like 10, 20, 30?

isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,143 @@
...
+    EXPECT_NEAR(s.SumSquare(),1400.0,TEST_EPSILON);
+    EXPECT_NEAR(s.ValidPixels(),3.0,TEST_EPSILON);
+    EXPECT_NEAR(s.Rms(),rms,TEST_EPSILON);    
+    EXPECT_NEAR(s.Minimum(),-10.0,TEST_EPSILON);
+    EXPECT_NEAR(s.Maximum(),30.0,TEST_EPSILON);
+    EXPECT_NEAR(s.ZScore(1.0),(1-mu)/std,TEST_EPSILON);    
+    double percent = 99.5;    
+    double k = sqrt(1.0 / (1.0 - percent / 100.0));
+    EXPECT_NEAR(s.ChebyshevMinimum(),mu -k*std,TEST_EPSILON);
+    EXPECT_NEAR(s.ChebyshevMaximum(),mu +k*std,TEST_EPSILON);
```
jessemapel: I really like these comparisons where the math is actually done.

isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    double a[10];
+    a[0] = 1.0;
+    a[1] = 2.0;
+    a[2] = 3.0;
+    a[3] = NULL8;
+    a[4] = HIGH_REPR_SAT8;
+    a[5] = LOW_REPR_SAT8;
+    a[6] = HIGH_INSTR_SAT8;
+    a[7] = LOW_INSTR_SAT8;
```
jessemapel: Is there a reason why you're using the ___8 names instead of just Null, Hrs, ... ?

isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,143 @@
...
+    EXPECT_NEAR(t.LisPixels(),1.0,TEST_EPSILON);
+    EXPECT_NEAR(t.LrsPixels(),1.0,TEST_EPSILON);
+    EXPECT_NEAR(t.HisPixels(),1.0,TEST_EPSILON);
+    EXPECT_NEAR(t.HrsPixels(),1.0,TEST_EPSILON);
+
+    EXPECT_NEAR(t.OutOfRangePixels(),0.0,TEST_EPSILON);
+    EXPECT_NEAR(t.OverRangePixels(),0.0,TEST_EPSILON);
+    EXPECT_NEAR(t.UnderRangePixels(),0.0,TEST_EPSILON);
+ 
+}
```
jessemapel, line 179: This test seems to be testing valid range and special pixels. Those two cases should be tested separately.

jessemapel (changes requested):
It looks like there are still untested methods

krlberry (changes requested):
Could you change all the `EXPECT_NEAR`s with `TEST_EPSILON`s to `EXPECT_DOUBLE_EQ`? We decided to use this macro instead.

isis/tests/Statistics.cpp
twilson271828: No reason.  I changed them to Hrs,Lrs,...etc.

isis/tests/Statistics.cpp
```diff
@@ -0,0 +1,234 @@
...
+    s.AddData(30.0);
+
+    //Values were computed in R
+    double mu = 13.333333333333334;
+    double std = 20.816659994661325;
+    double var = 433.33333333333331;
+    
+    double rms = 21.602468994692867;
+ 
+    FileName xmlFile("./Statistics.xml");
```
jessemapel, line 198: Instead of opening a file and writing to it, you can write directly to a QString with QXmlStreamWriters.https://stackoverflow.com/questions/6592063/qstring-to-xml-in-qt

jessemapel: And then read from it the same way.

Kelvinrr:
Can one of the admins verify this patch?

twilson271828:
I am working on Hayabusa 2 at present, and it's going very well so I will try to finish addressing
the changes requested this afternoon.

jlaura:
@twilson271828 What is the status on this? If the changes can't/won't be made can we close the PR?

jessemapel:
Merged in #3291

PR #605, opened 2018-11-21
Modfying TestUtilities.h and associated files

SgStapleton:
Updating TestUtilities.h and associated files to account for AssertIException to be broken into AssertIExceptionMessage and AssertIExceptionError


krlberry (approved):
:+1: This looks good!

PR #606, opened 2018-11-26
Fixed unit tests not being build

jessemapel:


krlberry (approved):
:+1: I was looking at this change from the testing merge after the nightly failures, too! Glad you figured out what the problem was and how to fix it.

PR #608, opened 2018-11-26
fixed hical having the wrong xml file

jessemapel:


jessemapel:
This should fix nightly builds for prog28 and prog29

SgStapleton (approved):
Looks like a straight-forward change

PR #645, opened 2018-12-14
Added AlphaCube group to outputlabel in the event the image is cropped.

twilson271828:
## Types of changes

## Checklist:

## Licensing


jlaura:
@scsides Can you put some comments onto this PR about areas that are of testing concern? The other contributors on this repo can then weigh in on methods of testing those areas.

twilson271828:
@jlaura I pinged Jesse to review this since I based it on his stats app tests.

scsides:
@USGS-Astrogeology/developers 
This Hayabusa2 sprint is currently envolves two types of tests:
1) unit test for the camera model
2) application tests for the ingestion and calibration programs

For the unit tests, it seems reasonable to convert the test to a gtest. However, all of the camera tests in ISIS3 use cube files. I will be attempting to minimize the size of these files first and looking into what it takes to eliminate them. The labels are what is actually used, so mocking them seems reasonable.

The application tests for both the ingestion program and the calibration program also require cube files at this time. Converting these tests to gtest is not a trivial effort, and there is reasonable concern regarding mocking of complex files. The mocking of PDS files will require changes to the ProcessImportPDS class. Not impossible but needs to be well thought out.

Thoughts:
Test only label output not cube DNs. NOT a good idea for some cases
Mock the input and output cubes and test a subset of DNs, mostly edge cases
This is a big problem that needs far more attention outside of little projects 
The testing sprints did not solve these problems
Reducing the test data volume is a high priority. Do it carelessly could be costly.

jessemapel (commented):
@scsides I agree that there is going to need to be some amount of file to test this. Between the Process classes and how the calibration program uses Cubes, it would take a significant amount of work to get completely away from them.

On the ingestion app:
I think that 90% of it can be tested via gtest. All of the label translation and AlphaCube group creation can be split into separate functions that don't require any files and can be easily tested by gtest. This way all of the valid input cases can be identified from the SIS and then tested them without needing lots of different input files. The remainder of the ingestion app is then ImportFits behavior. I don't think it's up to this app's test to cover that. I'm inclined to cover it with at least one test of Hayabusa2 because I don't think ImportFits is properly covered, but this app doesn't do anything special with it.

On the calibration app:
About half of the logic tested without any actual files.

In Hyb2OncCalUtils.h:
* isis2mat is already testable with a mock cube
* mat2isis and translate can be modified so they take a pointer to an output cube instead of an output file name. Then, the output cube can be mocked.

In the application main the bulk of the code will be hard to test in gtest. The utility functions and the process function could be tested but the app would need to be completely refactored to remove global variables. I think it's a good idea to eventually remove all of those global variables, I'm not sure if there is time for that at the moment though.

On the camera unit tests:
Right now cameras cannot be tested without an actual image. I don't think there's anything we can do here at the moment.

All in all, I think there's a few places some things can be made more testable without much work, but a lot of it will still need to be tested by just running the apps.

PR #657, opened 2018-12-18
Phocube will export RA and declination planes even if the point is not on the planet. Fixes #4446

kaitlyndlee:
## Description
phocube will export RA and declination planes regardless if the point is in outer space or not.

## Related Issue
https://isis.astrogeology.usgs.gov/fixit/issues/4446

## Motivation and Context
It is useful to export RA and declination even when a pixel falls off body. See ticket for more information.

## How Has This Been Tested?
Lynn Weller tested changes using a Messenger flyby image. I also tested by running phocube twice on the same cube, once with the changes and once without. This way I could test if only the RA and declination bands were affected. Added makefile test with off body pixels to test that RA and dec are exported.

## Types of changes

## Checklist:

## Licensing


isis/src/base/apps/phocube/main.cpp
```diff
@@ -447,15 +441,35 @@ void phocube(Buffer &out) {
...
+        }
+
+        for (int b = (skipDN) ? 1 : 0; b < bandsToProcess; b++) {
           out[index] = Isis::NULL8;
           index += 64 * 64;
         }
       }
+
+      // Export the RA and declination regardless if the point is in outer space or not.
+      if (ra) {
```
krlberry: I wasn't sure about having separate if statements and associated variables `ra` and `declination` initially since ra and dec are set from the same `RADEC` user parameter, but it is consistent with the way `BodyFixedX`, `BodyFixedY`, `BodyFixedZ` are handled in this application, so :+1:!

krlberry (approved):

isis/src/base/apps/phocube/tsts/off_body/Makefile
```diff
@@ -0,0 +1,10 @@
+# Tests RA and declination bands to make sure values are displayed even if not on the planet.
+APPNAME = phocube
+
+include $(ISISROOT)/make/isismake.tsts
+
+commands:
+	 $(APPNAME) from=$(INPUT)/EW0131773041G.cal.cub \
```
krlberry: Would it be useful to also check the cube label for this test?

isis/src/base/apps/phocube/phocube.xml
```diff
@@ -302,6 +302,10 @@ xsi:noNamespaceSchemaLocation=
 		<change name="Kaj Williams" date="2017-06-09">
       Renamed albedo to albedoRank, and morph (or morphology) to morphRank (or MorphologyRank). Ref #4008.
     </change>
+    <change name="Kaitlyn Lee" date="2018-03-10">
+      RA and decination are exported regardless if the point is not on the planet. RA and dec bands
```
krlberry: spelling: decination -> declination (or can just say DEC)

krlberry (changes requested):
Found one typo in documentation, otherwise, looks good!

isis/src/base/apps/phocube/tsts/off_body/Makefile
kaitlyndlee: Would we be checking the label to see if the BandBin group is ordered correctly or is there another reason why we should? I think this is a good idea either way. Quick question: Does cubediff compare labels?

isis/src/base/apps/phocube/main.cpp
```diff
@@ -205,20 +205,19 @@ void IsisMain() {
...
   if (morphologyRank) name += "Morphology Rank";
   if (albedoRank) name += "Albedo Rank";
-  if (ra) name += "Right Ascension";
-  if (declination) name += "Declination";
   if (bodyFixedX) name += "Body Fixed X";
   if (bodyFixedY) name += "Body Fixed Y";
   if (bodyFixedZ) name += "Body Fixed Z";
-
+  if (ra) name += "Right Ascension";
+  if (declination) name += "Declination";
```
jessemapel: Why have the bands been re-ordered? This could have down-stream impacts on people's existing processing tools that depend upon explicit band number output from phocube. IE, stats from=phoed.cub+4 may have different output now.

isis/src/base/apps/phocube/main.cpp
```diff
@@ -447,15 +441,35 @@ void phocube(Buffer &out) {
...
       }
-      // Trim outerspace
+
+      // Trim outer space except last two bands when RA and declination are selected
+      // We always skip two because RA and declination are selected together.
       else {
-        for(int b = (skipDN) ? 1 : 0; b < nbands; b++) {
+        int bandsToProcess;
+        if (ra && declination) {
+          bandsToProcess = nbands - 2;
```
jessemapel: Are ra and declination required to be together by the app xml?

isis/src/base/apps/phocube/tsts/off_body/Makefile
jessemapel: I believe that the app test cubediff only checks some of the label. Generally, if you want to check things in the cube label, you should use catlab to output it as a PVL.

isis/src/base/apps/phocube/main.cpp
kaitlyndlee: Yes.

isis/src/base/apps/phocube/tsts/off_body/Makefile
krlberry: `cubediff` does not check the cube label at all -- just the DNs. :+1: on needing to catlab the cube label out to a pvl to check the label.

krlberry (dismissed):

isis/src/base/apps/phocube/main.cpp
kaitlyndlee: Before we write values to each band, we check to see if the pixel is off body. If it is, then we used to right out NULL on each band. If it is on the planet, then we assign the pixel value for the corresponding keyword band by band. It was easier to move the RA and DEC bands to the end and outside of the if else clauses to check if we are off body than to refactor how the code previously wrote to each band.

krlberry (commented):
Sorry, didn't mean to 'approve' yet -- hit the wrong button and couldn't figure out how to undo it.

isis/src/base/apps/phocube/main.cpp
kaitlyndlee: Now thinking about it, there might be a way that does not require that much refactoring. We would just have to keep track of which bands RA and DEC are and if the pixel is off body, then skip to the RA and DEC bands, write out the values to those bands, and write out NULL for the rest of the bands.

scsides: Scripts have a tendency to use band numbers instead of searching for the correct band name and then getting the corresponding band number. Moving these band will undoubtedly break some scripts. I suggest adding the necessary code to keep the bands in the original order.

kaitlyndlee:
See https://github.com/USGS-Astrogeology/ISIS3/issues/2466 for question about global variable usage.

scsides (approved):
This looks good for merging.

jessemapel:
Does this need to be released now or is it waiting until 3.7 in May?

PR #3160, opened 2019-03-13
Added LatitudeTest using Gtest

kaitlyndlee:
## Description

Converted the Latitude makefile unit test to gtest. I also added a few comments to Angle and Latitude for clarification on a line of code.

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/3144

## Motivation and Context

Work contributed to effort to convert all tests to gtest.

## Types of changes

## Checklist:

## Licensing


kaitlyndlee:
I used a fixture for the Mapping groups, but I could have used gmock. Do we have a standard for mocking vs. using fixtures?

jlaura:
@kdl222 That is a really good question. @Kelvinrr @kberryUSGS @SgStapleton @scsides @acpaquette @twilson271828 @jessemapel - Thoughts on this? Seems that we should come up with a rule / rule of thumb.

isis/src/base/objs/Angle/Angle.cpp
```diff
@@ -225,7 +225,8 @@ namespace Isis {
       IString msg = "Cannot compare a invalid angles with the < operator";
       throw IException(IException::Programmer, msg, _FILEINFO_);
     }
-
+    // != comparison allows for angles that are considered equal to be treated
+    // as being equal. == operator override uses qFuzzyCompare().
```
jlaura, line 229: I am confused by this. Is this saying that the code makes != equal == in some cases?

jlaura: never mind - I see what this is doing...

scsides:
From what I've seen so far, the fixtures are simpler and easily read. When the amount of data is small and repeated they work great. Large data structures, cubes for example, may fit mocking better.

jessemapel:
I agree with @scsides. If it's a simple object like a PVL object, it should be done as a fixture. If it's a complex object that may be hard to properly initialize, then it should be mocked.

SgStapleton:
Cubes are a clear choice to mock. Would it be worth compiling a list of objects that are better-off mocked? This would not be an exhaustive list, of course, but it would at least serve as a reference for test developers when trying to make those decisions.

jessemapel:
> Cubes are a clear choice to mock. Would it be worth compiling a list of objects that are better-off mocked? This would not be an exhaustive list, of course, but it would at least serve as a reference for test developers when trying to make those decisions.

The googlemock way to do this is to have a central collection of mock objects that can be re-used, maybe `IsisMocks.h`. Anything in that collection should be mocked.

jessemapel (approved):

PR #3185, opened 2019-03-27
Removed extra exceptions writing controlnets

### LINKED ISSUE #2660: QNET Writes Out Bad Control Net, Gives Error, Refuses to Save Properly, Work Lost!
---

Author Name: **Stuart Robbins** (Stuart Robbins)

Original Assignee: Jesse Mapel

---

Since I updated to the latest version of ISIS, I will occasionally get an error in QNET that says, "Error saving control network.  **I/O ERROR** Failed writing control network to file [{path}]. **I/O ERROR** Can't write control net file."  Where {path} is the path and name of the net.

The network is perfectly fine.  It had literally just saved it a minute ago.  However, this error will signify that it has written out a partial network that can't be opened, and I can't save nor save-as somewhere else under a different name to write out the network properly.  Meaning all work is lost on that network with that name, unless there's a backup.  I've ben trying to remember to copy my network file just before I save, but with the latest just now and forgetting to do so, I'm logging this bug.

It is literally random -- it doesn't seem to happen when doing anything specific other than saving.  I'm on a Mac, it happens under OSX 10.11.4 and 10.13.5.

jessemapel:
Removed new exceptions introduced in controlnetversioner refactor.

## Description

When we moved the conversion from control point to protobuf message into the versioner, we also added several new exceptions checking for unusual values. Previously, unusual values were just reset to default. I changed all of them except for an empty pointID back to defaulting because they will not cause an issue. An empty pointID will causes serious issues because that is how the point is identified in the network graph and an empty string will cause problems there.

## Related Issue

hopefully fixes #2660 

## Motivation and Context

Control nets were getting obliterated and work couldn't be saved.

## How Has This Been Tested?

I tried to reproduce the error and could not. Full test suite passed.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/control/objs/ControlNetVersioner/ControlNetVersioner.cpp
+++ b/isis/src/control/objs/ControlNetVersioner/ControlNetVersioner.cpp
@@ -1803,10 +1803,7 @@ namespace Isis {
           pointType = ControlPointFileEntryV0002_PointType_Fixed;
           break;
         default:
-          QString msg = "Unable to create ProtoPoint [" + toString(protoPoint.id().c_str())
-                        + "] from file. Type enumeration ["
-                        + toString((int)(controlPoint->GetType())) + "] is invalid.";
-          throw IException(IException::Programmer, msg, _FILEINFO_);
+          pointType = ControlPointFileEntryV0002_PointType_Free;
           break;
       }
       protoPoint.set_type(pointType);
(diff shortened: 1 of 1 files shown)


krlberry:
:+1: This looks good to me. Is it worth adding a "potential future work" git issue to investigate the underlying problem that may be causing this -- how does the control network end up in a state such that these exceptions are being thrown? Or perhaps that is covered by #2660.

krlberry (approved):
:+1: Does what was discussed in the stand-up as a potential solution to a problem that can't be replicated consistently.

PR #3343, opened 2019-06-27
Minor Update to Fix Voyager Bundle Adjustment

dcookastro:
Added ability to turn off back-of-planet test during bundle adjustment and fixed minor bugs

## Description
Added an option to not do the back-of-planet test so that it can be turned off during bundle adjustment.
  Added optional bool argument to CameraGroundMap::GetXY and RadarGroundMap::GetXY to control 
    whether the test is done or not
  Change the variable name of the angle variable in CameraGroundMap::GetXY to cosAngle to be more descriptive
  Modified call to GetXY in BundleAdjust::ComputePartials, ControlPoint::ComputeResiduals, and ControlPoint::ComputeResidualsMillimeters to not do back-of-planet test
  Changed test value for valid distance/angle from DBL_EPSILON to 1.0e-50 in SurfacePoint methods
    MetersToLongitude, GetLatSigmaDistance, GetLocalRadius, and GetLonSigmaDistance.

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/2591

## Motivation and Context
These changes fix known problems with running jigsaw on control nets imported from Isis2.

## How Has This Been Tested?
Known cases that failed ran successfully after the changes.

All system tests were run and passed excluding existing problems with the current nightly build.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [ x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.
- [ x] All new and existing tests passed.

## Licensing

- [x ] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.


isis/src/base/objs/CameraGroundMap/CameraGroundMap.cpp
```diff
@@ -193,28 +195,42 @@ namespace Isis {
...
-    if (angle > 1) {
-      emission = 0;
-    }
-    else if (angle < -1) {
-      emission = 180.;
-    }
-    else {
-      emission = acos(angle) * 180.0 / Isis::PI;
-    }
+    if (test == true) {
```
jessemapel: This should just be if (test)

isis/src/base/objs/CameraGroundMap/CameraGroundMap.cpp
```diff
@@ -254,6 +270,7 @@ namespace Isis {
     SurfacePoint spoint(Latitude(lat, Angle::Degrees),
                         Longitude(lon, Angle::Degrees),
                         Distance(radius, Distance::Meters));
+    // This is called by the application socetlinescankeywords
```
jessemapel: This should be part of the function documentation, maybe an @see tag?

isis/src/base/objs/RadarGroundMap/RadarGroundMap.h
```diff
@@ -117,7 +120,7 @@ namespace Isis {
       virtual bool SetGround(const Latitude &lat, const Longitude &lon);
       virtual bool SetGround(const SurfacePoint &surfacePoint);
       virtual bool GetXY(const SurfacePoint &spoint, double *cudx,
-                         double *cudy);
+                         double *cudy, bool test=false);
```
jessemapel, line 123: This is going to be accessed polymorphically through the parent CameraGroundMap class which means `test` will default to true instead of false.

https://stackoverflow.com/questions/5273144/c-polymorphism-and-default-argument

I'm not sure what the correct solution is here.

jessemapel (commented):
Browsing the code base, it looks like the places where GetXY gets called are:

BundleAdjust
ControlPoint
socetlinescankeywords

The new argument is set to false everywhere but socetlinescankeywords. Should it be set false there? If so, does it make sense to just remove the check completely because it won't ever get called with true?

dcookastro:
Jesse,

Thanks for reviewing my changes.  The reason I added the argument was to
keep the test for socetlinescankeywords.
I fixed the (test == true) statement and added the @see statement. I
realize the default will come from the parent, but I wanted to indicate
that it should be set to false.  I forced it to be false in all the
existing Isis calls.  Would it be better just to put a comment instead?

jessemapel:
What is the current functionality in the radar ground map? It doesn't attempt to do any type of checking? I would be tempted to just ignore the argument in the radar ground map and document as such. That doesn't seem like a good solution either though.

Right now, anything that calls camera.groundmap().GetXY() for a radar sensor will succeed, but with this change, that code will hit the new exception.

dcookastro:
My motivation in adding the exception was to alert the programmer that the
test was not being applied despite the bool being set to true. The current
functionality of radar ground map is to not do the back-of-planet test or
care whether it was done or not. I guess I wanted to make sure anyone using
the method thought about what they were doing, since the behavior differs
from the parent class. I agree this is not a good solution.  Ken and I
discussed this before he left too, and I could not come up with a good
solution.

jessemapel:
I'm okay with going forward with this then.

jessemapel (approved):

scsides (approved):
Agree with the conversation about testing the backside. Can't say I can think of a better way either. Everything else looks good.

scsides:
sgstapleton not available, so I'll go ahead and merge

PR #3350, opened 2019-07-01
Fix ckwriter failing test - update to use fixed kernels aside from the sclks

krlberry:
## Description
The ckwriter lroc test began failing after the ISIS kernel management scripts were updated to clean up all but the last 10 sclk kernels for lroc. ckwriter looks for and uses the original kernels it was spiceinited with, and the sclk which had been used for the original spiceinit was cleaned up, causing the test to fail.

The test was updated to spiceinit the input cubes with all the kernels fixed to their original values except for the sclk, so that ISIS will pick up the newest sclk file (which will exist.)

It was necessary to change the structure of the file as well, since the arguments to spiceinit needed to be different for each input cube, the loop in the prior version could no longer be used.

## Related Issue
Internal testing started to fail. Does this need an issue?

## Motivation and Context
Updates an internal test to still work with other kernel management maintenance updates. 

## How Has This Been Tested?
The test passes after this update.

## Types of changes

## Checklist:

## Licensing


scsides (approved):
Looks like a clean test

PR #3379, opened 2019-08-07
Changed FileName to use US English for date conversions

### LINKED ISSUE #3340: spiceinit for Cassini can't find newest PCK
**ISIS version(s) affected: All at time of issue creation 3.7.1

**Description**  
I’ve been running into some trouble with the Spiceinit module in ISIS 3.6.2. I’m trying to process  data from the Cassini mission. I’m getting the following error:
“No existing files found with a date version matching [cpck{ddMMMyyyy}.tpc] in[.../isis3/data/cassini/kernels/pck]”
when running: 
`spiceinit from=v1637464787_1_ir
`

See discussion at: https://astrodiscuss.usgs.gov/t/spiceinit-error/171

**How to reproduce**  
With ISIS 3.7.1 or earlier and the bash shell
```
export LC_TIME=nl_NL.UTF-8
spiceinit from=v1637464787_1_ir FAILS

export LC_TIME=“en_US.UTF-8”
spiceinit from=v1637464787_1_ir SUCCEEDS
```

**Possible Solution**  
Find out why the setting of LC_ALL in the Application class constructor is not being honored.

**Additional context**  
spiceinit / kerneldb / filename / highestversion is searching for a kernel file that uses a date as the version number in: .../data/cassini/kernels/pck/cpckDDMMMYYYY.tpc
That search comes back empty when the LC_TIME locale is set to nl_NL.

jessemapel:
If the user is set to a non-English language, then finding the latest version of files that use short names of months fails. Now FileName always uses US English for string to date conversions.

## Description

For example, if a file uses something like 15Dec2017 and the users language is set to Spanish, then FileName cannot properly convert Dec to the 12th month because it is looking for the Spanish word for December. There is existing code in ISIS that sets these at the system level, but the QT locale utilities are not properly updating on linux (oddly this doesn't seem to be a problem on Mac). So, I added code to set the default in QT to US English and changed calls to use the default instead of the system locale.

## Related Issue

Fixes #3340 

## Motivation and Context

Users with non-English language settings had odd failures during spiceinit. See https://astrodiscuss.usgs.gov/t/spiceinit-error/171 for more info.

## How Has This Been Tested?

Tested on Mac and Linux with Cassini VIMS data because it uses kernels ranked by the date they were created.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/FileName/FileName.cpp
+++ b/isis/src/base/objs/FileName/FileName.cpp
@@ -28,6 +28,7 @@
 #include <QDate>
 #include <QDebug>
 #include <QDir>
+#include <QLocale>
 #include <QPair>
 #include <QString>
 #include <QTemporaryFile>
@@ -622,8 +623,7 @@ namespace Isis {
                     foundFile.mid(truncateRange.second + 1);
 
       }
-
-      QDate fileDate = QDate::fromString(foundFile, fileQDatePattern);
+      QDate fileDate = QLocale().toDate(foundFile, fileQDatePattern);
 
       if (fileDate.isValid()) {
         // No missions before Sputnik 1, so we must be in the new millenium
--- a/isis/src/base/objs/FileName/FileName.cpp
+++ b/isis/src/base/objs/FileName/FileName.cpp
@@ -623,7 +623,7 @@ namespace Isis {
                     foundFile.mid(truncateRange.second + 1);
 
       }
-      QDate fileDate = QLocale().toDate(foundFile, fileQDatePattern);
+      QDate fileDate = QLocale(QLocale::English, QLocale::UnitedStates).toDate(foundFile, fileQDatePattern);
 
       if (fileDate.isValid()) {
         // No missions before Sputnik 1, so we must be in the new millenium
--- a/isis/tests/FileListTests.cpp
+++ b/isis/tests/FileListTests.cpp
@@ -11,12 +11,12 @@ TEST(FileList, NonExistantFileConstructor)
   }
   catch(Isis::IException &e)
   {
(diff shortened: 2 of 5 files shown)


isis/src/base/objs/FileName/FileName.cpp
```diff
@@ -622,8 +623,7 @@ namespace Isis {
                     foundFile.mid(truncateRange.second + 1);
 
       }
-
-      QDate fileDate = QDate::fromString(foundFile, fileQDatePattern);
+      QDate fileDate = QLocale().toDate(foundFile, fileQDatePattern);
```
scsides: This member was designed to work with US English locale ONLY. Would it be better to force this call to use US English, and not rely on the Application to set it correctly?

scsides (changes requested):
This should work. Comments on my request appreciated

scsides (approved):

isis/tests/FileListTests.cpp
```diff
@@ -11,12 +11,12 @@ TEST(FileList, NonExistantFileConstructor)
...
   catch(Isis::IException &e)
   {
-    EXPECT_TRUE(e.toString().toLatin1().contains("Unable to open [FakeFile]."))
+    EXPECT_TRUE(e.toString().toLatin1().contains("Unable to open [FakeFile]"))
       << e.toString().toStdString();
   }
   catch(...)
   {
-    FAIL() << "Expected an IException\"Unable to open [FakeFile].\"";
+    FAIL() << "Expected an IException\"Unable to open [FakeFile]\"";
```
jessemapel, line 19: Technically this is outside of the scope of the PR, but it's such a simple change I think it will be okay.

Issue #3192, opened 2019-03-28
cam2map is very slow on large images


scsides:
**Description**  
The cam2map program runs for 20 to 30 minutes on full length LROC NAC images. It would be helpful if the program could be sped up.

**Example**  
All long LROC NAC images (52K lines) take from 20-30 minutes to run through cam2map

PR #3381, opened 2019-08-09
Set the buffer size to the area being worked instead of 1x1

scsides:
## Description
The cube buffer for the "ForwardPatch" algorithm in 
ProcessRubberSheet (used in cam2map) was set to one line and one sample. The buffer was increased to the size of the area being worked on (i.e., zero to a few pixels larger than the output patch size, depending on the magnitude of the warp). This allows cube data to be buffered at the patch level instead of one pixel at time.

## Related Issue

#3192
## Motivation and Context

For an output patch size of 5x5 this saves 24 output buffer transfers per patch over the 1x1

## How Has This Been Tested?

All ISIS unit and regression tests have been run and passed.

Only applications that use the forward patch algorithm in ProcessRubberSheet are affected. There are small DN differences resulting from the output DN being written to the cube from the processing of a different patch than when the buffer was 1x1.

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/ProcessRubberSheet/ProcessRubberSheet.cpp
```diff
@@ -1075,22 +1078,43 @@ namespace Isis {
...
+    if (foundNull) {
+     Brick readBrick(*OutputCubes[0], osampMax-osampMin+1, olineMax-olineMin+1, 1);
+     readBrick.SetBasePosition(osampMin, olineMin, iportal.Band());
+     OutputCubes[0]->read(readBrick);
+     for (brickIndex = 0; brickIndex<(osampMax-osampMin+1)*(olineMax-olineMin+1); brickIndex++) {
+       if (readBrick[brickIndex] != Null) {
+         oBrick[brickIndex] = readBrick[brickIndex];
+       }
+     }
+    }
```
jessemapel, line 1111: Instead of having to check if for foundNull and then re-process, you could always start out by reading into oBrick and then if the interpolator returns a null, not assigning over that pixel in oBrick. This way, you don't overwrite anything with Nulls. It'll be a little bit more costly to read every patch, but I'm not sure if that is significant. May be worth pushing through a profielr.

jessemapel (commented):
Changes look good. The foundNull logic seems overly complex for just ignoring any Nulls.

isis/src/base/objs/ProcessRubberSheet/ProcessRubberSheet.cpp
scsides: I'm actually in the process of doing just that. It turns out the output patches can have a lot of overlap. Without this current logic the edges of the projected area are a lot rougher. This also slows it down more than I like, but the output is nearly identical to before our changes.

SgStapleton (approved):
Looks good!

PR #3395, opened 2019-08-13
Updates to how bundleout.txt is being written 

SgStapleton:
The bundleout.txt file from a run of jigsaw is updated.

## Description

This update includes new information (such as units and control network information) and a reformatting of older information. The code was also reworked for maintainability. 

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/3227

## Motivation and Context

This results in easier to read, and more informative log files.

## How Has This Been Tested?

A full run of the entire testing suite was conducted.

## Types of changes

## Checklist:

## Licensing


jlaura:
:tada: :+1:

isis/src/control/objs/BundleUtilities/BundleObservation.h
```diff
@@ -120,20 +133,23 @@ namespace Isis {
...
       const BundleObservationSolveSettingsQsp solveSettings();
 
-//    QStringList serialNumbers();
-
       bool applyParameterCorrections(LinearAlgebra::Vector corrections);
       bool initializeExteriorOrientation();
       void initializeBodyRotation();
       void updateBodyRotation();
 
-      QString formatBundleOutputString(bool errorPropagation, bool imageCSV=false);
```
jessemapel, line 136: This is API breaking :(((((((((

jessemapel (commented):
These changes all look solid, but this is API breaking. @jlaura do you have thoughts on this?

I'm tempted to say that we are not taking library API into account for major version numbers. The other option would be just doing ISIS 4 next time?

SgStapleton:
It would be, yes. I believe I actually know a way of re-working the logic that would allow for this to be non-breaking possibly, but was not sure how much more time was really wanting to be invested in this? @jlaura, any thoughts?

isis/src/control/objs/BundleUtilities/BundleObservation.h
acpaquette: Is it worth it to try to maintain the old API to get this feature into the release?

jessemapel: that may be the solution for right now. Maybe just flag it as deprecated?

SgStapleton: What would be the best way to flag deprecation? And is the idea here to re-implement the monstrous code that was removed...?

jessemapel: By re-implement, I just mean copy and paste the old code along side the new code. Flagging as deprecated is as simple as warning when the method gets called. c++14 has special things for deprecated functions, but we only use c++11 right now.

acpaquette:
There are a few other things that can get cleaned up with this PR but I'm not sure how much time we want to spend on that if we want to get this into ISIS.

jlaura:
@lwellerastro Can you weigh in here on the timeliness of this? The team has a standup at 11 today in their office - perhaps best to discuss this issue in the standup and determine a path. If this is going to go in, we should strive to get it merged today (so someone else will need to pick this up and make the changes).

acpaquette:
I can comment on what I think should be changed before hand so other people can get a better idea of some of the issues with the current code.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -869,266 +857,356 @@ namespace Isis {
...
+   *
+   * @param fpOut The open std::ofstream object which is passed in from
+   * BundleSolutionInfo::outputText()
+   * @param errorPropagation Boolean indicating whether or not to attach more information
+   *     (corrections, sigmas, adjusted sigmas...) to the output.
+   */
+  void BundleObservation::bundleOutputString(std::ostream &fpOut, bool errorPropagation) {
 
-    }// end if(!imageCSV)
+    char buf[4096];
```
acpaquette, line 898: Both of these outputs formats logic have consistent pieces between the two. Why the String output uses a character buffer and the CSV output uses a QString is odd and both should probably use one or the other.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -869,266 +857,356 @@ namespace Isis {
...
-          adjustedSigma = QString::number(m_adjustedSigmas[i], 'f', 8);
-          sigma = ( IsSpecial(m_aprioriSigmas[i]) ? "FREE" : toString(m_aprioriSigmas[i], 8) );
-        }
-        // Provide default values for position if not solving position
-        else {
+
+    // We need to use an offset of -3 (1 coef; X,Y,Z) if we used the default center coordinate
+    // (i.e. we did not solve for position), as m_corrections and m_*sigmas are populated
+    // according to which parameters are solved
+    int offset = 0;
```
acpaquette, line 1047: This offsetting is somewhat unnecessary and I think there is a cleaner way to address this. Mainly splitting the resulting Position and Pointing parameters into two different arrays in the fetchBundleOutputData call instead of a single massive array of parameters.

acpaquette (commented):
These are the two main areas I would like to address to get this PR cleaned up. There are probably some other areas that can be touched on but I don't think it would be worth it without redoing a lot of the work that has already gone into this.

lwellerastro:
This request is many, many years in the making - not having unit information in the output is unbelievable and apparently very difficult/painful to do from a coding standpoint.  That being said, having to wait longer for that particular information is doable.  However, and more importantly, Ken had Tyler add the total RMS of the image measures to the Image Measures summary in the bundleout.txt, a number Ken would calculate elsewhere and provide project chiefs with as an indicationof the overall measure of how the bundle did (this is what would usually end up in publications).  I would really like to have that number.  Also, there were some relevant things at the top of the bundleout.txt being left out that were being added in, like the input image list and one or two other things.  Would be nice.

I don't understand why this is API breaking.  The only thing that was being updated was the bundleout.txt - a file absolutely no one is trying to parse.  There weren't changes to any of the parameters or csv files.  What's being broken other than truth files?

@jlaura, I worried if I say "eh, let's just wrap this up next sprint (November-ish?), it will not get done because something "more important' will punt it right off the table and users still won't have information that should have been included day one.

jlaura:
@lwellerastro Checkout @jessemapel's comment above; line 136 in the bundleobservation.h file. This is changing how someone calls our software and we have agreed not to make API breaking changes without ticking the major version number (e.g., 4.0.0). Therefore, as written, we can not get this code put in this sprint. We have a 4.0.0 release planned Jan2020 when we will release versions 3 and 4 concurrently.

lwellerastro:
> @lwellerastro Checkout @jessemapel's comment above; line 136 in the bundleobservation.h file. This is changing how someone calls our software and we have agreed not to make API breaking changes without ticking the major version number (e.g., 4.0.0). Therefore, as written, we can not get this code put in this sprint. We have a 4.0.0 release planned Jan2020 when we will release versions 3 and 4 concurrently.

@jlaura I see it but I don't understand it.  There were no changes to user input, only bundleout.  Sorry you have to spell it out for me.

Is that line saying, if errorprop=true, no image.csv file will be created??  Why???  Insert lots of curse words here.  Lots.  Why would that be the case?  I don't understand and you probably don't have an answer either.

Update: Talked with Adam and he shed light on the API break for me.  Hope his idea on how to deal with things works.

jlaura:
No need to apologize. Totally on me for not describing well. We have a link here that talks a bit more about breaking vs. non-breaking: https://github.com/USGS-Astrogeology/ISIS3/wiki/RFC2:-Release-Process#terms

Basically, if the method was originally `foo(a, b, c)` and we want to change it to `foo(b,a)` we have broken compatibility for anyone that was using our code. In the case above, it looks like the line is being removed entirely. This then breaks the API because we do not know if someone external to us is using that method. If/when we go to remove that we need to do a proper deprecation and to alert anyone that is using it that it will be going away.

jessemapel:
@jlaura @lwellerastro During today's stand up we decided to add the old function back and either flag it as deprecated or make it call the new functions. This way these changes can still go ourt in 3.9. @acpaquette and I will be working on this today.

lwellerastro:
> This then breaks the API because we do not know if someone external to us is using that method.

Thanks @acpaquette and @jlaura, my user only mentality had me only thinking about user input/output when considering breaking the API, not the class or object changing that might break someone trying to use it outside of the calling application.

jlaura:
@jessemapel @acpaquette That sounds like an awesome plan!

jessemapel:
Continued yet again in #3404

Issue #3171, opened 2019-03-22
Updates to tgocassisrdrgen exported PDS4 label are needed.


krlberry:
**ISIS version(s) affected**: 3.6.2+

**Description**  
Problems:

* The PDS4 label exported by `tgocassisrdrgen` identifies the image as a ExoMars TGO FREND file, not an ExoMars TGO CaSSIS file in the `logical_identifier`
* `data_mosaic` is not a valid level of data processing.

From the label:

``` 
<logical_identifier>urn:esa:psa:em16_tgo_frd:data_mosaic:MY34_004224_188_1_PAN</logical_identifier>

```

This is an example of a correctly-formatted `logical_identifier`:

```
urn:esa:psa:em16_tgo_cas:data_raw:cas_raw_sc_20161124t160000-20161124t220000-8-1-1-20161124t185305854-blu-4-0
```
"frd" needs to be changed to "cas" and `data_mosaic` is not valid, so this needs to be changed to something that is valid.

**How to reproduce**  
Run `tgocassisrdrgen` on any CaSSIS cube and look at the exported label.

**Possible Solution**  
Update the appropriate translation file or code to change the output string.

**Additional context**  
This is further work on the result of the last CaSSIS sprint: #2635 

Tanya Lim from PSA pointed out this issue at LPSC.

I will add any additional needed changes that are identified to this ticket.

jessemapel:
@kberryUSGS Is this fixed?

krlberry:
Fixed.

PR #3402, opened 2019-08-15
Update to CaSSIS Exported PDS4 images and Mosaicking program

krlberry:
Updates `tgocassisrdrgen`, `tgocassismos` and supporting classes to produce output PDS4 images and labels up to the most recent spec. 

## Description
Changes to tgo-related programs:

* tgocassismos: updated to include the Archive group, as it contains many items required for the PSA's schema. (Other updates are not functional and just for clarity.) 

* tgocassisrdrgen: updated to use new schema, include a pixel description, fix some output label values, and remove units from the focal-length in the output "by-hand"

Changes to supporting classes:

* iTime: updated to allow the precision of the output UTC if desired. This is backwards compatible.

* Cube: added a method to return the lat/lon range of the cube. Added because the min/max lat/lon in the Mapping group was not accurate enough.

* ProcessExportPDS4: added a pixel description, fixed output of "special pixel values", add ability to specify a schema without a schematron, fix output of a couple of mapping values that don't output in floating-point, which is invalid for PDS4, update to use min/max lat/lon from cube rather than mapping group, fix formatting of times.

## Related Issue
#3171

## Motivation and Context
The CaSSIS team needs to export images in PDS4 format that use the PSA's schema. 

## How Has This Been Tested?
The existing app tests and unit tests have been updated and the resulting labels have been reviewed.

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
```diff
@@ -1126,32 +1165,73 @@ namespace Isis {
...
+    QDomElement projectionElement = getElement(xmlPath, baseElement);
+    QDomElement tempElement = projectionElement.firstChildElement();
+    QDomElement nameElement = tempElement.nextSiblingElement();
+
+    QDomElement longitudeElement = nameElement.firstChildElement("cart:longitude_of_central_meridian"); 
+    QDomElement originElement = nameElement.firstChildElement("cart:latitude_of_projection_origin");
+
+    double longitudeElementValue = longitudeElement.text().toDouble(); 
+    double originElementValue = originElement.text().toDouble(); 
+    QString toset1 = QString::number(longitudeElementValue, 'g');
```
krlberry: If anyone has any better idea for how to do this, let me know. Forcing it to be a float somehow didn't force it to output with the required ".0" so I just added it this way.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
   }
 
 
+/**
+ * Returns the latitude and longitude range for the Cube. More accurate than the minimum and 
+ * maximum latitude and longitude from the mapping group. 
+ * 
+ * @param minLatitude minimum latitude of the cube
```
jessemapel: Why is this worded differently than the other parameters?

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+ * @param maxLatitude maximum latitude present in the cube
+ * @param minLongitude minimum longitude present in the cube
+ * @param maxLongitude maximum longitude present in the cube
```
jessemapel: These should be flagged as output parameters. You can use `@param[out]` for this.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+ */
+  void Cube::latLonRange(double &minLatitude, double &maxLatitude, double &minLongitude, double &
+                         maxLongitude) {
+    Camera *cam;
+    TProjection *proj;
+
+    bool isGood = false;
+    bool noCamera;
```
jessemapel: This would be better described as `useProj` or something to indicate what it does.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+
+    if (hasGroup("Instrument")) {
+      noCamera = false;
+    }
+    else {
```
jessemapel: You can remove this else block if you default the bool flag to true.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+        QString msg = "Unable to create camera when calculating a lat/lon range.";
+        throw IException(e, IException::User, msg, _FILEINFO_);
+      }
+    }
+
+    // Iterate over all samp/line combos in cube
+    minLatitude = 99999;
+    minLongitude = 99999;
+    maxLatitude = -99999;
+    maxLongitude = -99999;
```
jessemapel, line 2182: If nothing in the image successfully projects, then these will get returned. I'm not sure how to deal with this case or even if we should at all.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+
+    for (int sample = 0; sample < sampleCount(); sample++) {
+    // Checks to see if the point is in outer space
+      for (int line = 0; line < lineCount(); line++) {
```
jessemapel: This is potentially a very time consuming process. Something like a quad tree approach would probably work significantly faster but may be a little less accurate.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2133,6 +2134,96 @@ namespace Isis {
...
+      }
+    }
+
+    // Iterate over all samp/line combos in cube
+    minLatitude = 99999;
+    minLongitude = 99999;
+    maxLatitude = -99999;
+    maxLongitude = -99999;
+
+    for (int sample = 0; sample < sampleCount(); sample++) {
```
jessemapel: What about fractional pixels? Technically the image goes from 0.5 samples to sampleCount() + 0.5. Same for lines.

isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
```diff
@@ -882,36 +898,69 @@ namespace Isis {
         elementArrayElement.appendChild(offsetElement);
       }
 
-      // Add the Special_Constants class to define ISIS special pixel values: 
+      // Add the Special_Constants class to define ISIS special pixel values if pixel type is
+      // Real 
+      if (p_pixelType == Isis::Real) {
```
jessemapel: We know we are going to have to support PDS4 for a lot of different missions. This can be easily expanded to work with any ISIS pixel type by using a switch statement.

isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
jessemapel: Can you use the precision argument from [Qstring::number](https://doc.qt.io/qt-5/qstring.html#number-6)?

isis/src/tgo/apps/tgocassismos/main.cpp
```diff
@@ -184,30 +185,35 @@ void IsisMain() {
...
+   // After selecting the last StartTime, calculate the StopTime
+   iTime lastStartTimeValue = iTime(lastStartTime);
+   iTime stopTimeValue = lastStartTimeValue + exposureDuration.toDouble(); 
+   stopTime = stopTimeValue.UTC(3);
 
    // Get the blob of original labels from first image in list
    // Pvl *org = cubeList[0]->label();
+   // Get the archiveGroup from the first cube in the list
+
+    PvlGroup archiveGroup = cubeList[0]->label()->findGroup("Archive", Pvl::Traverse);
```
jessemapel: You can use Cube::group() here

isis/src/tgo/apps/tgocassisrdrgen/main.cpp
```diff
@@ -96,49 +96,62 @@ void IsisMain() {
     process.setVersionId( ui.GetString("VERSIONID") );
   }
 
+  process.setPixelDescription("Pixel values are in units of I/F (intensity/flux). I/F is defined as the ratio of the observed radiance and the radiance of a 100% lambertian reflector with the sun and camera orthogonal to the observing surface.");
```
jessemapel: Break this long string at <100 chars

isis/src/tgo/apps/tgocassisrdrgen/main.cpp
```diff
@@ -96,49 +96,62 @@ void IsisMain() {
...
+  QDomDocument &pdsLabelNext = process.GetLabel();
+  QDomElement observationNode = pdsLabelNext.documentElement().firstChildElement("Observation_Area");
+  QStringList xmlPath;
+  xmlPath << "Observation_Area"
+          << "Mission_Area"
+          << "cas:CASSIS_Data"
+          << "cas:telescope_information"
+          << "cas:focal_length";
+  QDomElement focalLengthNode = process.getElement(xmlPath, observationNode);
+  focalLengthNode.removeAttribute("unit");
```
jessemapel: Does this need a check if it has the units attribute?

jessemapel (commented):
Sensible bunch of changes. The latLonRange method has a lot of nesting going on that makes it hard to read, but I can't think of a good way to make it simpler without using things like ternary operators or function pointers.

jessemapel:
> Can one of the admins verify this patch?

isis/src/base/objs/Cube/Cube.cpp
krlberry: Not sure why, really. Fixing this.

isis/src/base/objs/Cube/Cube.cpp
krlberry: Good point.

isis/src/base/objs/Cube/Cube.cpp
krlberry: Also a good point.

isis/src/base/objs/Cube/Cube.cpp
krlberry: Good point.

krlberry: I updated it to throw an exception if any of these are still set to their original 9999 values.

krlberry:
@jessemapel I updated everything aside from making the latitude/longitude search quicker. Not that this is the best justification, but similar searches happen elsewhere in ISIS. Can the switch to something quicker but potentially less accurate come in as a future enhancement? This was added in the first place as the existing values for Min/Max lat/lon were not precise enough.

isis/src/base/objs/Cube/Cube.cpp
krlberry: Updated this to throw an exception if any of these remain non-updated.

isis/src/base/objs/ProcessExportPds4/ProcessExportPds4.cpp
krlberry: Amusingly, I went down this odd rabbit hole in the first place because 'g' removes trailing 0's and I didn't realize 'f' didn't.

jessemapel:
@kberryUSGS I think it's totally fine to leave this as not super efficient. It's not a bottle neck in any pipeline.

jessemapel (approved):

jessemapel:
@kberryUSGS This can be merged whenever you are ready to check in the updated test data.

krlberry:
@jessemapel Thanks :+1:

isis/src/base/objs/Cube/Cube.cpp
jlaura: I don't think a quad tree helps you here, because you would have to get bounds to do the construction. A sampling approach might though - take the min/max pixels and then randomly sample n additional pixels to get an approximate lat/lon range. In most cases this will be the correct values. To account for weird geometries, you could sample without replacement in chunks of n, if the min/max values do not change after chunk m, break.

PR #3470, opened 2019-10-11
Added Working BundleMeasure gtest File

egaultUSGS:
## Description

We are currently porting the old ISIS3 tests to the gtest framework, and BundleMeasure was the first Class I was assigned to port over. Since testing most of the functions in BundleMeasure would essentially be testing the Classes it is composed of, only four tests were implemented.  
These are:

- A constructor test
- A copy constructor test
- An assignment operator test
- An isRejected function test.

Note: A test fixture is used to create the BundleMeasure Object for testing, since the only available constructor has many dependencies.

## Related Issue

Can't currently find the issue that regards porting the old unit tests to gtest.

## Motivation and Context

We are porting ISIS3's testing suite to gtest, and BundleUtilities was a good place for me to start.

## How Has This Been Tested?

The test file compiles properly and produces the expected (passing) outputs when the testing suite is run.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/BundleMeasureTests.cpp
+++ b/isis/tests/BundleMeasureTests.cpp
@@ -1,3 +1,5 @@
+#include <memory>
+
 #include "BundleControlPoint.h"
 #include "BundleMeasure.h"
 #include "BundleSettings.h"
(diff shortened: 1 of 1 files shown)


isis/tests/BundleMeasureTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    // Since BundleMeasure and BundleControlPoint do not have
+    // default constructors, and C++ does not allow for object declarations
+    // pointers to each respective object are used.
+    BundleMeasure *testBundleMeasurePtr;
+    BundleControlPoint *testBundleControlPointPtr;
+
+
+    void SetUp() override
+    {
```
Kelvinrr: Yo, curly bracket on it's own line? That's not very cash money of you.

isis/tests/BundleMeasureTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+//
+// TEST_F( BundleMeasure_CreateObj, IsRejected )
+// {
+//   EXPECT_EQ( false, testBundleMeasurePtr->isRejected() );
+//   testBundleMeasurePtr->setRejected( true );
+//   EXPECT_EQ( true, testBundleMeasurePtr->isRejected() );
+//   testBundleMeasurePtr->setRejected( false );
+//   EXPECT_EQ( false, testBundleMeasurePtr->isRejected() );
+//
+// }
```
Kelvinrr: Any reason most of the tests are commented out?

isis/tests/BundleMeasureTests.cpp
egaultUSGS: I have a class that requires brackets to be placed in this manner, so I guess it's a habit from that. I'll fix it to be more in line with how the other gtests do it.

isis/tests/BundleMeasureTests.cpp
egaultUSGS: Mostly to focus on the issue we think is causing a lot of the problems with the rest of the tests. For instance, the first CopyConstructor test doesn't cause errors, but fails on the final assertion, and CopyConstructor2 causes a segfault to occur, despite doing essentially the same thing as the first CopyConstructor test.  
Also, the first two lines of code in the AssignmentOperator test cause a segfault, but running the test with only one line or the other will not.

isis/tests/BundleMeasureTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+      BundleSettingsQsp bundleSettingsQsp( new BundleSettings );
+
+      // Create a BundleControlPoint object to feed its pointer into
+      // the BundleMeasure constructor.
+      BundleControlPoint bundleControlPoint( bundleSettingsQsp, &controlPoint );
+
+      // Finally, the SetUp function creates the BundleMeasure object
+      // from all of the objects made previously.
+      // std::cout << &controlMeasure << '\n';
+      BundleMeasure bundleMeasure( &controlMeasure, &bundleControlPoint );
```
jessemapel: Both of these objects will be destroyed when they go out of scope once the setup function returns. I'd expect this to cause your first test to seg fault.

You should heap allocate them instead of stack allocate them. I'd also recommend using a smart pointer such as a unique_ptr.

isis/tests/BundleMeasureTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+      testBundleMeasurePtr = unique_ptr<BundleMeasure>( new BundleMeasure( controlMeasure.get(),
+                                                                           testBundleControlPointPtr.get() ) );
+
+    }
+};
+
+TEST_F( BundleMeasure_CreateObj, Constructor ) {
+  EXPECT_FALSE( testBundleMeasurePtr->isRejected() );
+
+  EXPECT_EQ( NULL, testBundleMeasurePtr->camera() );
```
jessemapel, line 55: You should use `nullptr` instead of `NULL`

jessemapel (commented):
Test looks good so far. Still needs to test all the methods.

isis/tests/BundleMeasureTests.cpp
egaultUSGS: When I was talking to Adam and Stuart about the best way to test these methods, they felt that testing any methods other than the ones I have PR'd would be essentially the same as testing ControlMeasure and BundleControlPoint over again. 

I can certainly still implement tests for these methods, but I'm not exactly sure which camera and image I should be using for testing purposes.

I will be sure to change the NULL to nullptr. Would it only be necessary to change it for the camera method, or are there other places that should also use it?

jessemapel: Looking at the code that seems reasonable

jessemapel (approved):

Issue #3476, opened 2019-10-21
Backing out the ALE integration merge


jessemapel:
#3464 merged the new code that integrates [ALE](https://github.com/USGS-Astrogeology/ale) into spiceinit.

When this was merged, Jenkins was down and since has come back up. We have several test and build failures on Jenkins right now and it is not clear which are caused by the ALE integration and which are caused by changes in things like the docker image Jenkins slaves use.

I've spent about a day working on this and I think there's too much that needs to be addressed for this to be handled right now. So, I'm proposing we back out ALE until the ASC gets back on the project and these problems can be addressed.

krlberry:
:+1: on backing out the changes to make it possible to see what is failing without the ALE changes. Do we know when we'll be able to get back on the project?

jessemapel:
@kberryUSGS We have a software scheduling meeting tomorrow. Either way, I'd like to have at least 1 person work on this starting either next week or after the products work.

jlaura:
:+1: Backing out and potentially addressing during the products sprint sounds like a fine idea here. Can we accomplish backing the PR out this week so that we start the products sprint with (potentially) passing builds?

jessemapel:
I'm going to leave this up for the day, if there are no objection by tomorrow morning I'll back ALE out.

jessemapel:
changes have been backed out and Jenkins builds are passing again.

PR #3477, opened 2019-10-22
Revert "Ale integration into dev "

jessemapel:
Reverts USGS-Astrogeology/ISIS3#3464

Backing out ALE integration changes as dev is currently failing builds/tests. These changes will go back in the future with updated tests.

See #3476 for more context


krlberry (approved):
:+1: Looks like it successfully reverts the ALE PR.

Kelvinrr (approved):
The builds are failing because of a Boost lib issue, seems like it isn't explicitly installed in either case? Not sure why it's an issue on Jenkins but not when creating the env locally. Whenever @tthatcher95 can get multi branch builds going it's probably worth having the Ale integration branch running so we can make sure things pass there before merging.

scsides (approved):
That looks like everything.

PR #3488, opened 2019-11-06
Fixes Qmos grid to properly consider PositiveWest and PositiveEast

Kelvinrr:
## Description

PositiveWest in qmos didn't display a line at 360 degrees, this was because the original grid code didn't incorporate positive west at all in its logic. 

This change moves the responsibility of converting between positive west/east to the grid drawing function instead of the projection code. This makes sure the grid is drawn correctly given arbitrary lat/lon ranges. 

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/3417

## How Has This Been Tested?

Manually ran qmos with both PositiveWest and PositiveEast projections and changed parameters to ensure grid updates properly. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
+++ b/isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
@@ -43,8 +43,12 @@ namespace Isis {
         Latitude maxLat;
         Latitude startLat;
         Latitude endLat;
+      
 
       if (tproj->IsPositiveWest()) {
+        // GridGraphicsItem is written assuming positive east 
+        // for all angles. On positive West, lons come in swapped so we 
+        // need to account for this.
         Longitude temp = lonMin;
         lonMin = lonMax;
         lonMax = temp;
@@ -195,6 +199,9 @@ namespace Isis {
               double y = 0;
               
               bool valid;
+
+              // Set ground according to lon direction to get correct X,Y values 
+              // when drawing lines. 
               if (tproj->IsPositiveWest()) {
                 valid = tproj->SetGround(lat.degrees(), lon.positiveWest(Angle::Degrees));
               }
@@ -289,7 +296,9 @@ namespace Isis {
             while (!atMaxLat) {
               double x = 0;
               double y = 0;
-                
+              
+              // Set ground according to lon direction to get correct X,Y values 
+              // when drawing lines.  
               bool valid;
               if (tproj->IsPositiveWest()) {
                 valid = tproj->SetGround(lat.degrees(), lon.positiveWest(Angle::Degrees));
--- a/isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
+++ b/isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
@@ -44,7 +44,7 @@ namespace Isis {
         Latitude startLat;
         Latitude endLat;
 
-      if (mappingGroup["LongitudeDirection"][0] == "PositiveWest") {
+      if (tproj->IsPositiveWest()) {
         Longitude temp = lonMin;
         lonMin = lonMax;
         lonMax = temp;
@@ -195,7 +195,7 @@ namespace Isis {
               double y = 0;
               
               bool valid;
-              if (mappingGroup["LongitudeDirection"][0] == "PositiveWest") {
+              if (tproj->IsPositiveWest()) {
                 valid = tproj->SetGround(lat.degrees(), lon.positiveWest(Angle::Degrees));
               }
               else {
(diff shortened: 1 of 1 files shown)


isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
```diff
@@ -193,7 +193,14 @@ namespace Isis {
 
               double x = 0;
               double y = 0;
-              bool valid = tproj->SetUniversalGround(lat.degrees(), lon.degrees());
+              
+              bool valid;
+              if (mappingGroup["LongitudeDirection"][0] == "PositiveWest") {
```
jessemapel: I would use `tproj->IsPositiveWest()` here instead of the string comparison

isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
```diff
@@ -283,16 +289,20 @@ namespace Isis {
             while (!atMaxLat) {
               double x = 0;
               double y = 0;
-              bool valid = tproj->SetUniversalGround(lat.degrees(), lon.degrees());
+                
+              bool valid;
+              if (mappingGroup["LongitudeDirection"][0] == "PositiveWest") {
```
jessemapel: Same as other check

jessemapel (commented):
Logic looks good, just a small tweak for clarity and future-proofing

isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
```diff
@@ -193,7 +193,14 @@ namespace Isis {
 
               double x = 0;
               double y = 0;
-              bool valid = tproj->SetUniversalGround(lat.degrees(), lon.degrees());
+              
+              bool valid;
+              if (mappingGroup["LongitudeDirection"][0] == "PositiveWest") {
+                valid = tproj->SetGround(lat.degrees(), lon.positiveWest(Angle::Degrees));
```
Kelvinrr, line 206: Changed `setUniversalGround` to `setGround`. `setUniversalGround` tries to naively account for a positive west angle if the projection object is set to `positiveWest`, although the grid code expects exclusively positive east angles. This basically keeps everything in positive east angles until the set ground call and lets the `Longitude` object handle angle conversion.

jessemapel: Can you put a comment in the code that briefly says this?

Kelvinrr: This is more of a comment on the previous state and how inexplicit it was than the current. I can put a comment on the swap mentioning the forced swap and maybe a minor comment of why we are explicitly converting but the latter seems like a low-value comment.

jessemapel: I think it's valuable to document that the grid code expects exclusively positive east angles. Maybe this should just be an issue?

Kelvinrr: I think it's worth making an issue as to why lons are incorrect when passed into the constructor on positive west so we can avoid the awkward min/max lon swap.  Preference for positive east lons seems to be common in ISIS?

isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
```diff
@@ -283,16 +289,20 @@ namespace Isis {
...
+                valid = tproj->SetGround(lat.degrees(), lon.positiveEast(Angle::Degrees));
+              }
 
               if (valid) {
                 x = tproj->XCoord();
                 y = -1 * tproj->YCoord();
 
                 if(havePrevious) {
-                  x = proj->XCoord();
-                  y = -1 * proj->YCoord();
```
acpaquette, line 294: Why'd this get removed? Just kinda curious what the difference is between the tproj and the proj object

acpaquette (commented):
Looks good overall :+1:

acpaquette (approved):
Pushing the approve button 😬

isis/src/qisis/objs/MosaicSceneWidget/GridGraphicsItem.cpp
krlberry: I second this question

jessemapel: It's duplication of the step above without any changes. It'll overwrite itself with the exact same value.

jessemapel (approved):

krlberry (approved):
:+1: Also looks good to me. Thanks for adding that comment.

PR #3490, opened 2019-11-06
Updates voy2isis to add an optional parameter to specify the instrument name

### LINKED ISSUE #3426: Voy2isis ingestion error leading to spiceinit failure
**ISIS version(s) affected**: 
ISIS3.8.1. Error apparently did NOT occur with an older versions (maybe 3.4.13 see details below)

**Description**  
On certain PDS images (*.imq see below), when running Voy2isis I receive the following warning:
 
Group = Warning
  Message = "The INSTRUMENT_NAME for [./c2061155.img] is empty.The
             InstrumentId in the output cube will instead be set to [Unknown]
             and the labels will not translate."
End_Group
**USER ERROR** Instrument ID [Unknown] does not match Narrow or Wide angle camera. The cube was created, but the labels were not translated.

The cubes were indeed created, but when I subsequently run spiceinit I get the following error:

**ERROR** An time string that was either not a time(N/A) or incorrectly formatted was passed into a naif routine. The short explanation provided by NAIF is [SPICE(UNPARSEDTIME)]. The Naif error is [The input string contains an unrecognizable substring beginning at the character marked by <Z>: "1979-07-08T08:26:23<Z>"].

We (really @lwellerastro, who discovered the problem) have a work around because she found "clean" .cub files in one of Tammy's old directories. These seem to have been generated with ISIS3.4.13 (according to cathist). However, external users won't be able to use these images, which will be included in the Europa project. 

**How to reproduce**  
The 3 images for which this problem occurs (there are likely more) can be found in /work/users/mbland/Europa-Voyager/failing. They are c2061155.imq, c2061159.imq, c2061203.imq. Using ISIS3.8.1 on Astrovm4 run

voy2isis from=c2061155.imq to=c2061155.cub
spiceinit from=c2061155.cub

The old "clean" versions of the cubes can be found in
/work/projects/EuropaBasemap/Europa_processed/VGR/

**Possible Solution**  
@lwellerastro noticed a difference in the 'SpacecraftClockCount' between the old "clean" cubes and the cubes now coming out of voy2isis. The new cubes have a '.' in the count, which is probably causing a read error (unexpected character). This almost certainly traces back to the instrument warning on ingest. It's not clear why the Instrument name is empty.

**Additional context**  
From @lwellerastro: There are a number of redmine posts (not all made it to github it looks like) that sort of have to do with this problem, #4421 in particular.  There are some voy2isis Brian Burns posts (closed) that have problems with ingesting pds data that required translation table changes that might not be helping.  Could be red herring as Ken says, so maybe not worth knowing about.  Brian did point out that the rings node as reprocessed the data and have created and archive containing .img/.lbl files but wasn't able to ingest them using voy2isis (#4345 on redmine and #2252 on github), and our response was probably wouldn't happen since it would require a program change.  Understandable.  However, their labels are good for these problem children, but we can't use them.

krlberry:
## Description
This updates `voy2isis` to add an optional parameter to specify to instrument name. This can be used when the `INSTRUMENT_NAME` is missing from the PDS archive image file to be ingested. 

This PR includes this change, a user documentation update, and a new test.

## Related Issue
This fixes #3426. 

## Motivation and Context
Three voyager .imq files from the PDS archive failed to appropriately ingest and spiceinit due to missing the INSTRUMENT_NAME in their original headers. 

## How Has This Been Tested?
A test was added to our testing suite for this specific case, and the existing voyager tests pass.

See #3426 for screenshots of the updated UI and output. 

## Types of changes

## Checklist:

## Licensing


isis/src/voyager/apps/voy2isis/main.cpp
```diff
@@ -484,16 +490,30 @@ QByteArray fixLabels(QString fileName, History *hist){
...
+    UserInterface &ui = Application::GetUserInterface();
+    if (ui.GetString("INSTRUMENT") == "AUTOMATIC") {
+      labels.replace("INSTRUMENT_NAME", "INSTRUMENT_NAME                  = Unknown");
+      PvlGroup insNameWarning("Warning"); 
+      PvlKeyword insNameMsg("Message", "The INSTRUMENT_NAME for [" + fileName + "] is empty."
+                                + "The InstrumentId in the output cube will instead be set to "
+                                + "[Unknown] and the labels will not translate. To create a cube "  
+                                + "with translated labels, re-run this "
+                                + "application with INSTRUMENT set to NAC or WAC.");
+
```
Kelvinrr, line 505: Should this be an error and exit?

jessemapel: I think raising an exception is a good idea because you cannot do further processing. If users want to just see the raw image data they can use pds2isis.

isis/src/voyager/apps/voy2isis/voy2isis.xml
```diff
@@ -55,6 +63,10 @@
     <change name="Christopher Combs" date="2018-01-19">
       Made changes to allow merge of two voy2isis tickets. References #4421, #4345.
     </change>
+    <change name="Kristin Berry" date="2019-11-06">
+      Added the INSTRUMENT parameter to allow the specification of the InstrumentId 
+      when it cannot be determined from the labels.
```
jessemapel: Add the github issue number to your history comment.

isis/src/voyager/apps/voy2isis/main.cpp
krlberry: @jessemapel `pds2isis` actually probably won't work in the cases where this message is created to be displayed later. The method this code is in fixes a malformed pds label so that it can be ingested to ISIS without failing. I went ahead and tried this on the test data associated with this ticket and `pds2isis` does indeed fail with: `**I/O ERROR** Unknown label type in [c2061155.imq]. It is possible the label file does not describe an image product (IMAGE, CUBE, or SPECTRALCUBE).` 

I didn't add this error message (or the particular way of saving it to display but not actually throwing an immediate exception) and I suspect being able to successfully create a valid cube with DNs is probably why. @Kelvinrr It will error and exit later if it still can't find an `INSTRUMENT_NAME` to use.

Kelvinrr: It still seems best to error out sooner than later. Even if you just wanted raw image data, getting a hard error and then re-running with a forced instrument name seems reasonable enough?

Kelvinrr (approved):
I'm lukewarm on error vs warning (I recall ISIS has no warnings?). Other than that it looks good to me.

jessemapel:
@Kelvinrr I agree, exception over a warning is better. ISIS does not handle warnings that well and users could miss it.

isis/src/voyager/apps/voy2isis/main.cpp
krlberry: @Kelvinrr @jessemapel  I'm reluctant to change the way that `voy2isis` currently functions on images with broken input labels unless that's also a desired change from the user end? The relevant change in behavior would be that no output cube would be created if an error in parsing the label was encountered, and the program would begin to exit with a non-zero exit code, both of which could break scripts that currently use this application. There are more ways that parsing the pds label can fail than just lacking an `INSTRUMENT_NAME`, for example the second half of that function deals with a potentially broken `IMAGE_ID` and it seems odd to handle them inconsistently. (We did just add the option to specify one by-hand, but not the other, _but_ the `IMAGE_ID` can be populated with an alternative value without problem.)

The strategy of "ingest the image with as much of the label as possible" was added as part of https://fixit.wr.usgs.gov/fixit/issues/4421 (not migrated to github?) which has substantial discussion I'm still looking through on it. Looking at the code, it seems to me that in the past, someone made a deliberate effort to not have the program crash if the label was malformed, and if we're going to modify this behavior I'd like to understand why before doing so.

jessemapel: It sounds like Tammy was pulling for doing as much as possible. In that case I'm okay with sticking with that.

jessemapel (approved):

Issue #3466, opened 2019-10-03
Problem with cam2map when processing CTX data


0dvaz0:
**ISIS version(s) affected**: 3.9.0 (running on Xubuntu 18.04 )

**Description**  
Hello,

I have recently upgrade ISIS and I started to have a problem when processing CTX data.
Cam2map does not run successfully when I define ground range latitude/longitude.
It seems to be running but the progress bar remains at 0% (even during a overnight run!).
There is no error messages and the cube that is created is all zeros.

However, if I run cam2map using all defaults, it returns a valid output.

Any idea why is that happening?
Am I doing something wrong?

Cheers
David

**How to reproduce**  

wget -c https://pds-imaging.jpl.nasa.gov/data/mro/mars_reconnaissance_orbiter/ctx/mrox_2174/data/F03_037038_1742_XI_05S227W.IMG

mroctx2isis from=F03_037038_1742_XI_05S227W.IMG to=F03_037038_1742_XI_05S227W.cub

spiceinit from=F03_037038_1742_XI_05S227W.cub

ctxcal from=F03_037038_1742_XI_05S227W.cub to=F03_037038_1742_XI_05S227W.lev1.cub

ctxevenodd from=F03_037038_1742_XI_05S227W.lev1.cub to=F03_037038_1742_XI_05S227W.lev1eo.cub

cam2map from=F03_037038_1742_XI_05S227W.lev1eo.cub to=F03_037038_1742_XI_05S227W.map.cub pixres=MPP resolution=5.4449133592799 defaultrange=MAP minlat=-5.0238176134696 maxlat=-3.1995325374047 minlon=131.92029413774 maxlon=132.75709469592

**Possible Solution**  

**Additional context**

lwellerastro:
I can reproduce this.  Data under /work/users/lweller/CTX/git_3466/.

Running in default mode, the app takes about 13 minutes or so to run:

> cam2map from=F03_037038_1742_XI_05S227W.cal.eo.cub to=F03_037038_1742_XI_05S227W.lev2Def.cub

When getting specific, either as the OP has in their command or whittling the res, lat and lon to just 4 decimal places, cam2map doesn't budge.  After 10 minutes or so of 0% progress I killed it.

I ran similar steps against Galileo SSI data I am working with and encountered no problems with getting very specific about lat/lon range, so it appears there is something going on with this particular CTX image, though it is not clear why.  It looks pretty standard.

[...]

PR #3497, opened 2019-11-07
Replaced code, removed earlier, that abandons work on a tile Fixes: #3466

scsides:
…is outside the output image

## Description
Previous work on ProcessRubberSheet removed an explicit test to abandon work on a tile if none of the four corners projected into the output cube. This test was put back to keep the algorithm from quadtreeing tiles that are outside the output cube.  It is possible for all four corners of a tile to be outside the output cube, but still have data that needs to be projected (e.g., a corner of the output cube intersects an edge of the tile, but not any corners, or the tile contains the entire output cube).

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/3466

## Motivation and Context

The issue reported a case where the program appeared to run forever. It was in fact running very slow (100's of time slower than before the test was removed). The slowness was magnified by the lat/lon range used. Only a small portion of the input image was in the output range.

## How Has This Been Tested?

Timing test were performed before and after the change. All times returned to their expected values (i.e., they were better than before the original change that removed the tests).

Putting the four corner test back in increased the performance of cam2map forward patch when a large portion of the input cube does not project into the output cube. The increase is estimated to be at least one order of magnitude. This change also slows down the same forward patch algorithm for case where most of the input cube projects into the output cube, but only slightly (estimated to be a few seconds).

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/ProcessRubberSheet/ProcessRubberSheet.cpp
+++ b/isis/src/base/objs/ProcessRubberSheet/ProcessRubberSheet.cpp
@@ -901,10 +901,6 @@ namespace Isis {
       osamps.push_back(tsamp);
       olines.push_back(tline);
     }
-    else {
-      splitPatch(ssamp, esamp, sline, eline, iportal, trans, interp);
-      return;
-    }
 
     // Upper right control point
     if (trans.Xform(tsamp,tline,esamp,sline)) {
@@ -912,10 +908,6 @@ namespace Isis {
       ilines.push_back(sline);
       osamps.push_back(tsamp);
       olines.push_back(tline);
-    }
-    else {
-      splitPatch(ssamp, esamp, sline, eline, iportal, trans, interp);
-      return;
     }
 
     // Lower left control point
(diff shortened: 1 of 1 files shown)


jessemapel (dismissed):

isis/src/base/objs/ProcessRubberSheet/ProcessRubberSheet.cpp
```diff
@@ -937,7 +925,15 @@ namespace Isis {
...
     }
-    else {
+
+    // If none of the 4 input tile corners transformed inside the output cube, give up on this tile
+    // NOTE: Probably should split if the tile is large
+    if (isamps.size() == 0) {
+      return;
+    }
+
+    // If at least one of the 4 input tile corners did NOT transform, spit it
```
acpaquette: Should probably be split instead of spit

acpaquette (commented):
This looks reasonable, just one spelling thing. From talking to @scsides this is mainly a revert to what the old code was doing

jessemapel (approved):

PR #3563, opened 2019-11-26
Env update to support building with ALE

Kelvinrr:
## Description

Tested working locally on CentOS, MacOS and Ubuntu. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #3567, opened 2019-12-02
🚧Pinned Bullet, PCL, and Boost for ALE integration🚧

jessemapel:
Un-pinning these was causing issues with Bullet seg-faulting, PCL not being found, and Boost failing to find certain modules.


jessemapel:
This is not ready to go 100% yet. I want to run the CI tests in a clean-environment.

acpaquette:
There are a good number of changes in here that might collide with #3563

jessemapel:
@acpaquette  Yeah. With Kelvin out for the next few days we need to decide what to do with that PR.

jessemapel:
I may rebase my changes onto it.

environment.yml
```diff
@@ -42,7 +43,7 @@ dependencies:
   - conda-forge/label/gcc7::nn
   - opencv>=3.0.0,<=4.0.0
   - openssl
-  - pcl
+  - pcl==1.8.1
```
Kelvinrr, line 46: If the problem with PCL is that it can't be found, this is an issue with the FindPCL script having passing the version number into the include finding call. The better solution would be to make it more flexible, whether by using the PCL-Config script that PCL ships with or by simply changing our findPCL to be more flexible with versions.

Kelvinrr:
Don't see why #3563 isn't mergeable as is? Should get nightlies building on everything other than CentOS and it at least gets the CentOS env file up to date with the others.

jessemapel:
@kelvinrr #3563 is failing after it builds. The nightly builds are building, but failing during testing.  It's failing in a strange place, and there is no code difference in that part.

jessemapel:
I'm going to close this right now and work on #3563

PR #3578, opened 2019-12-09
Working on a working Linux env

jessemapel:
Pinned x264 so that ffmpeg and opencv are good. Also modified find ALE script to look for the nested include directory and renamed it to match the exported target name.


krlberry (approved):
:+1: Looks like what we were working on yesterday + more to help cmake to find ale.

PR #3596, opened 2019-12-13
Getting CentOS environment working

jessemapel:
Now building on Jenkins.

### DIFF
--- a/environment_gcc4.yml
+++ b/environment_gcc4.yml
@@ -53,7 +53,7 @@ dependencies:
   - suitesparse
   - superlu
   - tnt
-  - x264
+  - x264=20131218
   - xalan-c
   - xerces-c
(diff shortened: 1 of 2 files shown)


environment_gcc4.yml
```diff
@@ -1,12 +1,12 @@
...
   - usgs-astrogeology
+  - conda-forge
+  - probcomp
   - defaults
 
 dependencies:
-  - libcxx
-  - cxx-compiler
-  - ale=0.6.4
+  - ale=0.7.2
```
jessemapel, line 9: Since I have to push a custom ALE build, I'm leaving this hard pinned to this version. Allowing this to be greater, would pull a higher version off conda-forge if available.

scsides (approved):

PR #3628, opened 2020-01-04
Moved spiceinit to a callable

jessemapel:
## Description

Moves spiceinit application to a callable function and tests it via GTest. I left the old app tests in place until we decide on an approach for broad app conversion.

## Related Issue

Continued test conversion

## Motivation and Context

This makes the spiceinit tests more flexible by explicitly testing for only certain things. This also reduces the time and data volume required to run the tests.

New tests take ~10 seconds, old tests take ~30 seconds.

Old test data volume is 27MB.

## How Has This Been Tested?

All tests have been run and passed on Ubuntu and Mac

## Types of changes

## Checklist:

## Licensing


jessemapel:
In case we leave this PR open for discussion, I wanted to outline the general process:

1. Create `spiceinit.cpp`
    1. Move all of the code from `main.cpp` into `spiceinit.cpp`
    1. Change `IsisMain()` to `void spiceinit(UserInterface &ui, Pvl *log)`
    1. Replace all calls to `UserInterface &ui = Application::GetUserInterface();` with passed in `ui`.
    1. Replace all calls to `Application::Log` with appending the group to the passed in `log` pointer.
1. Create `spceinit.h`
    1. Declare one function: `void spiceinit(UserInterface &ui, Pvl *log)`
1. Update `main.cpp`
    1. include `spiceinit.h`
    1. Replace contents of `IsisMain` with:
        1. Getting the UserInterface
        1. Calling the new `spiceinit` function
        1. On error or success, log all groups in the `log` pointer
1. Create spiceinitOptions struct in `spiceinit.h`
    1. Create a member variable for every optional argument in `spiceinit,xml`
    1. Set all of the member variables default values to the value in `spiceinit.xml`
1. Create a function in `spiceinit.cpp` that converts a UserInterface into a spiceinitOptions struct
    1. For all arguments with a default item/value you can just set the value.
    1. For other arguments, check that it was entered before over-writing.
1. Create `void spiceinit(Cube * icube, spiceinitOptions &options, Pvl *log)`
    1. Copy the contents of `void spiceinit(UserInterface &ui, Pvl *log)`
    1. For every helper function that takes a UserInterface, make a new one that takes a spiceinitOptions
    1. Everywhere that the UserInterface is used, instead check the spiceinitOptions
1. Replace the contents of `void spiceinit(UserInterface &ui, Pvl *log)` with a call to parse the UserInterface into a spiceinitOptions and opening the input/output Cube and then a call to `void spiceinit(Cube * icube, spiceinitOptions &options, Pvl *log)`.
1. Create the GTests
    1. For each old app test
        1. Convert the command line arguments into spiceinitOptions
        1. For any Cubes, create a temporary cube file and replace its label with the original test data's label.
        1. Call `void spiceinit(Cube * icube, spiceinitOptions &options, Pvl *log)`
        1. Check only what the test is looking for in the output Cube

isis/src/base/apps/spiceinit/spiceinit.cpp
```diff
@@ -0,0 +1,808 @@
...
+                  Kernel fk, Kernel ik,
+                  Kernel sclk, Kernel spk,
+                  Kernel iak, Kernel dem,
+                  Kernel exk);
+
+  void requestSpice(Cube *icube,
+                    Pvl &labels,
+                    QString missionName,
+                    const spiceinitOptions &options,
+                    Pvl *log);
```
Kelvinrr, line 47: If these functions are only called in the spiceinit function, it would be nice if they were functions closures. If not used repeatedly, then just inline them. Unfortunately, because they're written this way it might be easier to leave them all as they are.

isis/src/base/apps/spiceinit/spiceinit.h
```diff
@@ -0,0 +1,61 @@
...
+#include "Cube.h"
+#include "Pvl.h"
+#include "UserInterface.h"
+
+namespace Isis {
+
+  /**
+   * Struct specifying how to spiceinit a Cube.
+   */
+  struct spiceinitOptions {
```
Kelvinrr, line 17: Not sure about creating a separate place to maintain enumerated parameters, the goal should be to reduce units of complexity, not add new ones.

isis/src/base/apps/spiceinit/spiceinit.cpp
```diff
@@ -0,0 +1,808 @@
...
+    // Open the input cube
+    Process p;
+    CubeAttributeInput cai;
+    Cube *icube = p.SetInputCube(ui.GetFileName("FROM"), cai, ReadWrite);
+
+    spiceinitOptions options = getSpiceinitOptions(ui);
+
+    spiceinit(icube, options, log);
+
+    p.EndProcess();
```
Kelvinrr, line 66: It seems like the way to go for overloading for cube object support for now.

Kelvinrr (approved):
Just some thoughts. I'll run this on Jenkins later tonight when it's not getting hit by other devs and then merge.

isis/src/base/apps/spiceinit/spiceinit.h
jessemapel: This is my biggest problem with how I did this, but I couldn't find a good other way to create an options object without using the UI.

PR #3682, opened 2020-02-15
Cnet Winnow App Conversion and Functional Test

acpaquette:
Converts cnetwinnow into the main.cpp and associated cnetwinnow.cpp/.h files. This PR also removes the old app tests and adds a new cnetwinnow functional test.

## Description
Moves all of the application code from main.cpp into cnetwinnow.cpp with an updated interface to the application. Also adds a new gtest for cnetwinnow with a control network fixture.

## Motivation and Context
Mainly to remove the testing overhead and data needed by the test. Once converted we can remove about 5GB of data from the test data area.

## How Has This Been Tested?
Tested using the new gtest in the PR!

## Types of changes

## Checklist:

## Licensing


isis/src/control/apps/cnetwinnow/cnetwinnow.h
```diff
@@ -0,0 +1,21 @@
+#ifndef spiceinit_h
```
AustinSanders: Is this intended to be spiceinit_h or cnetwinnow_h?

isis/src/control/apps/cnetwinnow/cnetwinnow.h
```diff
@@ -0,0 +1,21 @@
...
+#define spiceinit_h
+
+#include <vector>
```
AustinSanders: Are these imports necessary? Specifically vector, QString, QList, QVector

isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
```diff
@@ -0,0 +1,437 @@
...
+    //build a histogram from the control net
+    Progress progress;
+
+    //read the ControlNet
+    ControlNet net(ui.GetFileName("CNET"), &progress);
+
+    cnetwinnow(net, progress, ui);
+  }
+
+  void cnetwinnow(ControlNet &net, Progress &progress, UserInterface &ui) {
```
Kelvinrr: From our conversation, I think we want to add the FileList object as a parameter too for these control apps.

isis/tests/CnetWinnowTests.cpp
```diff
@@ -0,0 +1,57 @@
...
+    count++;
+  }
+
+  Progress progress;
+  cnetwinnow(*network, progress, ui);
+
+  int postWinnowMeasureCount = network->GetNumValidMeasures();
+  int postWinnowPointCount = network->GetNumValidPoints();
+  ASSERT_EQ(postWinnowMeasureCount, 31);
+  ASSERT_EQ(postWinnowPointCount, 13);
```
Kelvinrr: I think it's worth asserting somehow that the network was also updated on disk? If so we might have to do this on all app tests going forward.

isis/tests/CnetWinnowTests.cpp
```diff
@@ -0,0 +1,57 @@
...
+  QList <ControlPoint*> pointList = network->GetPoints();
+
+  QList <ControlMeasure*> measures;
+  for (ControlPoint *point : pointList) {
+    measures = point->getMeasures();
+    for (ControlMeasure *measure: measures) {
+      measure->SetResidual(count, count);
+    }
+    count++;
+  }
```
Kelvinrr: If a count is being used anyways, a traditional for loop would probably have been better in this case.

isis/tests/CnetWinnowTests.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "Fixtures.h"
+
+#include "gmock/gmock.h"
+
+using namespace Isis;
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/cnetwinnow.xml").expanded();
+
+TEST_F(ThreeImageNetwork, CnetwinnowFunctionalDefault) {
```
Kelvinrr: Since I updated the docs to mention naming format last night, it saves me time renaming things later if you change this to FunctionalTestCnetwinnowDefault unless anyone has an issue with this format.

isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+
+  void cubeConvexHullAndMeasures(QString &serialNum,ControlNet &net, double &area, int &validMeasures,
+                                 QList <ControlMeasure *> *measToIgnor=NULL);
+
+  void cnetwinnow(UserInterface &ui) {
```
Kelvinrr: Both functions should have doc strings

isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
acpaquette: I agree but the a fromlist is only used to get the serial numbers as a serialNumList so I think that makes more sense to pass in this particular case

isis/src/control/apps/cnetwinnow/cnetwinnow.h
acpaquette: No, they are not. I will remove them

isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
Kelvinrr: Another param question, is it necessary to pass a progress object? Is it because of the control net constructor? Any way to simply pass in the network object then associate the cnet object with the progress object after the fact?

acpaquette: Looks like it defaults to zero in the ControlNet.h file so we can default it in the cnetwinnow call as well

acpaquette: We can at least get around needing it if we just want to call cnetwinnow.

isis/tests/CnetWinnowTests.cpp
acpaquette: I agree I just don't know how worth while it is. In this case I just open the network in a new ControlNet object and check for the same results.

Kelvinrr (dismissed):
Not sure what to do here since it seems the Jenkins server is dead. If this was tested locally feel free to merge.

isis/tests/CnetWinnowTests.cpp
Kelvinrr: I think simply checking that the file exists is enough considering the object has definitely changed.

isis/src/control/apps/cnetwinnow/main.cpp
```diff
@@ -1,435 +1,19 @@
...
-void cubeConvexHullAndMeasures(QString &serialNum,ControlNet &net, double &area, int &validMeasures,
-                          QList <ControlMeasure *> *measToIgnor=NULL);
-
-
 void IsisMain() {
   // Get user interface
   UserInterface &ui = Application::GetUserInterface();
-
-
   //build a histogram from the control net
```
jessemapel: Remove old comment

isis/tests/CnetWinnowTests.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "FileName.h"
+#include "SerialNumber.h"
+
+#include "Fixtures.h"
+
+#include "gmock/gmock.h"
+
+using namespace Isis;
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/cnetwinnow.xml").expanded();
```
jessemapel, line 20: These should be defined in the app source/header instead of in the test files.

Kelvinrr: These paths exist exclusively for tests, putting them elsewhere implies otherwise and not sure why we would want to do that. If we ever wanted a real interface to these apps it would include something that doesn't require users to know what this XML is or what it's for.

AustinSanders (dismissed):

Kelvinrr (approved):

PR #3688, opened 2020-02-19
Autoseed, Footprintinit, and 1/2 of Findimageoverlaps

acpaquette:
Converts autoseed, footprintinit, and findimageoverlaps apps into the assocaited .cpp, .h and main.cpp files. Also adds tests for autoseed, and footprintinit

## Description
Converts all apps into the new format with a set of cpp and h file along with the associated main method. It also adds new gtests for autoseed and footprintinit. A test for findimageoverlaps will also be added.

## Motivation and Context
Updates the ISIS api and reduces testing overhead for the modified methods.

## How Has This Been Tested?
This is the test

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #3692, opened 2020-02-19
Added temp dir for test files

jessemapel:
## Description

This creates a temporary directory to house temp files generated during tests. The temporary directory is then cleaned up once the test exits. I also changed the test cube files so that they are generated with the proper suffix.

## Related Issue

No related issue, this is part of the GTest conversions.

## Motivation and Context

The current implementation sometimes leaves .cub files dangling. This will also be important when testing applications like jigsaw that generate a range of files in the same directory.

## How Has This Been Tested?

Existing GTests have all been tested and still work. Makefile and old unit tests are un-affected.

## Types of changes

## Checklist:

## Licensing


jessemapel:
Once this is merged #3693, #3682, and #3688 can be updated. Or if they are merged before this I can update the code.

isis/tests/Fixtures.cpp
```diff
@@ -2,14 +2,26 @@
...
 
+  void TempTestingFiles::SetUp() {
+    ASSERT_TRUE(tempDir.isValid());
+  }
+
+
   void DefaultCube::SetUp() {
+    TempTestingFiles::SetUp();
+
+    tempFile.setFileTemplate(tempDir.path() + "/XXXXXX.cub");
```
acpaquette: Are the templates still necessary if the temp directory as a whole (files included) will be cleaned up or am I missing something?

acpaquette (commented):
Other than the one comment this looks good!

isis/tests/Fixtures.cpp
jessemapel: The template is required to put the file in the temp directory and by adding `.cub` we ensure that we only make one file.

scsides (dismissed):

AustinSanders (dismissed):

kaitlyndlee (dismissed):
Looks good!

isis/tests/Fixtures.cpp
Kelvinrr: I don't see where in the docs it says the files under the temp dir need to be tempfile objs. It seems like any file in the directory gets cleaned up regardless of how the file got there. Why not create the temp dir, save the path (or tempdir object) as a member of the fixure, and make the underlying file names match 1-1 with what is in the test/data/<fixture> area?

jessemapel: The files in the tempdir don't need to be tempfiles also. I'll change these to just regular files.

isis/tests/cnetcheckFunctionalTests.cpp
```diff
@@ -76,27 +72,24 @@ TEST_F(ThreeImageNetwork, FunctionalTestCnetcheckNoPoints) {
   // remove all controlpoints from test network
   QList<ControlPoint*> cpoints = network->GetPoints();
 
-  for (size_t i = 0; i<cpoints.size(); i++) {
+  for (int i = 0; i<cpoints.size(); i++) {
```
Kelvinrr, line 75: Why int? In this case, it will probably not matter but seems like a weird change.

Kelvinrr (dismissed):
Only a benign comment in-line.

isis/tests/cnetcheckFunctionalTests.cpp
jessemapel: It raised a warning because QT containers return ints for size instead of unsigned ints.

Kelvinrr:
My vote right now would be to get @acpaquette's PRs merged and then update this if no one has an issue with it. #3682 only needs one quick change to be mergeable. #3682 I trust will be mergeable by the end of the day.

Kelvinrr:
Alright, change of plans. PRs are taking too long to merge and I'd rather not make it Jesse's problem. I am going to merge this as it is also fixing a bug in the original DefaultCube Fixture and everyone else will update their PR. I'll update part-timer's PRs.

Kelvinrr (approved):

PR #3694, opened 2020-02-20
Cmake update to link mission libraries with gtest executable

krlberry:
## Description

This PR updates the `CMakeList` file for our gtests to link the `runISISTests` executable it generates against the other isis libraries available in addition to just "libisis3," including the shared libraries built for each mission (for example `libnewhorizons`.) Only the core modules defined in: 
https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/CMakeLists.txt are actually linked against without this. 

Before this change, linking errors result from adding any tests to our gtest suite for applications outside of the core modules. I ran into this when adding gtests for `leisa2isis`. This work is PR'd in separately, as this may be useful for anyone else working on tests outside the core modules. 

## Related Issue
None, work for testing sprint. Issue uncovered during development.

## Motivation and Context
Allows a gtest to be written for `leisa2isis` and gtests to be written for other mission applications.

## How Has This Been Tested?
The gtest for `leisa2isis` complies and runs instead of failing with a linking error with these changes. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [ ] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

- [ ]

PR #3699, opened 2020-02-20
Update cmake for gtest again to successfully link against shared libraries

krlberry:
Follow on to PR #3694 

## Description
The GLOB in the CMakeLists doesn't work without the .so or .dylib. Apologies -- I removed the extensions from the prior PR and thought I had tested, but I had ISIS set to the wrong build-dir. This is not the best solution and I only tested it on Linux. I'm building it on prog26 now to test mac, but wanted to put this PR up in case needed.

## Related Issue
Testing work. 

## Motivation and Context
See previous PR. 

## How Has This Been Tested?
Works on Linux locally.

## Types of changes

## Checklist:

## Licensing


jessemapel (approved):
This feels kind of like a hack, but I don't think we have time to do this correctly, and it won't make it harder to do it correctly in the future.

isis/tests/CMakeLists.txt
```diff
@@ -3,13 +3,16 @@ cmake_minimum_required(VERSION 3.10)
 add_dependencies(isis3 isis3)
 
 file(GLOB test_source "${CMAKE_SOURCE_DIR}/tests/*.cpp")
-file(GLOB isis_libs "${CMAKE_BINARY_DIR}/lib/lib*")
+
+# Non-ideal temporary solution to link against other shared libraries for mission tests, etc
+file(GLOB isis_libs "${CMAKE_BINARY_DIR}/lib/*.so")
+file(GLOB isis_mac_libs "${CMAKE_BINARY_DIR}/lib/*.dylib")
```
Kelvinrr, line 9: oof

Kelvinrr (commented):
oof

Kelvinrr (approved):
oof

krlberry:
I 100% agree that this is a blatant kludge :/
I'll put in a ticket to fix this later

isis/tests/CMakeLists.txt
krlberry: I'm agreeing with the oof. Putting in a ticket to actually fix this later.

Kelvinrr:
I'm gonna wait for PRs to stabilize a little then merge this.

isis/tests/CMakeLists.txt
scsides: This looks like it will forcibly link against all libraries including the plugins that are not supposed to be loaded at initial run time. May not actually test the plugins correctly. As long as the new ticket has this documented, I'm ok with it for now.

isis/tests/CMakeLists.txt
```diff
@@ -3,13 +3,16 @@ cmake_minimum_required(VERSION 3.10)
...
 
 # Link runISISTests with what we want to test and the GTest and pthread library
 add_executable(runISISTests
                IsisTestMain.cpp
                ${test_source})
 
-target_link_libraries(runISISTests isis3 ${isis_libs} ${ALLLIBS} ${GTEST_LIBRARIES} ${GTEST_MAIN_LIBRARIES} Threads::Threads)
+target_link_libraries(runISISTests isis3 ${isis_libs} ${isis_mac_libs} ${ALLLIBS} ${GTEST_LIBRARIES} ${GTEST_MAIN_LIBRARIES} Threads::Threads)
```
scsides, line 16: I realize one of the two lists of libs should always be empty, but why not put  *.so and *.dylib into one list (isis_libs) and not have the os specific isis_mac_lib even exist OR have a list populated based on the OS?
But is should work, so I'm ok with this too

scsides (approved):

PR #3704, opened 2020-02-21
leisa2isis app test conversion and one gtest replacement

krlberry:
## Description
Converts `leisa2isis` application to callable and add a gtest. Slight modification to `ProcessImport` made to support this. (Not api-breaking) 

## Related Issue
Test sprint. 

## Motivation and Context
This is the first ingestion app to be converted and there are some obvious areas of improvement that I didn't get to, including: 
* Update ProcessImport's `StartProcess` to accept a functor or lambda-capture-style function because this _still_ has a global variable in it (which I renamed to `outputCubeLeisa2Isis` but still)
* Needs coverage for more than one replacement gtest
* Needs to test DNs, not just labels.
* The cmake update to support this was a kludge and a better solution needs to be implemented. See ticket #3705.
* Some way to not just use completely binary FITS files -- ideally we could at least swap out labels to test various options, but I don't think detached labels are supported, and ISIS doesn't have a FITS class, so I'm not 100% sure how to do this right now.

## How Has This Been Tested?
The created test passes.

## Types of changes

## Checklist:

## Licensing


isis/tests/FunctionalTestsLeisa2isis.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // Archive Group
+  PvlGroup &archive = isisLabel->findGroup("Archive", Pvl::Traverse);
+  ASSERT_DOUBLE_EQ(double(archive["MidObservationTime"]), 225897372.0736388);
+  ASSERT_EQ(archive["MidObservationTime"].unit().toStdString(), "s past J2000");
+  ASSERT_DOUBLE_EQ(double(archive["ObservationDuration"]), 251.0);
+  ASSERT_EQ(archive["Detector"][0].toStdString(), "LEISA" );
+  ASSERT_EQ(archive["ScanType"][0].toStdString(), "LEISA" );
+
+  // BandBin Group
+  // Check size, first, 2 middle, and last values? Enough?
```
krlberry, line 65: Feedback on if this is reasonable? Couldn't think of a good way to test all of them at once.

isis/src/newhorizons/apps/leisa2isis/main.cpp
```diff
@@ -1,307 +1,27 @@
...
-      in[i] = Isis::NULL8;
-    } 
-    else {
-        in[i] = 0;
-    }
-  }
-  outputCube->write(in); 
-}
-
+}
```
scsides: Silly I know, but no CR  at EOF. I'll add it and save.

scsides (approved):
Good to go

isis/src/base/objs/ProcessImport/ProcessImport.cpp
```diff
@@ -1263,6 +1259,24 @@ namespace Isis {
...
+   *
+   * @return @b Isis::Cube Output cube.
+   *
+   * @throws Isis::iException::Message "Unsupported pixel type."
+   */
+  Isis::Cube *ProcessImport::SetOutputCube(const QString &parameter) {
+    CubeAttributeOutput &att =
+      Application::GetUserInterface().GetOutputAttribute(parameter);
+
+    SetAttributes(att);
```
acpaquette, line 1279: This looks fine, but I don't understand why this change was made?

acpaquette (approved):
Just one question the rest of this looks fine!

isis/tests/FunctionalTestsLeisa2isis.cpp
acpaquette: Unless looping over cube is worth while I think just checking a few lines within the cube is good. That or we could deapp-ify cubediff. :grimacing:

isis/src/newhorizons/apps/leisa2isis/main.cpp
krlberry: Thanks!

isis/src/base/objs/ProcessImport/ProcessImport.cpp
krlberry: @acpaquette This change was made more-or-less to avoid changing `ProcessImport` too much. The application parameter version of `SetCube("TO")` sets up the attributes's pixel propagation and min/max values for you, but the `SetCube(filename, attributes)` does not. We need to use the second function signature now in `leisa2isis` since `ProcessImport` no longer has access to the UI to handle the "TO" for you, but the attributes aren't set up in this way in in this version and this resulted in an error along the lines of "you told me to propagate pixel type, but not what the type is." I moved SetAttributes out so I could call it directly from `leisa2isis` without changing the existing functionality of any of the functions inside `ProcessImport`

Issue #3669, opened 2020-02-05
Make smithed ck kernels for Enceladus available in ISIS


blandoplanet:
**Description**  
In FY19 the USGS (i.e., @lwellerastro) generated a new set of ck for a subset of Cassini ISS images of Enceladus (n=621, which is most of the moderate to high resolution images). The kernels will soon be available on Astropedia, along with a description of the method by which they were generated. Ideally, these kernels would also be available as an option in ISIS as well (e.g., using 'cksmith=true' in spiceinit). The complication is that the kernels also require a new pck to be specified and it's not clear to me how (or even whether) the pck can be handled.

The kernels are available here...
/work/users/lweller/Enceladus/SmithedKernels/Enceladus_CISS_2019Shape_camera.bc
/work/users/lweller/Enceladus/SmithedKernels/cpck15Dec2017_2019Shape.tpc

At the moment they can applied simply by running
spiceinit from=image.cub ck=Enceladus_CISS_2019Shape_camera.bc pck=cpck15Dec2017_2019Shape.tpc

lwellerastro:
n=621 should be n=625

jessemapel:
@blandoplanet Is there any reason why a user would use smithed CKs, but not the updated PCK? There are already some smithed CKs from CICLOPS that are of the form YYDOY_YYDOYcb_ISS.bc. I'm assuming your kernels should take precedence over these?

blandoplanet:
Users should always use the updated PCK with the new CK. There isn't a case for using the CKs alone.

I am unfamiliar with the CICLOPS CK kernels, and don't know how they were derived or what images they cover. I would argue that the new USGS kernels should take precedence, but that can be discussed.

jessemapel:
@blandoplanet @lwellerastro How were these images spiceinit'd originally? I'm running into a problem where the segment boundaries in the smithed CK are smaller than the exposure times of the images. For the first image, N1487299402_1, it's time range is 2005-048T02:16:57.245 to 2005-048T02:16:57.295 but the segment boundaries in the CK are 2005-048 // 02:16:57.267 to2005-048 // 02:16:57.273, which doesn't cover the full 50 ms exposure.

For the vast majority of uses, ISIS only needs the center exposure time, which is covered, and I can adjust the DB file that spiceinit uses so it looks like the kernel covers the entire image, but I'm concerned about that being a band-aid fix.

blandoplanet:
[...]

PR #3716, opened 2020-02-27
Used CK quality for PCK selection in spiceinit

jessemapel:
## Description

This change uses the same kernel quality options for PCKs and CKs in spiceinit. This ensures that if target body parameters are solved for along-side sensor pointing, both updated kernels will be loaded.

This is a backwards compatible change because no PCK kernel DBs have type specification and hence ignore the allowed types check. Future kernel dbs that do have type specifications in them will operate differently with versions of ISIS before and after this change.

## Related Issue

#3669 

## Motivation and Context

As part of the Enceladus global mosaic created by @blandoplanet & @lwellerastro they generated updated pointing kernels and an update PCK. In order to load the correct PCK, we needed a way to specify that the updated one was desired. This was approach was selected because it peserves existing behavior and doesn't require users to enter new parameters (such as having a PCK quality parameter).

## How Has This Been Tested?

All tests passed locally on osx10.13 and ubuntu18

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/spiceinit/spiceinit.cpp
+++ b/isis/src/base/apps/spiceinit/spiceinit.cpp
@@ -23,7 +23,7 @@
 
 using namespace std;
 
-namespace Isis { 
+namespace Isis {
 
   void getUserEnteredKernel(UserInterface &ui, const QString &param, Kernel &kernel);
   bool tryKernels(Cube *icube, Process &p, UserInterface &ui, Pvl *log,
@@ -42,16 +42,16 @@ namespace Isis {
    * @param ui The Application UI
    * @param(out) log The Pvl that attempted kernel sets will be logged to
    */
-  void spiceinit(UserInterface &ui, Pvl *log) { 
+  void spiceinit(UserInterface &ui, Pvl *log) {
     // Open the input cube
     Process p;
-    
+
     CubeAttributeInput cai;
     Cube *icube = p.SetInputCube(ui.GetFileName("FROM"), cai, ReadWrite);
     spiceinit(icube, ui, log);
     p.EndProcess();
   }
- 
+
 
   /**
    * Spiceinit a Cube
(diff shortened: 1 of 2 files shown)


isis/src/base/apps/spiceinit/spiceinit.cpp
```diff
@@ -144,15 +144,15 @@ namespace Isis {
       Kernel lk, pck, targetSpk, fk, ik, sclk, spk, iak, dem, exk;
       QList< priority_queue<Kernel> > ck;
       lk        = baseKernels.leapSecond(lab);
-      pck       = baseKernels.targetAttitudeShape(lab);
+      pck       = ckKernels.targetAttitudeShape(lab);
```
krlberry, line 147: I don't see a problem with this, since all previous pcks didn't have any quality value specified. In the future, do you think the ability to specify something like `pckpredicted=true` would be useful?

krlberry (approved):

isis/src/base/apps/spiceinit/spiceinit.cpp
jessemapel: I considered this, but the current use case is when we solve for both instrument pointing and body rotation in the bundle adjust which doesn't need that. Also, for the current use case not requiring users to enter a separate parameter seemed better because there's no way to check that they entered one but not both in the kerneldb.

scsides: This seems like a good solution. Even if we needed to add "pcksmithed" in the future this would not get in the way.

scsides (approved):

acpaquette (approved):
Looks good! :+1:

krlberry:
Any idea if the `1128 - cassini_module_test_vims (Failed)` failure on jenkins (mac only) actually has anything to do with this PR? I tried to compare with dev, but apparently dev hasn't built for a few days?

jessemapel:
@kberryUSGS I only touched the smithed kernels and that test uses reconstructed so I'm not sure. I'll still check though.

jessemapel:
Cassini VIMS test failures were due to an SPK update on February 14th, not these changes.

jlaura:
#3716 is merged. Is this closable?

Issue #3727, opened 2020-03-03
Move the application data into the git repo


scsides:
**Description**  
The ISIS application specific data (e.g., PDS3 to ISIS cube translation) are currently mixed in with the general data (e.g., Cassini/kernels/ck/*). Move the application specific data into the git repository and out of the ISIS data area.

Addition:
The changes proposed to have many, not all, text files under revision control include:
* During build and install steps: Move files to: $ISISROOT/appdata/... from the src tree
  * Move the template files
  * Move translation files 
  * Move serial number files

**Example**

KrisBecker:
Originally in #3728.

While I agree this is technically not _API_  breaking as defined by the current policy, it potentially breaks a lot of stuff.  It appears this type of change should be covered by policy.  Perhaps a period of deprecation could be specified where both environment variables, ISIS3DATA and ISISDATA, refer to the same value to ease the transition and provide an environment in which users can assess the impact in their environments. It might be a good idea to treat this version as a "breaking change" release.

Here are but a few potential issues I have identified:

### ISIS3DATA is hardcode in the ISIS appdata branch
Most of these occurrences are in unit tests, some are in main applications and others are in comments and should probably be changed for consistency. However, in some cases, the change would not be backward compatible (is this a breaking change?). Its likely these issues have already been identified, but here is a list of the occurrences in the appdata branch that directly use $ISIS3DATA:

```
(isis3/appdata)$ find ./isis/src -name '*.cpp' -exec grep -A1 -C0 -H ISIS3DATA {} \;
./isis/src/voyager/apps/voy2isis/main.cpp:  QString lsk = "$ISIS3DATA/base/kernels/lsk/naif????.tls";
./isis/src/voyager/apps/voy2isis/main.cpp:  QString sclk = "$ISIS3DATA/voyager";
./isis/src/tgo/objs/TgoCassisCamera/TgoCassisDistortionMap.cpp:    // of the CCD. (See $ISIS3DATA/tgo/assets/distortion/DistortionModelA3CorrRoots.jpg).
./isis/src/tgo/objs/TgoCassisCamera/TgoCassisDistortionMap.cpp:    // of the CCD. (See $ISIS3DATA/tgo/assets/distortion/DistortionModelA3DistRoots.jpg).
./isis/src/rosetta/objs/RosettaVirtisCamera/unitTest.cpp:    Cube visCube("$ISIS3DATA/rosetta/testData/V1_00388238556.cub", "r");
[...]

PR #3742, opened 2020-03-06
Update LRO appdata

scsides:
Initial commits for moving lro small files out of ISISDATA and into git
## Description
Initial commits for moving lro small files out of ISISDATA and into git

## Motivation and Context
RFC4

## How Has This Been Tested?
NOTE: Many tests are still failing. Each mission is being converted and PRed separately. When base is PRed many tests will start working. Leftovers will be fixed one at time

## Types of changes

## Checklist:

## Licensing


KrisBecker (commented):
I am a bit concerned to see absolute paths using $ISISROOT being used here. This appears to be heading in the same direction as our current ISIS3DATA->ISISDATA issue.

What was nice about the relative translation file scheme was that you had more control over isolating/insulating your work using a local IsisPreferences configuration. Looks like this option is being abandoned?

jessemapel:
The ability to change your translation files and such is still available, just instead of having to specify a different data directory, you change the files in your source tree the same way you would change the code.

I can definitely see us having to make all of these changes again if we change the package name. Would it be worth putting some sort of function to get the build root and have that access ISISROOT? That way if we change the ISISROOT variable it's only a change in 1 place instead of having to change all of these again?

krlberry:
This looks like a good addition to the discussion about this work on issue #3727 that shouldn't get lost when this PR is closed, since this comment is about what we're doing in general and applies to multiple PRs. @KrisBecker @jessemapel  could you move your comments on this PR to the issue at #3727?

KrisBecker:
@jessemapel...

> The ability to change your translation files and such is still available, just instead of having to specify a different data directory, you change the files in your source tree the same way you would change the code.

On more than one occasion, I have used IsisPreferences for testing problems in a live, running system with minimal changes and without the need to completely copy the system elsewhere or directly alter any system file in the running tree. This can change the environment just enough so that the behavior you are testing cannot be reproduced. This PR does not preserve this capability.

jessemapel:
@krisbecker Can you provide a more specific example of the functionality you are concerned about?

KrisBecker:
@jessemapel you have two examples in the current [IsisPreferences ](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/IsisPreferences) file in the repo. The Clipper and Osirisrex data directories show a path outside the shipped directory structure. 

```
Clipper      = $ISIS3DATA/../datalocal/clipper
OsirisRex    = $ISIS3DATA/../datalocal/osirisrex
```
How is it possible to isolate translation files in the source tree and not ship them prematurely?

scsides:
@KrisBecker @jessemapel, This discussion seems related to every PR that is or will be part of this branch. It is important to capture all of these comments in the related issue. Can we please move this over to the issue?  #3727

krlberry (approved):
:+1: Looks good to me.

PR #3790, opened 2020-03-23
Getsn tests

tgiroux:
## Description
Reformatted the getsn app in isis/src/base/apps/getsn/
Rewrote tests with gtest in isis/tests/FunctionalTestsGetsn.cpp
Removed old tests in isis/src/base/apps/getsn/tsts

## Related Issue
n/a

## Motivation and Context
The motivation is to bring the tests up to date and implement newer technology through google tests.
Some context is that this project has been an opportunity for me (Tim Giroux) to get familiar with ISIS for the first time.

## How Has This Been Tested?
In the build directory run `ctest -VV -R Getsn` to run the new tests.
This runs FunctionalTestsGetsn.cpp and gives 7 / 7 tests passed.
At the moment, if you run `ctest -R getsn`, one of the old tests (getsn_app_test_default) is still recognized. It runs and fails.

n/a

## Types of changes

## Checklist:

## Licensing


Kelvinrr:
What happened to the template? Even if a lot of it doesn't apply to a test change you can leave the template as is and put N/A or leave parts blank.

tgiroux:
> What happened to the template? Even if a lot of it doesn't apply to a test change you can leave the template as is and put N/A or leave parts blank.

My bad, it should be better now.

isis/src/base/apps/getsn/main.cpp
```diff
@@ -1,93 +1,40 @@
...
-  // Extract label from cube file
-  Pvl *label = cube.label();
-
-  PvlGroup sn("Results");
+  Pvl appLog;
 
-  if (WriteFile) sn += PvlKeyword("Filename", from);
-  if (WriteSN) sn += PvlKeyword( "SerialNumber", SerialNumber::Compose( *label, ui.GetBoolean("DEFAULT") ) );
-  if (WriteObservation) sn += PvlKeyword( "ObservationNumber", ObservationNumber::Compose( *label, ui.GetBoolean("DEFAULT") ) );
+  getsn(ui, &appLog);
```
jessemapel, line 30: You need to wrap this in a try catch block so that it properly logs even if the function errors. See https://github.com/USGS-Astrogeology/ISIS3/blob/d9049c1941301837fce014ea59f046af1c69c6a6/isis/src/base/apps/spiceinit/main.cpp#L15

isis/tests/FunctionalTestsGetsn.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "TestUtilities.h"
+#include "Cube.h"
+
+#include "gmock/gmock.h"
+
+using namespace Isis;
+
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/getsn.xml").expanded();
+static QString FROM_PARAM = "FROM=data/defaultImage/defaultCube.pvl";
```
jessemapel: This implicitly links this test to the fixture. If we change the location of the fixture data or the name of the label file it will need to be updated here as well. The path to the label should either be a member of the fixture or a static variable that is defined in the Fixture code so that it only needs to be updated in a single place.

acpaquette: This path should be available from the DefaultCube fixture from either the cube or the pvl label, more likely from the cube. If we get the path from that it will reduce other instances of changing paths like this if the data area ever ends up changing.

acpaquette (commented):
Small issue with the from file in the tests (Already mentioned in Jesse's review) but other than that everything looks good!

isis/tests/FunctionalTestsGetsn.cpp
Kelvinrr: Considering that he is using the interface with the cube being passed in, from doesn't have to even be part of the args and it should be able to be eliminated altogether.

jessemapel: I agree, it would be better to use the method that passes in a Cube object. You can use Cube::fileName to get the path for checking the output too.

isis/src/base/apps/getsn/getsn.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+
+  void getsn( Cube *cube, UserInterface &ui, Pvl *log ) {
+
+    // Determine if output should be written base on parameters
+    bool WriteFile = ui.GetBoolean("FILE");
+    bool WriteSN = ui.GetBoolean("SN");
+    bool WriteObservation = ui.GetBoolean("OBSERVATION");
+
+    QString from = ui.GetFileName("FROM");
```
Kelvinrr: Replace this with `QString from = cube->fileName()` otherwise the function will fail if the parameter doesn't exist.

isis/src/base/apps/getsn/getsn.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // Determine if output should be written base on parameters
+    bool WriteFile = ui.GetBoolean("FILE");
+    bool WriteSN = ui.GetBoolean("SN");
+    bool WriteObservation = ui.GetBoolean("OBSERVATION");
+
+    QString from = ui.GetFileName("FROM");    
+    QString format = ui.GetString("FORMAT");
+    bool pvl;
+    if (format == "PVL") {
```
AustinSanders, line 42: I think this could be simplified by changing the conditional to:
```
bool pvl = format == "PVL";

File: isis/tests/FunctionalTestsGetsn.cpp [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+
+// check for all correct outputs
+TEST_F(DefaultCube, FunctionalTestGetsnAllTrue) {
+  QString expectedFilename = "data/defaultImage/defaultCube.pvl";
+  QString expectedSN = "Viking1/VISB/33322515";
+  QString expectedON = "Viking1/VISB/33322515";
+  QVector<QString> args = {FROM_PARAM,
```
AustinSanders: Since we're passing the testCube directly to getsn, do we still need to set the FROM_PARAM?

isis/src/base/apps/getsn/getsn.cpp
Kelvinrr: that kind of boolean grossness is kinda all over ISIS, this PR is about simply getting tests in and this is the copy-pasted code from main.cpp, this sort of thing can be handled some other time if it's a big enough issue.

isis/src/base/apps/getsn/main.cpp
Kelvinrr: If the function errors, the exception that is thrown naturally is still logged, is swallowing the error an re-interpreting it a good idea?

Kelvinrr (dismissed):
I'm fine with it as is after tests pass.

isis/src/base/apps/getsn/main.cpp
jessemapel: This caused an issue in spiceinit where the exception didn't report what kernel sets were attempted prior to failure, that was done via the log. In this case, I think the exception will be sufficient.

jessemapel (dismissed):

isis/src/base/apps/getsn/getsn.cpp
scsides: Depends on what you call gross. The ifs are easy to read and there is no question as to operator precedence.  They are also easy to add additional code inside of. Where the collapsed version would have to be expanded first.

```
bool pvl = false;
if (ui.GetString("FORMAT") == "PVL") {
  pvl = true;
}

scsides (commented):
Was jigsaw supposed to get into this PR?

scsides (changes requested):
See my jigsaw question before merging

Kelvinrr:
@scsides  good catch, definitely not. ~~Tim must have accidentally added them somehow?~~ I think it's changes from the dev branch I originally branched from.

jessemapel (approved):
The jigsaw changes were from things not being rebased 100% properly.

Kelvinrr:
removed the jigsaw changes even though I'm like 95% sure they were inconsequential changes already in dev.

jessemapel:
> removed the jigsaw changes even though I'm like 95% sure they were inconsequential changes already in dev.

They were the changes from the Jigsaw multisensor PR and wouldn't have caused any actual changes. When that happens, it doesn't impact the final merged code, it just clutters the PR and the squashed merge commit.

scsides (approved):

Kelvinrr:
Tests that failed also failed in dev before this PR, so it seems clean.

Issue #3659, opened 2020-01-24
Enlarge Reporting the final lines and samples backwards


AaronBoyd:
**ISIS version(s) affected**: 3.4.1(and probably before) through and including 3.9.1

**Description**  
Enlarge reports the OutputLines as the output samples and the OutputSamples as the output lines

**How to reproduce**  
run enlarge (and sscale and lscale)

view the results group generated to see the switch

**Possible Solution**  
this appears to be just a typo that needs to be corrected.

**Additional context**  

  Group = UserParameters
    FROM   = image1
    TO     = image2
    INTERP = nearest
    MODE   = SCALE
    SSCALE = 1
    LSCALE = 14
  End_Group

  Group = Results
    InputLines      = 78
    InputSamples    = 704
    StartingLine    = 1
    StartingSample  = 1
    EndingLine      = 78
    EndingSample    = 704
    LineIncrement   = 0.071428571428571
    SampleIncrement = 1.0
    OutputLines     = 704
    OutputSamples   = 1092
  End_Group

jessemapel:
Yep, looks flipped https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Enlarge/Enlarge.cpp#L144

jlaura:
![](https://github.trello.services/images/mini-trello-icon.png) [Enlarge Reporting the final lines and samples backwards](https://trello.com/c/7QOSiU5s/82-enlarge-reporting-the-final-lines-and-samples-backwards)

acpaquette:
@ladoramkershner Can this be closed?

PR #3797, opened 2020-03-24
correcting swapped line sample prints in enlarge application output file

ladoramkershner:
Corrected the enlarge application output file prints.

## Description

Swapped the line and sample variables in the enlarge output file prints to accurately represent values.

## Related Issue

Issue #3659 

## Motivation and Context

Issue #3659 

## How Has This Been Tested?

Existing tests for enlarge were run and passed. Since this is just an output file change and does not change the functionality of the application.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/Enlarge/Enlarge.truth
+++ b/isis/src/base/objs/Enlarge/Enlarge.truth
@@ -1,4 +1,16 @@
 Testing Enlarge Class ... 
+Group = Results
+  InputLines      = 126
+  InputSamples    = 126
+  StartingLine    = 1
+  StartingSample  = 1
+  EndingLine      = 126
+  EndingSample    = 126
+  LineIncrement   = 0.5
+  SampleIncrement = 0.5
+  OutputLines     = 252
+  OutputSamples   = 252
+End_Group
 Object = IsisCube
   Object = Core
     StartByte   = 65537
--- a/isis/src/base/objs/Enlarge/Enlarge.cpp
+++ b/isis/src/base/objs/Enlarge/Enlarge.cpp
@@ -141,8 +141,8 @@ namespace Isis {
     resultsGrp += PvlKeyword("EndingSample",    toString((int)mdEndSample));
     resultsGrp += PvlKeyword("LineIncrement",   toString(1. / mdLineScale));
     resultsGrp += PvlKeyword("SampleIncrement", toString(1. / mdSampleScale));
-    resultsGrp += PvlKeyword("OutputLines",     toString(miOutputSamples));
-    resultsGrp += PvlKeyword("OutputSamples",   toString(miOutputLines));
+    resultsGrp += PvlKeyword("OutputLines",     toString(miOutputLines));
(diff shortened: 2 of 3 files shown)


krlberry (dismissed):
:+1:

jessemapel (commented):
Change is good, but this probably slipped through because the results group generated by this is never checked in the unittest! A quick fix would be to update the old style unit test to capture the output of this function and print it, but it's probably a good candidate for converting to gtest. Things get a little complicated with the SubArea class, but it all looks do-able via mocks.

ladoramkershner:
Is there a place where we can make a note of this change for the gtest conversion? I see that there is a trello card for the conversion of the enlarge test, but this seems like a larger issue and we may want to test all application outputs. I am not sure where to put that information/request.

jessemapel:
@ladoramkershner We could open a new issue and fix this as a separate PR. I think it makes more sense to make the fix to the tests under this PR though.

ladoramkershner:
But don't we want it on every application that has an output? Making one issue for all application with a write out seems like too large of an issue, but making one issue per application sounds tedious. I think making a note that it should be added in during gtest conversion would be good (although very similar to making one large blanket issue)

jessemapel:
> But don't we want it on every application that has an output?

I am unsure of what you are reference with it. I'm concerned that the output of this function is being discarded in the unit test:

https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Enlarge/unitTest.cpp#L34

ladoramkershner:
I am meaning to say, is it common practice to test the output file of a function in isis unit tests? If yes, then a single ticket makes sense. If not, then one large ticket to say we should START testing file outputs seems non-digestable in any single sprint.

jessemapel:
Ah. I'm not talking about file output, but the actual return of the function. Enlarge::UpdateOutputLabel returns a PvlGroup and in the unit test for the Enlarge class, it is simply thrown away and never checked. The unit test should be updated to properly check the function return so that this doesn't regress.

ladoramkershner:
Oh okay. Sorry! Okay I can make a ticket for this!

ladoramkershner:
So I added the Results group into the truth file, but I am realizing that the isisTruth.cub used for the test is square, so it would not have caught the error Aaron reported anyways 😆

isis/src/base/objs/Enlarge/Enlarge.truth
```diff
@@ -1,4 +1,16 @@
...
+  InputSamples    = 126
+  StartingLine    = 1
+  StartingSample  = 1
+  EndingLine      = 126
+  EndingSample    = 126
+  LineIncrement   = 0.5
+  SampleIncrement = 0.5
+  OutputLines     = 252
+  OutputSamples   = 252
+End_Group
```
jessemapel, line 13: Make sure that you update unittest.cpp to output this too!

jessemapel (approved):
Looks good. Once CI passes this can be merged.

PR #3757, opened 2020-03-13
2.5D Occlusion and Testing Update

acpaquette:
Allows cam2map to account for occlusion when projecting images. Also updates cam2map with the application changes and test, and the application change for map2cam.

## Description
Allows cam2map to account for occluded portions of images when they are map projected. This option can only be used with the reverse patch algorithm and is turned off by default to keep any currently used pipelines for generating map projected products to remain the same.

This PR also updates cam2map and map2cam to be callable as functions. The tests for cam2map where also updated with additional fixtures as well as a few new mocks.

## Related Issue
This is related to a larger occlusion project with no specific issues on the ISIS repo.

## Motivation and Context
This allows ISIS to more accurately project highly oblique images, so long as the image is accurately coregistered to its associated DEM.

## How Has This Been Tested?
This has been tested with the current ISIS tests along with a new set of gtests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #3798, opened 2020-03-24
Fixed cam2map ignoring band specifications on the input cube

jessemapel:
## Description

With the recent conversion of cam2map to a callable, the input Cube was changed to be manually opened instead of being opened by a Process object. This resulted in the input Cube not getting its input attributes (the only input Cube attribute is band selection FYI).

This is going to be a common problem we convert more of these applications. I considered adding a new openWithAttributes option to the Cube class, but I'm not sure if that is the correct dependency. For right now I simply copied the code from Process::SetInputCube.

Input on a good way to handle this so that we don't have 50 copies of this code is appreciated.

## Related Issue

No issue, fixes a bug introduced in #3757

## Motivation and Context

This is functionality that was temporarily lost, so we need to add it back. This was discovered because thmproc's app test uses cam2map's band selection via a Pipeline object.

## How Has This Been Tested?

Tests, except for currently nightly build failures, all passed on Ubuntu and existing cam2map and thmproc tests passed on OSX. I also manually tested that calling cam2map with from=some_lvl1.cub+2 properly projected only the second band.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/cam2map/cam2map.cpp
+++ b/isis/src/base/apps/cam2map/cam2map.cpp
@@ -22,7 +22,7 @@ namespace Isis {
   void cam2map(UserInterface &ui, Pvl *log) {
     // Open the input cube
     Cube icube(ui.GetFileName("FROM"));
-    CubeAttributeInput inAtt(ui.GetFileName("FROM"));
+    CubeAttributeInput inAtt(ui.GetAsString("FROM"));
     if (inAtt.bands().size() != 0) {
       icube.setVirtualBands(inAtt.bands());
     }
--- a/isis/src/base/apps/cam2map/cam2map.cpp
+++ b/isis/src/base/apps/cam2map/cam2map.cpp
@@ -22,7 +22,7 @@ namespace Isis {
   void cam2map(UserInterface &ui, Pvl *log) {
     // Open the input cube
     Cube icube;
-    CubeAttributeInput inAtt(ui.GetAsString("FROM"));
+    CubeAttributeInput inAtt = ui.GetInputAttribute("FROM");
     if (inAtt.bands().size() != 0) {
       icube.setVirtualBands(inAtt.bands());
     }
--- a/isis/src/base/apps/cam2map/cam2map.cpp
+++ b/isis/src/base/apps/cam2map/cam2map.cpp
@@ -1,6 +1,7 @@
 #include "cam2map.h"
 
 #include "Camera.h"
+#include "CubeAttribute.h"
 #include "IException.h"
 #include "IString.h"
 #include "ProjectionFactory.h"
@@ -21,6 +22,10 @@ namespace Isis {
   void cam2map(UserInterface &ui, Pvl *log) {
     // Open the input cube
     Cube icube(ui.GetFileName("FROM"));
+    CubeAttributeInput inAtt(ui.GetFileName("FROM"));
+    if (inAtt.bands().size() != 0) {
+      icube.setVirtualBands(inAtt.bands());
+    }
 
     // Get the map projection file provided by the user
     Pvl userMap;
--- a/isis/src/base/apps/cam2map/cam2map.cpp
+++ b/isis/src/base/apps/cam2map/cam2map.cpp
@@ -21,11 +21,12 @@ namespace Isis {
 
   void cam2map(UserInterface &ui, Pvl *log) {
     // Open the input cube
-    Cube icube(ui.GetFileName("FROM"));
+    Cube icube;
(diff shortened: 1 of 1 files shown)


jessemapel:
Apologies for the gtest update, had to update mine for the GMock stuff that was added and it pulled the latest version. Shouldn't cause any issues though.

isis/src/base/apps/cam2map/cam2map.cpp
```diff
@@ -20,7 +21,12 @@ namespace Isis {
 
   void cam2map(UserInterface &ui, Pvl *log) {
     // Open the input cube
-    Cube icube(ui.GetFileName("FROM"));
+    Cube icube;
+    CubeAttributeInput inAtt(ui.GetAsString("FROM"));
+    if (inAtt.bands().size() != 0) {
+      icube.setVirtualBands(inAtt.bands());
+    }
```
Kelvinrr, line 28: What exactly are virtual bands?

jessemapel: virtual bands are how ISIS handles only working with some of the bands in an image. The way that this showed up is running cam2map on a Themis IR image with 3 bands but only wanting to project the first band. So, running `cam2map from=themis_image.cub+1 to=just_the_first_band.cub` should only map project the first band. This is handled in the API by setting the virtual bands to "1".

acpaquette:
Can this not be done when the cube is opened from a file? Or does that create some odd circular dependency between cube and cubeattributes?

jessemapel:
@acpaquette That's what I think is best, but I'm concerned about a circular dependency. It honestly feels like CubeAttributes should been a part of the Cube class anyway. I'm nervous that there's some weird reason why that wasn't done when this was written.

acpaquette:
There are Cube create functions that take an CubeAttributeOutput object along with a file name to create output cubes, I don't see why we couldn't do the same thing with the CubeAttributeInput. It looks like the main problem we would potentially run into is removing it from the Process's setInputCube

acpaquette:
As it stands open seems like it's used for input cubes and create/copy are used for output cubes. I think if we just add the attribute logic to open to also read the input attributes we'd be good to go.

jessemapel:
Sorry, I didn't clearly put what I was looking for. A big part of this is the difference between `UserInterface::GetFileName` and `UserInterface::GetAsString`. Both return a QString, but GetFileName strips off anything after the file extension (I.E. `CubeAttributeInput inAtt(ui.GetFileName("FROM"));` doesn't fix this because it strips off the +1 component). I'm concerned about people not knowing that you have to both create a CubeAttributeInput and use the less common GetAsString. Something like passing the UI to the open call could work, but that seems horrendous.

jessemapel:
A good way to address this may be a function on the UserInterface class that opens and returns the input cube with its attributes properly set.

jessemapel:
I'm going to leave this PR as is and put in any changes to better address handling input cube attributes in a different PR.

krlberry (dismissed):
:+1: This looks good to me! I agree with the comments about finding a more-optimal generalized way to handle this problem, but also think it is reasonable to merge as-is, as long as a ticket goes in about the need for handling cube attribute input in general for the app conversions. It's probably worth putting in a ticket and referencing this issue.

scsides:
I'm concerned about the use of of GetAsString in any case. The UI was designed to send back objects of the correct type which is what getFilename is doing. Why ask the  UI for a string and then convert it to an integer when you know the value is an integer? It's type safety.

It seems like the callable apps interface is pushing repetitive work back into the apps (i.e., open the cube) when it was only for special cases.

I like the idea of having cube open deal with the attributes, but don't forget output cubes have attributes too.

scsides:
Disagree, the UI should not be opening cubes

jessemapel:
> I'm concerned about the use of of GetAsString in any case. The UI was designed to send back objects of the correct type which is what getFilename is doing.

getFilename is stripping off the attribute because it converts the string into a FileName object and then back to a string via FileName::expanded which removes all attributes. We could change getFilename to return an actual FileName object and then add a Cube::open(FileName) that takes attributes into account.

AustinSanders (dismissed):

scsides:
FYI, The AsString calls were for communication between GUI and command line, and testing.
Couldn't we use the UI:GetAttributes or as you suggest add a get CubeName. I agree, attributes are the responsibly of the Cube.

krlberry (approved):
:+1: switching to using GetInputAttribute() is a nice improvement.

scsides (approved):
That works.

AustinSanders (approved):

PR #3851, opened 2020-04-23
Simplified Env file 

Kelvinrr:
## Description

This minimizes the ISIS env to better support third party libraries along with ISIS. Also get's a lot of libraries on the latest conda-forge recipes. 

Some notes: 

- Boost deprication warnings in Boost's endian.hpp, this is actually from PCL and are not using deprication warnings and instead using preprocessor messages so we can't simply suppress the warnings nor can we update since the code is in PCL's base. 
- DatabaseFactory tests fail on Linux because the new Qt build which includes doesn't include Kristin's update to include MySQL support. 
- Some gtests are failing, but I believe these are may be unrelated, rather related to #3839 
- ~~FindFeatures tests are failing, because we are using a more up to date OpenCV, might want to consider transitioning to gtest and write tests that are more robust to OpenCV version~~ Findfeatures is failing because the newest version does not have some feature extraction algorithms, notably SIFT. The patent for SIFT apparently has expired very recently and the flag was removed from the SIFT in opencv (see https://github.com/opencv/opencv_contrib/pull/2453/files and related issue https://github.com/opencv/opencv_contrib/pull/2449). Not 100% sure why this change isn't reflected on conda-forge, chances are since the change was back in march, a release wasn't pulled yet. 

## Motivation and Context

Making ISIS dependencies more flexible 

## How Has This Been Tested?

Ran tests locally 

## Types of changes

## Checklist:

## Licensing


environment.yml
```diff
@@ -4,59 +4,36 @@ channels:
...
-  - doxygen
-  - eigen
-  - embree
-  - geos>=3.7,<3.8
-  - geotiff
-  - gmm
-  - gmp
+  # Huge Depedencies 
+  - qt
+  - gdal>=3
```
Kelvinrr, line 9: Gdal and ISIS share a lot of dependencies, it's basically easier to just include gdal rather than maintaining a bigger list. This only makes sense if we plan to incorporate Gdal into ISIS itself (which I believe is part of the plan moving forward), otherwise, I can remove this and expand it out with more specific dependencies.

scsides: I understand allowing gdal to pull in xerces, but the last time we did this that didn't allow xalan (requries older version of xerces) to be pulled in and I don't see xalan in the list anymore. xalan is required to build the documentation. We don't see this because we are lazy and don't build the docs often.

scsides (changes requested):
We should look into updating the conda build files too. The previous release was difficult because we didn't do them when we did the environment.yml

environment.yml
Kelvinrr: Thanks for pointing it out, updated with Xalan-c.

jessemapel: For the environmeny.yml we can do this, but for the meta.yml we don't need xalan if we include gdal.

jlaura:
@Kelvinrr @scsides How might this PR impact the isis2std JPEG work? Is this diverging too far that we should close this PR?

scsides:
Not sure that it will cause any problems, but even if it does both still need to be done.

PR #3852, opened 2020-05-01
Version ticks for 3.10.2 stable release

AustinSanders:
Ticked the version numbers and set variables for 3.10.2 release.

## Description
Ticked the version numbers and set variables for 3.10.2 release. Copied contents of environment.yaml to meta build/run sections.

## Related Issue
https://trello.com/c/IUrphCSv/362-do-the-release

## Motivation and Context
Readies the package for release.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -46,6 +46,7 @@ requirements:
   build:
   - armadillo==8.200.0
   - blas==1.1
+  - boost==1.67
   - bullet==2.86.1
   - bz2file==0.98
   - bzip2==1.0.6
@@ -77,7 +78,7 @@ requirements:
   - ninja==1.7.2
   - conda-forge/label/gcc7::nn
   - openblas==0.2.19
-  - opencv
+  - opencv==3.2.0
   - openssl==1.0.2n
   - pcl==1.8.1
   - pip==9.0.1
@@ -104,6 +105,7 @@ requirements:
   run:
   - armadillo==8.200.0
   - blas==1.1
+  - boost==1.67
   - bullet==2.86.1
   - bz2file==0.98
   - bzip2==1.0.6
@@ -132,7 +134,7 @@ requirements:
   - nanoflann==1.2.2
   - conda-forge/label/gcc7::nn
   - openblas==0.2.19
-  - opencv
+  - opencv==3.2.0
   - openssl==1.0.2n
   - pcl==1.8.1
   - pip==9.0.1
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -107,13 +104,11 @@ requirements:
   run:
   - armadillo==8.200.0
   - blas==1.1
-  - boost==1.67
```
jessemapel: Why is this removed? Xalan and Doxygen are not needed because they are for building docs.

recipe/meta.yaml
```diff
@@ -129,15 +124,15 @@ requirements:
...
+  - libtiff>=4.0.9
   - libxml2==2.9.7
   - mesalib==17.2.0
   - mysql==5.7.20
   - mysql-connector-c==6.1.6
   - nanoflann==1.2.2
   - conda-forge/label/gcc7::nn
   - openblas==0.2.19
-  - opencv==3.2.0
+  - opencv
```
jessemapel: Opening up OpenCV can break a lot of things. We have this hard pinned to 3.2.0 so that we get the "non-free" packages and the correct interface for findfeatures. Why was this changed?

jessemapel (commented):
Version # and CMake changes look good. Why was the environment changed?

recipe/meta.yaml
AustinSanders: I copied the dependencies directly from the environment.yml dependencies section.  Boost was not in environment.yml.

recipe/meta.yaml
AustinSanders: The section was copied from environment.yml, which does not specify a version for opencv.

AustinSanders: I'll change environment back to including boost + explicitly versioning opencv, but should I also change the environment.yml file?  Or change the docs that state that these sections should be copied from environment.yml?  Things seem a little out of sync here.

jessemapel: The two files are out of synce because of how channels are specified in the two. In an environment file we can specify a channel priority, in a conda-build meta.yml file we cannot. Having things more pinned in the meta.yaml helps ensure all of the correct packages are installed without the channel priority. For the 4.1 release, modifying the meta.yaml make sense as we are working on dependencies, but the 3.10 release doesn't have any of that in it, so we want to keep using the old environment.

jessemapel (approved):

Issue #3875, opened 2020-05-20
mappt report DN values for all bands (or at least the band for the DN value)


lavalaz:
**Description**  

Currently mappt only reports the DN value for a single band. To get the values for different bands at the same location you need to run mappt repeatedly for each band.  But then there is no record of which band the DN value came from in the output file.  This greatly reduces the usability of mappt.  It would be nice if the band was labeled not just with the band number but also the name so one did not need to retain the original .cub file to know what the numbers in the output file mean. 
**Example**

jessemapel:
Good candidate for work alongside #3872

Kelvinrr:
PR is up for this change: https://github.com/USGS-Astrogeology/ISIS3/pull/3890

This adds band and filter info as part of the output. 

Some questions:

1. I am not sure whether multiple band output is part of the requirements? The first half of the post implies it is but the rest implies you only want the band info to be part of the output? 
2. PR adds physical band number and filter name as part of the output. I understand some cubes will not have filter names and simply have filter number, is the current behavior in the PR of only reporting filter name if available otherwise default to "Null" the desired behavior?

lavalaz:
1. Just recording which band the DN value corresponds to is acceptable.  Being able to output the values of multiple bands would be nice but may be impractical given how varied (and large) the number of bands can be.  
2. "Null" if there is no name for the band is fine.  Thanks!

PR #3890, opened 2020-06-02
mappt will now report physical band number

Kelvinrr:
## Description

Mappt now prints out the physical band number as part of its output. Instead of adding a band param seems like the right thing to do is leverage cube attributes (e.g. from=some.cub+3). 

full example: 

```bash 
(isisdev) igswzawglt0091:isisbuild kelvin$ ./bin/mappt from=/work/users/lweller/Isis3Tests/Mappt/MultiBand/MVA_2B2_01_02024S140E3586.lev2.cub+2  to=here.cub type=IMAGE line=181 sample=41
Group = Results
  Filename                 = /work/users/lweller/Isis3Tests/Mappt/MultiBand/M-
                             VA_2B2_01_02024S140E3586.lev2.cub
  Sample                   = 41.0
  Line                     = 181.0
  Band                     = 2
  FilterName               = MV2
  PixelValue               = 45.214
  X                        = -51291.269237485
  Y                        = -419456.23284515

  # Input map coordinate system
  PlanetocentricLatitude   = -13.832779919698
  PositiveEast180Longitude = -1.6914776408294

  # Location in other coordinate systems
  PlanetographicLatitude   = -13.832779919698
  PositiveEast360Longitude = 358.30852235917
  PositiveWest360Longitude = 1.6914776408294
  PositiveWest180Longitude = 1.6914776408294
End_Group
```

## Related Issue

#3875

I would like some clarifications I asked for on the issue cleared up by Laz before merging this. 

## Motivation and Context

This should make mappt more useful for tracking a DN's band origin. 
## How Has This Been Tested?

Tested locally, test data still needs to be updated. 

## Types of changes

## Checklist:

## Licensing


isis/src/base/apps/mappt/main.cpp
```diff
@@ -157,15 +158,24 @@ void IsisMain() {
...
   int intLine = (int)(proj->WorldY() + 0.5);
+  
   b.SetBasePosition(intSamp, intLine, 1);
   icube->read(b);
+  
+  QString filterName = "Null";
+  PvlGroup bandBin = icube->label()->findObject("IsisCube").findGroup("BandBin");
+  if (bandBin.hasKeyword("FilterName")) {
+      filterName = bandBin.findKeyword("FilterName")[0];
+  }
```
Kelvinrr: From what I understand the `BandBin` group isn't super homogenized because of physical differences between instruments. That detail makes this change seem questionable to me.

jessemapel: I am also dubious of this change. We could pull the band value from all array keywords in the BandBin group.

Kelvinrr: If you're suggesting is getting DNs for every band, I am not sure if this solves the problem in Laz's original issue of wanting more than a physical band number. He seems fine with only one band output. 

I just don't know enough about how ISIS handles multiple bands to answer whether filter name is the only thing useful here? If I report FilterNumber (e.g. in THEMIS images), is filter number and physical band number always going to be the same therefore a useless thing to report? Are there other ways that bands are reported that might be useful to consider? Laz only cares about filter name so that's all I included but might as well make this more generic if it's easy enough to do and I am already in the code.

jessemapel: No, I'm not talking about pulling for every band. The BandBin group works by having array-valued keywords where the values at the same index are for the same band. So, the second value in each keyword is for the second band. Instead of just reporting filter name, you could copy the first value from each array-valued keyword in the BandBin group to the output.

For example here's the BandBin group from a TOG CaSSIS mosaic:

```
  Group = BandBin
    FilterName = (BLU, NIR, PAN, RED)
    Center     = (485, 985, 675, 840) <nm>
    Width      = (165, 220, 250, 100) <nm>
    NaifIkCode = (-143424, -143423, -143421, -143422)
  End_Group
```

You would then put the following in the output:
```
FilterName = BLU
Center     = 485 <nm>
Width      = 165 <nm>
NaifIkCode = -143424
```

Kelvinrr: I considered that but my only concern there is having an output params change based on the cube making outputs a pain to parse, at that point I might as well slap the bandbin group with only the first virtual band at the end of results (if only nested groups were a thing).  

I can make the change and if no one else objects\has a better solution then I don't have an issue either.

jessemapel: True, parsing output could be a pain. I think this can just stay as is then.

jessemapel (commented):
Code changes look good, just needs testing.

isis/tests/FunctionalTestsMappt.cpp
```diff
@@ -0,0 +1,171 @@
...
+  QVector<QString> args = {"type=image", "sample=-1", "line=-1", "allowoutside=true"};
+  UserInterface options(APP_XML, args);
+  Pvl appLog;
+
+  mappt(projTestCube, options, &appLog);
+  PvlGroup groundPoint = appLog.findGroup("Results");
+  EXPECT_DOUBLE_EQ( (double) groundPoint.findKeyword("Sample"), -1.0);
+  EXPECT_DOUBLE_EQ( (double) groundPoint.findKeyword("Line"), -1.0);
+  
+  // TODO: Doesn't throw error when allowoutside=false
```
scsides: Does this still need to be done?

Kelvinrr: ~yeah, my thinking is that it should be a separate issue.~

nvm, this was an actual mistake, I assumed it would throw an exception when outside, it instead appends to the PVL, I updated the PR to check the PVL rather than expecting an exception.

scsides (dismissed):
that works

isis/tests/FunctionalTestsMappt.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+TEST_F(DefaultCube, FunctionalTestMapptImageTest) {
+  QVector<QString> args = {"append=false", "type=image", "sample=1", "line=1"};
+  UserInterface options(APP_XML, args);
+  Pvl appLog;
+
+  mappt(projTestCube, options, &appLog);
+  PvlGroup mapPoint = appLog.findGroup("Results");
+  
+  std::cout << mapPoint << std::endl;
```
jessemapel: Why are the tests printing the output to cout?

isis/tests/FunctionalTestsMappt.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, mapPoint.findKeyword("FilterName"), "CLEAR");
+  EXPECT_EQ( (double) mapPoint.findKeyword("Band"), 1);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PlanetographicLatitude"), 9.3870849567571);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PlanetocentricLatitude"), 9.2788326719634);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveWest360Longitude"), 359.14528612684);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveEast360Longitude"), 0.85471387315749);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveEast180Longitude"), 0.85471387315749);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveWest180Longitude"), -0.85471387315751);
```
jessemapel: These needs to also check that the output contains the correct Filename, PixelValue, X, and Y

isis/tests/FunctionalTestsMappt.cpp
```diff
@@ -0,0 +1,173 @@
...
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, mapPoint.findKeyword("FilterName"), "CLEAR");
+  EXPECT_EQ( (double) mapPoint.findKeyword("Band"), 1);
+  EXPECT_NEAR( (double) mapPoint.findKeyword("Sample"), 1, 1e-8);
+  EXPECT_NEAR( (double) mapPoint.findKeyword("Line"), 1, 1e-8);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PlanetographicLatitude"), 9.3870849567571);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PlanetocentricLatitude"), 9.2788326719634);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveWest360Longitude"), 359.14528612684);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveEast360Longitude"), 0.85471387315749);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveEast180Longitude"), 0.85471387315749);
+  EXPECT_DOUBLE_EQ( (double) mapPoint.findKeyword("PositiveWest180Longitude"), -0.85471387315751);
```
jessemapel: This needs to check that the output contains the correct X and Y projection values.

isis/tests/FunctionalTestsMappt.cpp
```diff
@@ -0,0 +1,173 @@
...
+    QString line = flatStream.readLine();
+    QStringList fields = line.split(",");
+
+    if(lineNumber == 0) {
+      EXPECT_PRED_FORMAT2(AssertQStringsEqual, fields.value(1), "Sample");
+      EXPECT_PRED_FORMAT2(AssertQStringsEqual, fields.value(2), "Line");
+    }
+    else if(lineNumber == 1) {
+      EXPECT_DOUBLE_EQ(fields.value(1).toDouble(), 1);
+      EXPECT_DOUBLE_EQ(fields.value(2).toDouble(), 1);
```
jessemapel: This needs to check all of the columns of the CSV:
Sample, Line, PixelValue, X, Y, PlanetocentricLatitude, PlanetographicLatitude, PositiveEast360Longitude, PositiveEast180Longitude, PositiveWest360Longitude, PositiveWest180Longitude

You could run the app with each output once and then just check that the PVL and flatfile outputs are the same.

jessemapel (commented):
Tests need to check the full output. Once that's done this looks good to go.

jessemapel (approved):

PR #3897, opened 2020-06-04
Added the ability to skip special pixels in phocube

kaitlyndlee:
## Description

Added a parameter that defaults to true for backwards compatibility. This parameter allows users to skip any calculations on special pixels and sets the value to NULL in all bands except the DN band if selected. Needed to merge two process methods to propagate the input cube buffer to the main process method. There was concern that this would slow down phocube, but after some time tests, this does not seem that case.

## Related Issue

https://github.com/USGS-Astrogeology/ISIS3/issues/3646

## Motivation and Context

Has the possibility to speed up phocube if a user wishes to skip special pixels by not calling the setImage() method in Camera.

## How Has This Been Tested?

Created a new test called ignore_special_pixels and existing phocube tests pass.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/phocube/main.cpp
+++ b/isis/src/base/apps/phocube/main.cpp
@@ -287,9 +287,10 @@ void IsisMain() {
         // If dn is true, make sure to not overwrite the original pixels with NULL for the dn band
         if (!specialPixels && IsSpecial(in[inIndex])) {
           for (int band = (dn) ? 1 : 0; band < nbands; band++) {
-            out[index] = Isis::NULL8;
+            out[index] = Isis::Null;
             index += 64 * 64;
           }
+
           continue;
         }
 
@@ -462,16 +463,14 @@ void IsisMain() {
           for (int band = (dn) ? 1 : 0; band < nbands; band++) {
             if(ra && band == raBandNum) {
               out[index] = cam->RightAscension();
-              index += 64 * 64;
             }
             else if (declination && band == raBandNum + 1) {
               out[index] = cam->Declination();
-              index += 64 * 64;
             }
             else {
-              out[index] = Isis::NULL8;
-              index += 64 * 64;
+              out[index] = Isis::Null;
             }
+            index += 64 * 64;
           }
         }
       }
--- a/isis/src/base/apps/phocube/main.cpp
+++ b/isis/src/base/apps/phocube/main.cpp
@@ -16,44 +16,6 @@
 using namespace std;
 using namespace Isis;
 
-// Global variables
-Camera *cam;
-TProjection *proj;
-int nbands;
-bool noCamera;
-bool specialPixels;
-
-bool dn;
-bool phase;
-bool emission;
-bool incidence;
-bool localEmission;
-bool localIncidence;
-bool latitude;
-bool longitude;
-bool pixelResolution;
-bool lineResolution;
-bool sampleResolution;
-bool detectorResolution;
-bool obliqueDetectorResolution;
-bool northAzimuth;
-bool sunAzimuth;
-bool spacecraftAzimuth;
-bool offnadirAngle;
-bool subSpacecraftGroundAzimuth;
-bool subSolarGroundAzimuth;
-bool morphologyRank;
-bool albedoRank;
-bool ra;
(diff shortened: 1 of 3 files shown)


isis/src/base/apps/phocube/main.cpp
```diff
@@ -21,6 +21,7 @@ Camera *cam;
 TProjection *proj;
 int nbands;
 bool noCamera;
+bool specialPixels;
```
jessemapel: More globals is bad. Use a lambda with capture instead https://github.com/USGS-Astrogeology/ISIS3/wiki/Writing-ISIS3-Tests-Using-Gtest-and-Ctest#process-helper-functions-and-global-variables

jessemapel: If you have time it would be good to convert all of the globals to captures, but at least we shouldn't add to them.

isis/src/base/apps/phocube/main.cpp
```diff
@@ -532,17 +514,19 @@ void phocube(Buffer &out) {
...
-          else if (declination && b == raBandNum + 1) {
+          else if (declination && band == raBandNum + 1) {
             out[index] = cam->Declination();
+            index += 64 * 64;
           }
           else {
             out[index] = Isis::NULL8;
+            index += 64 * 64;
           }
-          index += 64 * 64;
```
jessemapel, line 545: Why did this move inside the if/if-else/else block? This is the same but it's more confusing now.

isis/src/base/apps/phocube/main.cpp
```diff
@@ -333,42 +319,38 @@ void IsisMain() {
...
+
+      if (dn) {
+        out[index] = in[index];
+        index += 64 * 64;
+      }
+
+      // If dn is true, make sure to not overwrite the original pixels with NULL for the dn band
+      if (!specialPixels && IsSpecial(in[inIndex])) {
+        for (int band = (dn) ? 1 : 0; band < nbands; band++) {
+          out[index] = Isis::NULL8;
```
jessemapel: Why NULL8 and not just NULL?

isis/src/base/apps/phocube/main.cpp
```diff
@@ -333,42 +319,38 @@ void IsisMain() {
...
+          index += 64 * 64;
+        }
+        continue;
+      }
```
jessemapel: All Cubes default allocate to NULL pixels. You can skip the band selection logic here by just doing `if (!specialPixels && IsSpecial(in[inIndex])) continue;`

isis/src/base/apps/phocube/main.cpp
kaitlyndlee: I was following the behavior of setting pixels off the body to NULL8 on line 527. Should this be set to NULL and we leave the current behavior of the off-body pixels? Or should they both be NULL?

isis/src/base/apps/phocube/main.cpp
kaitlyndlee: I originally wrote it that way because I thought it made the code more compact (in a previous change I worked on about a year ago). I then decided to move the line into the if/else block because I thought it increased readability. I can change it back if the cons outweigh the pros.

isis/src/base/apps/phocube/main.cpp
jessemapel: I mentioned this in the stand-up but I want it documented here for future folks. These should all be changed to Isis::Null because they are not dependent upon the bit-type of the image. In general, we use the specific bit-type special pixel values when the logic depend on the bit-type of the image and the standard special pixel values when the logic is valid for all bit-types.

isis/src/base/apps/phocube/main.cpp
jessemapel: Turns out this was a bad suggestion. The cube defaults to Null, but the buffer defaults to 0 and since it never reads from the output image before writing to it, it will fill stuff with 0 instead of Null. Since it would take more time to read from the output cube, it's better to just leave this assignment in.

jessemapel (approved):
Looks solid

PR #3900, opened 2020-06-05
Update export to PDS4 of PolarStereographic-projected ISIS cubes.

### LINKED ISSUE #3869: tgocassisrdrgen doesn't work with integer, rather than floating-point values in the input label

**Description**  
`tgocassisrdrgen` will fail when an input label is provided with numerical values with integer, rather than floating-point values are used in specific places, for example, the `StraightVerticalLongitudeFromPole` keyword. This problem will affect _ALL_ pds4 export done currently by ISIS based on the level the problem exists at.

Example of failed run: 
```
  Group = Error
      Program = tgocassisrdrgen
      Class   = "PROGRAMMER ERROR"
      Code    = 3
      Message = "Unable to translate and export mapping group"
      File    = ProcessExportPds4.cpp
      Line    = 185
    End_Group
  
    Group = Error
      Program = tgocassisrdrgen
      Class   = "USER ERROR"
      Code    = 2
      Message = "Unable to export projection [PolarStereographic] to PDS4
                 product. This projection is not supported in ISIS3"
      File    = ProcessExportPds4.cpp
      Line    = 1194
    End_Group
  
    Group = Error
      Program = tgocassisrdrgen
      Class   = "PROGRAMMER ERROR"
      Code    = 3
      Message = "Unable to find a translation value for
                 [StraightVerticalLongitudeFromPole, 90] in file
                 [/apps/opt/anaconda/Linux.x86_64/envs/isis400/../data/base/tran-
                 slations/pds4ExportPolarStereographic.trn]"
      File    = PvlTranslationTable.cpp
      Line    = 280
    End_Group
  End_Object
  End
```
**How to reproduce**  
Run `tgocassisrdrgen` on a cube with `StraightVerticalLongitudeFromPole=90`

**Possible Solution**  
I'm not sure about the best solution without looking into this more, but the problem is that we're using translation files for numerical values in a way that I don't think translation files were intended to be used.

**Additional context**  
Issue reported via email by Rod Heyd at the University of Arizona / CaSSIS team, and a work-around was identified, but this should be fixed.

krlberry:
## Description
The incorrect translation was being used to calculate the output value for straight vertical longitude from pole for polar stereographic projections for exporting ISIS map-projected cubes to PDS4 (new translation was verified with Trent.) The actual problem in the original ticket is fixed due to removing the numerical values in the translation file. 

## Related Issue
Fixes #3869

## Motivation and Context
The CaSSIS pipeline was failing when integer values were used for the center latitude in the mapping group. It turns out that center latitude should not have been used in this way in the first place.

## How Has This Been Tested?
Works on both cubes provided by the CaSSIS team as an example of the problem (listed in ticket.)

Test data for tgocassisrdrgen not updated in the system yet, but will do so after merge.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/appdata/translations/pds4ExportPolarStereographic.trn
+++ b/isis/appdata/translations/pds4ExportPolarStereographic.trn
@@ -248,13 +248,12 @@ End_Group
 
 Group = StraightVerticalLongitudeFromPole
   Auto
-  InputKey       = CenterLatitude
+  InputKey       = CenterLongitude
   InputGroup     = "IsisCube,Mapping"
   InputPosition  = (IsisCube, Mapping)
   OutputName     = cart:straight_vertical_longitude_from_pole
   OutputPosition = (Product_Observational, Observation_Area, Discipline_Area, cart:Cartography, cart:Spatial_Reference_Information, cart:Horizontal_Coordinate_System_Definition, cart:Planar, cart:Map_Projection, cart:Polar_Stereographic)
-  Translation    = (180, 90.0)
-  Translation    = (0, -90.0)
+  Translation    = (*, *)
 End_Group
 
 End


isis/appdata/translations/pds4ExportPolarStereographic.trn
```diff
@@ -248,13 +248,12 @@ End_Group
...
   Auto
-  InputKey       = CenterLatitude
+  InputKey       = CenterLongitude
   InputGroup     = "IsisCube,Mapping"
   InputPosition  = (IsisCube, Mapping)
   OutputName     = cart:straight_vertical_longitude_from_pole
   OutputPosition = (Product_Observational, Observation_Area, Discipline_Area, cart:Cartography, cart:Spatial_Reference_Information, cart:Horizontal_Coordinate_System_Definition, cart:Planar, cart:Map_Projection, cart:Polar_Stereographic)
-  Translation    = (180, 90.0)
-  Translation    = (0, -90.0)
+  Translation    = (*, *)
```
Kelvinrr, line 256: Sorry, can you explain this change? I personally don't know what the double-asterisk does.

ladoramkershner (approved):

isis/appdata/translations/pds4ExportPolarStereographic.trn
ladoramkershner: I am believe it just takes whatever was passed in and passes that same value out. 

I am a little curious why there is no longer a need for a longitude domain reconciliation, but I am guessing that since a Polar Stereographic projection contains all longitudes, as long as the user is consistent, it will resolve. I assumed this and (possibly prematurely) approved the changes. If you could verify or correct me, that would be great!

thareUSGS: PDS4 Cartography dictionary (thus PDS4 labels) doesn't have the concept of Longitude Domain. For a map projection, it is "mathematically" unnecessary. Although if we feel it is necessary it can be added. https://github.com/pds-data-dictionaries/ldd-cart

ladoramkershner: I am not sure I have enough knowledge to know if it is necessary in this context. However, If the PDS4 standard is set up without it and it is not mathematically necessary, I think it is fine. @Kelvinrr, do you disagree? If not, I am good to merge!

Kelvinrr: Seems good to merge, thanks for the explanations!

PR #3903, opened 2020-06-05
Isis2ascii Header Update

### LINKED ISSUE #3862: isis2ascii should just use spaces as the delimiter, even in the header
**Description**  

Replace the colons (":") in the header output by isis2ascii.  
**Example**  

Reading in a space-delimited ascii file is pretty standard.  But the colons in the header portion that isis2ascii writes are not usually recognized by default as delimiters.  While one can find a number of solutions to this, it seems there is no reason to make the user have to jump through these hoops before using the output from isis2ascii.

acpaquette:
Updated the isis2ascii header to be more csv complient when using space delimited parsing.

## Description
Changed the general format of isis2ascii to use spaces for all delimiters in the header over the previous format that used semi colons. This makes the header easier to parse when processing the output of isis2ascii using a space delimited ascii reader.

## Related Issue
Closes #3862

## Motivation and Context
Simplifies the parsing the isis2ascii output to only need a single reader that is using spaces as a delimiter.

## How Has This Been Tested?
Tested with the existing tests in ISIS

## Types of changes

## Checklist:

## Licensing


jessemapel (commented):
Can we get the tests on this converted to gtests? Fix looks good.

krlberry:
Agreed with Jesse that the changes look good! I also think that it's a good change, but does this sort of count as API-breaking? If anyone was using this in a script and already parsing the old format, it would break the script. (If there was already discussion about this that I missed, please disregard the question.) If nothing else, we should be sure to mention this in the release notes!

acpaquette:
@kberryUSGS That's something I thought about and wasn't 100% sure on either. It may break some pipelines but doesn't change any function signatures or how isis2ascii is actually called. Is a file format something we consider to be part of the API?

jessemapel:
@kberryUSGS The old header is such a strange format that I wouldn't even count it as something that can be parsed in a script. If this touched the non-header data, then I would consider it breaking.

krlberry:
@jessemapel Yeah, it's definitely a strange format. I'm okay with calling this non-API-breaking, then! (but still mentioning it in the release notes)

isis/tests/FunctionalTestsIsis2ascii.cpp
```diff
@@ -0,0 +1,130 @@
...
+      else if (pixelValue == 14) {
+        line[i] = Isis::HIGH_INSTR_SAT8;
+      }
+      else {
+        line[i] = (double) pixelValue;
+      }
+      pixelValue++;
+    }
+    testCube->write(line);
+  }
```
jessemapel: It may be worth moving this into the SmallCube fixture, or subclassing it to have a small cube with special pixel values in it.

jessemapel (dismissed):
Looks good. Moving the special pixel writing to the fixture seems like it could make some other tests easier.

jessemapel (approved):

jessemapel:
@acpaquette I completely forgot, can you put in a new PR with a history comment in ascii2isis.xml?

PR #3717, opened 2020-02-28
Adds the ability to save and restore a greyscale stretch to/from a Cube

krlberry:
## Description
This adds the ability to save and restore (and delete) a greyscale stretch from a cube using the Advanced Stretch Tool.

At this time, these three added functions (save/restore/delete) are accessed via buttons that launch dialog boxes from the Advanced Stretch Tool. The eventual goal is to have a single, more integrated interface for these actions, and for RGB stretches to be supported, in addition to greyscale.

This is what the new Stretch objects look like in an ISIS label when written to an ISIS Cube.

```
Object = Stretch
  Name        = sawtoothTrial1
  StartByte   = 30045920
  Bytes       = 178
  StretchType = SawtoothStretch
  Color = Greyscale
End_Object

```

## Related Issue
Requested by LROC (Does issue exist?)

## Motivation and Context
Requested by LROC (Does issue exist?) 
Adds the ability to save and restore stretches from the cube so they don't need to be saved and re-loaded by-hand. Previously, it was possible to save stretch pairs to a file, but it was not possible to reload them using the Advanced Stretch Tool. 

## How Has This Been Tested?
This has been tested by me only by interacting with a test cube in the Advanced Stretch Tool. This code is not covered by CI tests. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

[...]

PR #3906, opened 2020-06-09
Speed up q-isis applications by decreasing the number of times CubeViewport::paintPixmap() gets a new stretch

### LINKED ISSUE #3854: Qview Slowdown
**ISIS version(s) affected**: 4 and up

**Description**  
There is a noticeable slowdown when opening cubes in qview in versions of ISIS4 vs ISIS3, specifically when look at certain datasets (Notably Galileo data).

**How to reproduce**  
Grab these files:
/work/projects/EuropaBasemap/lweller/Galileo/FindFeatures/Island1/FindFeatures_ImprovedSpice/Observations/19ESREGMAP01/*cub (uncontrolled an controlled mosaic)

/work/projects/EuropaBasemap/lweller/Galileo/FindFeatures/Island1/FindFeatures_ImprovedSpice/Observations/19ESREGMAP01/12ESPWYCOL01_Island1V3.lis

While I was sitting in this directory, I ran qview in the following way from the command line:

qview *cub (to pop open the mosaics)
qview 12ESPWYCOL01_Island1V3.lis ( to pop open the level 1’s - included full path)
I typically open images from the command line when working with qview. I don’t know that the mosaics took any longer to load, but the level 1’s definitely did. Turns out these lev1’s are updated, but I’m not sure it matters. There is a list of spiced files (12ESPWYCOL01_Island1.lis) that opens just as slowly under isis4. Possibly even slower than the updated images.

**Possible Solution**  
¯\\\_(ツ)\_/¯

**Additional context**  
[Astro Discuss Post](https://astrodiscuss.usgs.gov/t/q-app-problems-under-isis4-1-0-rc2/397/5)

There has also been additional testing with Themis IR data and MRO CTX data, both do not seem to exhibit this problem. Both data sets only exhibit slight, insignificant changes in load times, maybe a few seconds difference when comparing ISIS versions.

krlberry:
## Description
I updated the `CubeViewport::paintPixmap()` method to move several `getStretch()` calls out of a for loop. The stretch values do not actually change during the loop, so there is no reason to call `getStretch()` over and over again. 

Prior to this change, `getStretch()` was being called over 900,000 times for a single test image! 

Each time `getStretch()` is called, the copy constructor for `Stretch` would be called. After PR #3717 which added the ability to save stretches to the cube, the copy constructor for Blob would then also be called 900,000 times for a single image. The source of the slowdown appears to have been "the slight increase in time required to also call the Blob copy constructor" times 900,000.

## Related Issue
Fixes #3854 (by fixing a long-pre-existing problem.)

## Motivation and Context
See related issue.

## How Has This Been Tested?
Locally, the slowdown is gone, and this is in fact faster than ever. 

A build containing these changes was tested by @lwellerastro.

I also checked to make sure that for a cubelist of input cubes that the stretch never changed inside the inner for-loop.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/qisis/objs/CubeViewport/CubeViewport.cpp
+++ b/isis/src/qisis/objs/CubeViewport/CubeViewport.cpp
@@ -1269,10 +1269,14 @@ namespace Isis {
 
           QRgb *rgb = (QRgb *) p_image->scanLine(y);
 
+          Stretch redStretch = p_red.getStretch();
+          Stretch greenStretch = p_green.getStretch();
+          Stretch blueStretch = p_blue.getStretch();
+
           for(int x = dataArea.left(); x <= dataArea.right(); x++) {
-            int redPix = (int)(p_red.getStretch().Map(redLine[ x - p_redBuffer->bufferXYRect().left()]) + 0.5);
-            int greenPix = (int)(p_green.getStretch().Map(greenLine[ x - p_greenBuffer->bufferXYRect().left()]) + 0.5);
-            int bluePix = (int)(p_blue.getStretch().Map(blueLine[ x - p_blueBuffer->bufferXYRect().left()]) + 0.5);
+            int redPix = (int)(redStretch.Map(redLine[ x - p_redBuffer->bufferXYRect().left()]) + 0.5);
+            int greenPix = (int)(greenStretch.Map(greenLine[ x - p_greenBuffer->bufferXYRect().left()]) + 0.5);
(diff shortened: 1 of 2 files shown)


acpaquette:
Seems good, I'm somewhat surprised this wasn't optimized when complied. This is probably good to merge once we hear back from @lwellerastro

jessemapel (dismissed):
Surprisingly simple optimization :+1:

jessemapel (approved):

acpaquette (approved):

PR #3909, opened 2020-06-09
Updated logic so that min/max are not calculated if values are provided

### LINKED ISSUE #3881: hist calculated min/max even if the user specifies the values
4.0.1 (and probably earlier versions, but I didn't actually go back and check)
**Description**  

The _hist_ application takes the time to compute minimum and maximum values "for the histogram" even when the user specifies the values to use.  This is a waste of time.  Quite a significant amount of time for large files.  This seems to be more like a bug than an enhancement request to me, but its a bit in the grey area, so feel free to move this ticket if you want.  
**How to reproduce**  

hist from=[any.cub] minimum=0 maximum=1.0 nbins=99
if you do this with the GUI, it is very helpful in showing what the application is busy doing.  Of course, you want to use a larger file to see the effect.  If you are having problems finding an example, you can always try a lunar or Mars global DEM.  
**Possible Solution**  

Accept the user values instead of computing them. 
**Additional context**

AustinSanders:
## Description
Added conditional logic such that min/max values are no longer calculated if the user passed in a min/max value on the command line or through the UI.

## Related Issue
Fixes #3881 

## Motivation and Context
Previously, the hist application would calculate the min, max, and number of bins whether or not the user entered values.  After calculation, the values would be overwritten if user-specified values were found.  The unnecessary calculations led to significant slowdown, which decreased usability.

## How Has This Been Tested?
Passed hist app tests.  Additional tests may be needed, but none were added due to the fact that no inputs/outputs should change as a result of this update.  This change should cut out extraneous calculations and nothing more.  I wasn't really sure how to programmatically test that the other steps were cut out.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/hist/main.cpp
+++ b/isis/src/base/apps/hist/main.cpp
@@ -27,16 +27,38 @@ void IsisMain() {
     QString msg = "The [TO] parameter must be entered";
     throw IException(IException::User, msg, _FILEINFO_);
   }
-  
-  // Setup the histogram
-  Histogram hist(*icube, 1, p.Progress() );
-  if(ui.WasEntered("MINIMUM") ) {
-    hist.SetValidRange(ui.GetDouble("MINIMUM"),ui.GetDouble("MAXIMUM") );     
-  }
 
-  if(ui.WasEntered("NBINS")) {
-    hist.SetBins(ui.GetInteger("NBINS"));
+  Histogram *hist;
+  if (ui.WasEntered("MINIMUM") && ui.WasEntered("MAXIMUM")){
+    int nbins = 0;
+
+    if (ui.WasEntered("NBINS")){
+      nbins = ui.GetInteger("NBINS");
+    }
+    else {
+      // Calculate bins based on data type.
+      // Bin calculations are based on logic in Histogram::InitializeFromCube
+      if (icube->pixelType() == UnsignedByte) {
+        nbins = 256;
+      }
+      else if (icube->pixelType() == SignedWord ||
+               icube->pixelType() == UnsignedWord ||
+               icube->pixelType() == UnsignedInteger ||
+               icube->pixelType() == SignedInteger ||
+               icube->pixelType() == Real) {
+        nbins = 65536;
+      }
+      else {
+        IString msg = "Unsupported pixel type";
+        throw IException(IException::Programmer, msg, _FILEINFO_);
+      }
+    }
+    hist = new Histogram(ui.GetDouble("MINIMUM"), ui.GetDouble("MAXIMUM"), nbins);
   }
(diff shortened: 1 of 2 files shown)


isis/src/base/apps/hist/main.cpp
```diff
@@ -27,16 +27,38 @@ void IsisMain() {
...
+      }
+      else {
+        IString msg = "Unsupported pixel type";
+        throw IException(IException::Programmer, msg, _FILEINFO_);
+      }
+    }
+    hist = new Histogram(ui.GetDouble("MINIMUM"), ui.GetDouble("MAXIMUM"), nbins);
   }
+  else {
+    hist = new Histogram(*icube, 1, p.Progress());
```
jessemapel, line 59: The Histogram needs to be deleted at the end of the function otherwise it's a memory leak.

isis/src/base/apps/hist/main.cpp
```diff
@@ -27,16 +27,38 @@ void IsisMain() {
...
-  // Setup the histogram
-  Histogram hist(*icube, 1, p.Progress() );
-  if(ui.WasEntered("MINIMUM") ) {
-    hist.SetValidRange(ui.GetDouble("MINIMUM"),ui.GetDouble("MAXIMUM") );     
-  }
 
-  if(ui.WasEntered("NBINS")) {
-    hist.SetBins(ui.GetInteger("NBINS"));
+  Histogram *hist;
+  if (ui.WasEntered("MINIMUM") && ui.WasEntered("MAXIMUM")){
```
scsides, line 32: Any reason this was pulled out of the Histogram class and not just made into a public member function in Histogram? Duplicate code.

AustinSanders: This calculation takes place before the constructor is called, so there's no object on which to call a member function.  The three ways that I considered handling this were:

1. Create a static member function
2. Call the constructor with an arbitrary number of bins, perform a calculation, then overwrite.
3. Pull the relevant parts of the logic into the file.

I avoided creating a static member function because I didn't feel like it was strictly necessary.  It felt weird to initialize an object with the incorrect number of bins just so that we could use a member function to calculate the right number of bins, so I just pulled the relevant parts of the logic into the file.

I do concede that this is partially duplicate code, but the original logic contains more steps than are necessary in that it calculates min/max.  This also leads to the evaluation of additional branches.

If anyone feels strongly about adopting a different means of implementing the changes, I can change it.

jessemapel (approved):
Duplicating the code is okay here. Histrogram gets used in enough places that a little duplicate code is a lower cost than potentially breaking other things.

kaitlyndlee (approved):

PR #3910, opened 2020-06-10
isis2ascii spacing fix

### LINKED ISSUE #2699: isis2ascii produces overlapping output in some cases
---

Author Name: **Kristin Berry** (Kristin Berry)

---

isis2ascii doesn't put space between the two adjacent output DN values in the case where we have negative output values in scientific notation, like so: 

0.0006183399-4.880164e-05

(Real example.)

This makes it difficult to actually read the isis2ascii output into another program.

I suspect this problem was caused by my own earlier ticket, but we should be able to output values that this precision, but with the whitespace needed: https://isis.astrogeology.usgs.gov/fixit/issues/2098

I would be happy to provide test data if it would be useful.

### LINKED ISSUE #3859: isis2ascii no space between values if a value is negative
4.0.0 and 3.9.x

**Description**  

isis2ascii uses spaces to delimit between output values.  there is no space between successive output real numbers if the number is negative

**How to reproduce**  

get a .cub with negative DN values.
run isis2ascii
try to ingest the output as a space-delimeted file
get fatal error message

**Possible Solution**  

**Additional context**  

Why would there be negative DN values in a .cub?  That is inconcievable!
image

acpaquette:
Increased the width of ascii DN output from 13 spaces to 14 spaces.

## Description
Sufficiently long negative values were using all 13 spaces from the formatting preventing certain applications from ingesting the output from isis2ascii. Adding this space should prevent this from happening but will place two spaces in front of any non-negative values.

## Related Issue
Closes #3859 

## Motivation and Context
The output from isis2ascii needs to be ingested by other applications using a csv reader. This was preventing some data from being read in through those other applications.

## How Has This Been Tested?
Currently not testing for this exact problem.

## Types of changes

## Checklist:

## Licensing


krlberry:
Will 1 space in front of negative values vs. 2 spaces in front of non-negative values still work with what most people are using to parse these? (I don't know either way whether or not space-delimited file readers tend to handle variable amounts of spacing well.) 

Assuming the above is true, code changes look good!

rbeyer:
@kberryUSGS it shouldn't.  As a "consumer" if you are told a line is space separated, most processes can handle grouping runs of whitespace together and assuming it is a single 'divider'.  In fact, since this code already writes out multiple whitespace characters if the value is less than 13 characters wide, that's probably fine.

However, I'm not sure that this fix is actually addressing the problem.  This code is writing out fixed-width values, but not setting any other formatting qualities on the stream, nor enforcing that there is a space between the values (it wasn't before, either).  Just making the fixed-width size larger may not fix it.  There is nothing in this patch that enforces a space between the elements of the Buffer `in`, only a widening of the field with the hope that whatever value that the `<<` operator applied to `in[i]` produces will be 13 characters or less, so that fill characters (which are spaces by default, get used to pad this entry to the specified width).

I think a better solution here would be to explicitly inject a space (or even a user-selectable character, so they could choose a comma, maybe?) between elements of `in`.  This way, even if those values used up all 13 or 14 characters, there would always be a space, because this code would explicitly write it.

Just in case someone was actually using fixed-width parsing of one of these files (shudder), I'd reduce width to 12 and inject a space.  This should keep lines in the output file the same length as before, but now some of the values might have the very last digit of their precision chopped off.

rbeyer (dismissed):

jessemapel (commented):
You should add a delimiter test for the new argument. Other than that, this looks good to go.

krlberry (dismissed):

rbeyer (approved):

jessemapel (approved):

PR #3927, opened 2020-06-17
Added new startup scripts that don't blast existing ISIS environment …

scsides:
New startup scripts that are not destructive to an existing environment

## Description

Added new startup scripts that don't overwrite the existing ISIS environment variables. If the variable is already set, the new scripts don't change it.

## Motivation and Context

Developing outside USGS network using the existing scripts is painful. These can be used instead of the ISIS3Startup.XXXX versions to work better.

## How Has This Been Tested?

The scripts were run locally, and the environment was examined to see that only unset variables were modified.

## Types of changes

## Checklist:

- [X} I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)

## Licensing

### DIFF
--- a/isis/scripts/isisStartup.csh
+++ b/isis/scripts/isisStartup.csh
@@ -13,7 +13,8 @@
 ################################################################################
 
 if ($?ISISROOT == 0) then
-  setenv ISISROOT /usgs/pkgs/isis3/install
+  echo "ISISROOT environment variable is not set"
+  exit 1
 endif
 
 if ($?ISISDATA == 0) then
--- a/isis/scripts/isisStartup.csh
+++ b/isis/scripts/isisStartup.csh
@@ -1,0 +1,37 @@
+#!/bin/csh
+################################################################################
+# This file should be sourced within your current shell using the "source"
+# command.
+#
+# On the command line type:
+# > setenv ISISROOT ????
+# > source isis3Startup.csh
+#
+# Replace the "????" in the above command line with the path you installed
+# the Isis distribution
+#
+################################################################################
+
+if ($?ISISROOT == 0) then
+  setenv ISISROOT /usgs/pkgs/isis3/install
+endif
+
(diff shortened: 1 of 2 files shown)


isis/scripts/isisStartup.csh
```diff
@@ -0,0 +1,37 @@
...
+# > setenv ISISROOT ????
+# > source isis3Startup.csh
+#
+# Replace the "????" in the above command line with the path you installed
+# the Isis distribution
+#
+################################################################################
+
+if ($?ISISROOT == 0) then
+  setenv ISISROOT /usgs/pkgs/isis3/install
```
Kelvinrr: Do we still use this area? Should it use an anaconda package instead?

Kelvinrr (commented):
just a (probably dumb) question.

kaitlyndlee:
I see the differences between the these scripts and the ones that are currently in the scripts directory, but why do we need two separate pairs of scripts with very similar names? I feel like this could be confusing in the future. I might be missing the point, but why not change the logic and the name of the scripts that are already there? Or change the names of these scripts to something more descriptive or different from the current scripts. `isis3Startup` vs `isisStartup` doesn't help differentiate between them or why there are two separate ones.

isis/scripts/isisStartup.csh
jessemapel: Yeah, I would just return an error message/usage.

jessemapel:
Could we just axe the isis3Startup scripts?

scsides:
@kaitlyndlee, Agreed, but didn't want to blast the old scripts incase someone uses them, like the initIsis.sh script

jessemapel:
Could we add a deprecation warning and give people 6 months/2 releases notice on those scripts going away? I think this is going to be one of the semantic versioning shifts.

jessemapel (approved):

PR #3929, opened 2020-06-19
Hayabusa2 w1, w2 and telescopic tests

Kelvinrr:
## Description

Adds gtests for Hayabusa2's cameras. 

## Types of changes

## Checklist:

## Licensing


isis/tests/Hyabusa2OncCameraTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+  nk["BODY_CODE"] = 499;
+  nk["BODY_FRAME_CODE"] = 10014; 
+  PvlObject newNaifKeywords("NaifKeywords", nk);
+  naifKeywords = newNaifKeywords; 
+
+  QString fileName = testCube->fileName();
+  // need to remove old camera pointer 
+  delete testCube;
+  // This is now a Hayabusa cube
+  testCube = new Cube(fileName, "rw");
```
acpaquette: Is it possible to put the bulk of this into a fixture that resides solely in this file?

isis/tests/Hyabusa2OncCameraTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+  nk["INS-37110_ITRANSL"] = {0.0, 0.0, 76.923076923077};
+  nk["INS-37110_BORESIGHT_LINE"] = 490.5;
+  nk["INS-37110_BORESIGHT_SAMPLE"] = 512.5;
+  nk["INS-37110_OD_K"] = {1.014, 2.933e-07, -1.384e-13};
+  nk["BODY499_RADII"] = {3396.19, 3396.19, 3376.2};
+  nk["CLOCK_ET-37_33322515_COMPUTED"] = "8ed6ae8930f3bd41";
+  nk["BODY_CODE"] = 499;
+  nk["BODY_FRAME_CODE"] = 10014; 
+  PvlObject newNaifKeywords("NaifKeywords", nk);
+  naifKeywords = newNaifKeywords;
```
jessemapel: I am concerned about re-using a cube for a different sensor that is already spiceinit'd.

- This could be adding duplicate keywords and the camera model may or may not get the right one.
- There could be keywords that the camera model needs that aren't being added and the existing keywords are silently filling in for them.
- What if a camera model tests for the absence of a keyword?

isis/tests/Hyabusa2OncCameraTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+  testCamera(*testCube, 11.215766294778371, 257.19997053715304, 51.0, 42.0, 173.0, 21.0, 54.0, 149.0, 174.0, 155.0);
+}
+
+void testCamera(Cube &c,
+                double knownLat, double knownLon,
+                double s1, double l1, 
+                double s2, double l2, 
+                double s3, double l3, 
+                double s4, double l4) {
+  Hyb2OncCamera *cam = (Hyb2OncCamera *) CameraFactory::Create(c);
```
jessemapel: We're going to want to re-use this for any test. This should instead be a generic Camera pointer.

isis/tests/Hyabusa2OncCameraTests.cpp
```diff
@@ -0,0 +1,245 @@
...
+void testLineSamp(Camera *cam, double sample, double line) {
+  bool success = cam->SetImage(sample, line);
+
+  if (success) {
+    double lat = cam->UniversalLatitude();
+    double lon = cam->UniversalLongitude();
+    success = cam->SetUniversalGround(lat, lon);
+  }
+  else {
+    FAIL() << "Failed to set sample/line";
```
jessemapel: It'll be more helpful when this fails if it outputs the line/sample that it set to.

jessemapel (commented):
The test looks good, but I think we should either completely re-write the label or use a cube that doesn't have a camera model already.

isis/tests/Hyabusa2OncCameraTests.cpp
Kelvinrr: In a perfect world, we would be able to instantiate an arbitrary camera model with some test ephem data (e.g. looking at some non real target and simple pointing.position data etc.) without having to more or less go through the spice init process.  But since this isn't an option (I didn't notice anything in the ISIS API that would allow it), wouldn't these issues be caught in the tests? I specifically had to go through the camera model to see what the model needed while using a real cube Hayabusa-2 image for reference.

jessemapel: > I specifically had to go through the camera model to see what the model needed while using a real cube Hayabusa-2 image for reference.

This is the part I am concerned about. If the person writing the test messes up here, it can slip through. It's unlikely but we've had issues with things like this before.

Kelvinrr: Are you suggesting we just make it go through the spiceinit process? More specifically, make a fixture using real cubes and ephemeris data per instrument? I'm not sure of an alternative.

isis/tests/Hyabusa2OncCameraTests.cpp
```diff
@@ -0,0 +1,180 @@
...
+    success = cam->SetUniversalGround(lat, lon);
+  }
+  else {
+    FAIL() << "Failed to set sample/line (Line: " << line << ", Sample: " << sample << ")." ;
+  }
+
+  if (success) {
+    EXPECT_NEAR(sample, cam->Sample(), .001); 
+    EXPECT_NEAR(line, cam->Line(), .001);
+  }
```
kaitlyndlee, line 233: Do we need to add in an else if the camera fails to set universal ground, like you do above with set image?

isis/tests/Hyabusa2OncCameraTests.cpp
jessemapel: I'm not worried about the table data and stuff that gets attached during the spiceinit process. I'm worried about the stuff that gets propagated to the label during ingestion. All of the values in the Instrument group mainly.

jessemapel: Here's the ONC-T Instrument group from the ingestion app test. Everything in this group could be used by the camera model. It seems like we're using bad test data if the instrument group of the test cube only contains a few of these values *and* contains values from a completely different image/instrument. It's a question of how "black box" should the testing on this be. This test is written based entirely off the code, not the expected use.

```
  Group = Instrument
    SpacecraftName                  = HAYABUSA-2
    InstrumentId                    = ONC-T
    InstrumentName                  = "Optical Navigation Camera"
    OriginalTargetName              = Earth
    TargetName                      = Earth
    StartTime                       = 2015-12-04T04:09:08.547
    StopTime                        = 2015-12-04T04:09:08.555
    ExposureDuration                = 0.0082 <seconds>
    SpacecraftClockStartCount       = 1/1012716101
    SpacecraftClockStartCount       = 1/1012716133
    RawSpacecraftClockCount         = 0x3C5CD245 <1/32 sec>
    OnboardDataRecordedTime         = 3C5CE063
    Binning                         = 1
    SelectedImageAreaX1             = 1
    SelectedImageAreaY1             = 1
    SelectedImageAreaX2             = 1024
    SelectedImageAreaY2             = 1
    SelectedImageAreaX3             = 1
    SelectedImageAreaY3             = 1024
    SelectedImageAreaX4             = 1024
    SelectedImageAreaY4             = 1024
    SmearCorrection                 = ONBOARD
    OffsetCorrection                = N/A
    FlatCorrection                  = NON
    RadianceConversion              = NON
    PhotometricCorrection           = NON
    BandRegistration                = NON
    L2BFlatFileName                 = N/A
    L2BSystemEfficiencyFileName     = N/A
    L2CShapeModelFileName           = N/A
    L2DPhaseFunctionFileName        = N/A
    L2DShapeModelFileName           = N/A
    ImageID                         = 36086
    SubImageCount                   = 2
    BusLineVoltage                  = 49.28 <V>
    ONCCurrent                      = 0.51 <V>
    FLACCurrent                     = 0.00 <V>
    ONCAETemperature                = 0.37 <degC>
    ONCTOpticsTemperature           = 19.21 <degC>
    ONCTCCDTemperature              = -29.46 <degC>
    ONCTElectricCircuitTemperature  = -10.31 <degC>
    ONCW1OpticsTemperature          = 1.47 <degC>
    ONCW1CCDTemperature             = -25.23 <degC>
    ONCW1ElectricCircuitTemperature = -10.29 <degC>
    ONCW2OpticsTemperature          = 1.11 <degC>
    ONCW2CCDTemperature             = -25.04 <degC>
    ONCW2ElectricCircuitTemperature = -6.16 <degC>
    FLACTemperature                 = -13.40 <degC>
    Compression                     = LOSSLESS
  End_Group
```

isis/tests/Hyabusa2OncCameraTests.cpp
acpaquette: This doesn't look like it was addressed unless I'm missing something?

acpaquette (commented):
Overall this looks pretty clean and ready to go! Small comment regarding a resolved comment from @jessemapel

scsides (commented):
Once Adam's comment is addressed, go ahead and merge it

isis/tests/Hyabusa2OncCameraTests.cpp
Kelvinrr: Originally didn't because it was using the shutter times function which I though was specific to hyb2onc camera, switched it to a FramingCamera pointer, so later we can at least re-use it for other framers.

PR #3935, opened 2020-06-25
Update hist application to accpet nbins when min/max are not set

krlberry:
## Description
PR #3909 introduced a bug in which `nbins` set by the user would be ignored if `minimum` and `maximum` were not also set. This PR fixes this issue and adds a test for `nbins`. (I did not migrate the tests as part of this PR.) 

## Related Issue
See PR #3909. Issue was uncovered by CI, so no ticket number. 

## Motivation and Context
Fixes failing test in CI. 

## How Has This Been Tested?
Existing hist tests pass locally and the `default` test for `hist` was updated to include a run which used `nbins`.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/hist/tsts/default/Makefile
+++ b/isis/src/base/apps/hist/tsts/default/Makefile
@@ -1,9 +1,20 @@
+# 2020-06-25 - Kristin Berry - Added a test to check that if nbins is set, it is used.
+#   To make this test, I set the number of bins to much smaller number than the internal default,
+#   and just checked that the output was different from the default test. 
+
 APPNAME = hist
 histTruth.txt.IGNORELINES = Cube
+histTruthNbins.txt.IGNORELINES = Cube
 TEMP = $(OUTPUT)/histTruth.txt
+TEMPNBINS = $(OUTPUT)/histTruthNbins.txt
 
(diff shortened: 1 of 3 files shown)


isis/src/base/apps/hist/tsts/default/Makefile
```diff
@@ -1,9 +1,20 @@
+# 2020-06-25 - Kristin Berry - Added a test to check that if nbins is set, it is used.
+#   To make this test, I set the number of bins to much smaller number than the internal default,
+#   and just checked that the output was different from the default test.
```
ladoramkershner: I think this is me not being good at reading Cpp, but where does the comparison to the default test happen? I see both of the results going to dev/null

krlberry: You're right, it's not doing this comparison! I'll update this comment (tomorrow.) My intent behind the comment was to explain that I didn't _really_ verify the truthdata for the test I added. I just verified that it was _different_ than the truthdata without `nbins` set.

kaitlyndlee (dismissed):
Looks good after the testing change gets pushed.

kaitlyndlee (approved):

PR #3942, opened 2020-07-07
Updates SpiceRotation to use ale::Orientations

krlberry:
## Description
Updates SpiceRotation to use ale::Orientations. 

Notably only uses Orientations to to calculate/store `p_CJ` (time-dependent rotation) and `p_av` (angular velocity) from SpiceRotation. Since the constant and non-constant frames and constant rotation are also part of the ale::Orientations object, it would have been possible to store those in the m_orientations object, too, but I didn't want to be creating new Orientations objects constantly to update/add a list of time-dependent frames, etc... 

Some tests will fail because this requires a new version of ale to run. 

## Related Issue
Ale work, no ticket.

## Motivation and Context
Use one library (ale) to do calculations for both ISIS and USGSCSM. 

## How Has This Been Tested?
SpiceRotation unit test matches, except for some very small numbers that are basically 0 vs. actual 0 in the output.

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/SpiceRotation/SpiceRotation.cpp
```diff
@@ -373,12 +384,29 @@ namespace Isis {
     int cacheSize = p_cacheTime.size();
 
     // Loop and load the cache
+    std::vector<ale::Rotation> rotationCache;
+    std::vector< std::vector<double> > cache;
```
jessemapel, line 388: Why is this cache here? it doesn't look like it gets used.

isis/src/base/objs/SpiceRotation/SpiceRotation.cpp
```diff
@@ -1222,15 +1303,32 @@ namespace Isis {
    * Set the rotation angles (phi, delta, and w) for the current time to define the
    * time-based matrix CJ. This method was created for unitTests and should not
    * be used otherwise.  It only works for cached data with a cache size of 1.
-   *
+   *  
+   * // FIXME - MAYBE THIS SHOULD BE IN THE UNITTEST THEN.... (see comment)
```
jessemapel, line 1307: Why is this here?

krlberry: Thanks for catching this. This is a note I made to myself that I didn't end up changing. There's a comment in this function that it was only added for unit testing and shouldn't be used for anything else... so I thought it might make sense to just subclass SpiceRotation, add this function, and use that for the unit test. Then I realized that this is going to be gtest converted at some point anyway, and going to the extra effort of moving this probably wasn't worth it.

krlberry: So, I'm just going to delete the comment.

jessemapel:
Closed for now. Will re-open once we get more work on this.

Issue #3810, opened 2020-03-27
How to cite ISIS in my paper/abstract?


rbeyer:
**Description**
What is the preferred way to cite my use of ISIS in papers and abstracts that I publish?

**Background and Discussion**
I cannot seem to find an answer in the [FAQ](https://github.com/USGS-Astrogeology/ISIS3/wiki/FAQ) or in the repo.

The only "official" answer appears to be [this document](https://isis.astrogeology.usgs.gov/documents/References/References.html), last updated in 2013.

In my opinion, setting up a DOI-minting process through [Zenodo](https://zenodo.org/) for the ISIS software would be ideal, and would provide an automatic, consistent mechanism for a new DOI for each tagged release, and would also provide a citation style that academics can easily use and incorporate into their work, like any other reference.  We have this set up for ASP, for example.

The reference for our most recent release is:

> Ross Beyer, Oleg Alexandrov, & ScottMcMichael. (2019). NeoGeographyToolkit/StereoPipeline: ASP 2.6.2 (Version v2.6.2). Zenodo. http://doi.org/10.5281/zenodo.3247734

And there is also a DOI that will always point to the 'latest' version, whatever that is: https://doi.org/10.5281/zenodo.598174

In the meantime, I guess I'll continue to cobble together a reference that points to some URL and maybe the most recent LPSC abstract I can find?

Thanks!

thareUSGS:
Funny I was just mentioning this to the team this morning. I understand there is one in the works (perhaps pushed into this [special issue](https://www.mdpi.com/journal/remotesensing/special_issues/planetary_3D_mapping_remote_sensing_machine_learning) - sorry self-plug) Besides the plethora of abstracts, the reference I normally see is:

Edmundson, K. L., Cook, D. A., Thomas, O. H., Archinal, B. A., and Kirk, R. L. (2012) Jigsaw: The ISIS3 bundle adjustment for extraterrestrial photogrammetry, International Annals of Photogrammetry, Remote Sensing, and Spatial Information Sciences, I-4, 203-208. doi:10.5194/isprsannals-I-4-203-2012, URL: [https://doi.org/10.5194/isprsannals-I-4-203-2012](https://doi.org/10.5194/isprsannals-I-4-203-2012)

or direct like this ( but I would also **prefer the Zenodo** idea @jlaura )
[...]

PR #3950, opened 2020-07-13
Adds a draft .zenodo file to get authors names

jlaura:
Adds a .zenodo.json file.
## Description

This addition is in order to get a listing of the authors in some defined order. This is just a draft because I am honestly not sure how best to order these. What I have done (that I do not like):

- List the project lead
- List the USGS folks by number of commits (I'm lazy and pulled this from git)
- List everyone else by number of commits on git
- List anyone not on GitHub but who made the Authors.rst list as generated by @AndrewAnnex 

## Related Issue

#3810 

## Motivation and Context

We need to have an accurate author ordering for citations.

## How Has This Been Tested?

It honestly hasn't and I am not sure how best to test this with Zenodo. The JSON is valid and I see that Zenodo has `sandbox.zenodo.org`, but I am not sure how to use the sandbox for testing.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,198 @@
+{
+  "creators": [
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Laura, Jason",
+      "orcid": ""
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Mapel, Jesse"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Berry, Kristin"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Lee, Kaitlyn"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Rodriguez, Kelvin"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Sides, Stuart",
+      "orcid": ""
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Paquette, Adam"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Williams, Kaj"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Sanders, Austin"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Adoram-Kershner, Lauren"
+    },
+    {
+      "affiliation": "United States Geological Survey, Astro Geology Science Center",
+      "name": "Shinaman, John"
+    },
+    {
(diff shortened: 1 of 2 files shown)


rbeyer:
So this `.zenodo.json` file is going to be manually managed, or do you have a script that does the deed based on the `AUTHORS` file, and the rules you described above?  If so, then authors would just have to worry about the `AUTHORS` file, and only those who are involved in release engineering have to deal with the .zenodo.json file.  Then the `.zenodo.json` file is a compiled thing based on info in AUTHORS and GitHub stats, which would free us from having to content track two different files that track the same info.

Of course, the thing that occurred to me before was that the .zenodo.json file allows for 'affiliation' and 'orcid' fields, so maybe the `.zenodo.json` file should be the canonical source of "author" information and maybe we create the human-readable `AUTHORS` file from it?  And then I realized I was in a loop, and ran away from the topic screaming.  Seems like we should do one or the other, but I honestly don't know which.

Did you ever find a way to test this .zenodo.json file?  When I looked into this, they didn't appear to have a linter or any reasonable way to test the contents of the .zenodo.json without cutting a GitHub release and just seeing what happens.

jlaura:
@rbeyer I definitely punted with the update both approach because I also am not sure which is 'of record'. I think the .zenodo.json file should be, but it is so alpha that I'm not brave enough to assert that.

I did not see any method for linting these. I saw mention of the sandbox URL in an issue on the zenodo GitHub, but the documentation for doing this is rough. I initially thought this was going to have to be 100% scripted through their REST API, but it looks like zenodo will magically parse this file....maybe? I am not super confident in this hence the first paragraph above.

So my 1/2 way solution for now is to do both and then, if zenodo looks like it is working, fully automate that (maybe with a GitHub action on a PR assuming we can get a post merge action to alter a file in the repo; this seems insanely dangerous for a code file by the way?) and then go from there. 🤷‍♂️

rbeyer (approved):

jessemapel:
I am in favor of just keeping one thing around. It would be easier to keep the zenodo file around and then generate the AUTHORS file from it. It would be straight forward to have CMAKE run a small Python script to do that and then install it wherever we want in the distribution.

I think we're going to have to hand-maintain the zenodo file because it contains things like affiliation and ORCID that are not available programmatically.

For order, I do not like number of commits for two reasons:

1. It's a bogus metric that doesn't mean anything. It also biases towards git commits over SVN commits because of how the two tools differ.
1. The order of the file will change. If we're hand maintaining the file, then keeping the order of the file fixed would be greatly preferable. Having to re-order the file every time we pull a new release would suck.

I would rather the order be project lead first and then everyone else alphabetically.

jlaura:
I'm all for lead and then alphabetical. I think this is on the TC agenda today, so if that sounds good for everyone I will get the list alphabetized. I can have this PR updated with a script to generate the authors.txt by the end of the week.

jessemapel:
@jlaura Where is this at? Generating a script for this/working on the authors/zenodo management is a good candidate for the process work this week.

jlaura:
I have not made it to do the generation, so if someone wants to pick that up, that would be ideal

jessemapel:
I am going to merge this PR and open an issue on the repo to improve the zenodo and authors file management.

PR #3991, opened 2020-08-27
Update file paths for data moved into the $ISISTESTDATA area.

### LINKED ISSUE #3982: Move testdata from the ISISDATA area into ISISTESTDATA
**Description**  
Right now we have test data in two different locations: ISISDATA and ISISTESTDATA.  The test data in the ISISDATA area are stored inside of the individual mission directories, e.g., `$ISISDATA/apollo15/testData`. I am not sure why we have test data in two areas and, in an effort to get metadata generated for release, the load of to be developed metadata is non-trivially increased with the test data being included.

I would like to see `$ISISDATA/*/testData` moved into the `$ISISTESTDATA` directory. This unifies our test data in one place, gives us a better picture about the total size of the test data area, and reduces the amount of metadata to be generated in order to release the ISIS data area.

It also looks like this will remove the `control` directory entirely.

**Example**  

N/A

krlberry:
## Description
Most of the test data associated with unit tests in ISIS is located in `$ISISDATA/<mission name>/testData`, rather than inside the $ISISTESTDATA area. Internally, this data has now been copied over to `$ISISTESTDATA/isis/src/<mission name>/unitTestData`. The old files will be deleted from the $ISISDATA area after this PR is approved.

This PR itself updates anywhere in ISIS that accesses the (mostly) unit test-related test data to point to the new path.

## Related Issue
Fixes #3982 

## Motivation and Context
Test data was mixed in with data needed to run ISIS. Since the way we distribute $ISISDATA is going to change in the future, this is a step to move data that didn't need to be distributed for ISIS to function into a different area.

## How Has This Been Tested?
Local tests pass. Don't merge before Jenkins runs.

## Types of changes

## Checklist:

## Licensing


krlberry:
Other comments: 
(1) @scsides mentioned a preference for naming the new directories `testData` rather than `unitTestData`, but I'd just gotten this working and wanted to put the PR up to get feedback and test CI. Anyone else have any thoughts on `testData` vs `unitTestData`?
(2) Some of the examples on the ISIS website use test data from the `$ISISDATA/<mission name>/testData` areas as examples, presumably because users would have access to them. I just updated the paths, but since they're moving to the `$ISISTESTDATA` area, users will no longer have access to them. Is there another place we could make these available? We could create a true "examples" directory under $ISISDATA, which contains a very small number of example cubes and use that. Any other ideas?

jessemapel:
A few high level comments:

1. I don't have a strong preference for the file name. If there's a good reason for just using testData then that's okay with me.
1. Making an examples directory in the data area sounds like a good idea. How many of these example images are there and how big are they?
1. We should wait a little bit before deleting the unit test data from the data area so that folks working without this update don't have a ton of things fail suddenly. Maybe wait 1-2 weeks. That will leave hyabausa2 out in the cold a bit though, the branch will need to be updated from dev again.

krlberry:
There are only 3 files used as examples on the ISIS website which included paths to the $ISISDATA/mission/testData area: ab102401.cub, isisTruth.cub, and unitTest_ControlNetVersioner_ProtoNetwork1_ProtoV0001.net. Total size: 2.4 MB. @jlaura Would including two example cubes and one example control network in the $ISISDATA area (associated with the ISIS documentation) work okay for metadata generation, etc?

jlaura:
@kberryUSGS We have lots of cubes in the data area for stuff like calibration, so having cubes isn't a problem. All the test data would require associated metadata, making the volume of to be generated metadata that much larger. What I do need to know is what those cubes are, who generated them, and how they were generated. Honestly, I am going to need to know that plus some additional stuff for everything in the data area so we don't need to answer those questions just yet.

jessemapel:
@jlaura The cathist application will tell you who generated them, when, and how.

isis/src/lro/objs/LroWideAngleCamera/unitTest.cpp
```diff
@@ -33,7 +33,7 @@ using namespace Isis;
 
 // 2013-12-19 We will have new CK kernels with temperature dependence coming
 // that will affect the lat/lon values here.  Also must get a new FK, IK and
-// IAK.  Don;t forget to move the new test cube to $ISISROOT/lro/testData!!!!
+// IAK.  Don't forget to move the new test cube to $ISISROOT/lro/unitTestData!!!!
```
jessemapel: This doesn't look like the right path. Should it be $ISISTESTDATA/isis/src/lro/unitTestData

jessemapel (dismissed):
Looks good, just one small documentation comment, that looks like it's always been wrong? So, I'm not sure if it's worth updating or just merging this.

isis/src/lro/objs/LroWideAngleCamera/unitTest.cpp
krlberry: Good catch! I'll fix this.

jessemapel (approved):

PR #3995, opened 2020-09-01
Simple script to generate AUTHORS.rst from .zenodo.json file

krlberry:
Adds a simple script to the scripts directory to generate the AUTHORS.rst file from the .zenodo.json file, so new contributors only need to add their name and other information in one place. This is a follow on to https://github.com/USGS-Astrogeology/ISIS3/pull/3950.

## Description
Adds a script to generate the AUTHORS.rst file from the .zenodo.json file. The script is very simple.

Also updates the contributing document, pr template, and current AUTHORS.rst with the output of the script.

For now, I updated the [release docs](https://github.com/USGS-Astrogeology/ISIS3/wiki/Public-Release-Process#part-d-update-the-authors-list): to include running this script to update the `AUTHORS.rst` file as a step in a creating a new ISIS release, but I did look into using GitHub Actions to do this more automatically and https://github.com/marketplace/actions/github-push looks like a promising way to use github actions to update the `AUTHORS.rst` automatically on an event like a PR merge, if this gets worked on more in the future.

## Related Issue
#3810

## Motivation and Context
Adds a script so that only one file (the .zenodo.json) is the standard place for authors to have their names, affiliations, etc, and the other (AUTHORS.rst) can be generated from it.

## How Has This Been Tested?
Works on the current .zenodo.json.

## Types of changes

## Checklist:

## Licensing


rbeyer:
I don't think that this script should have default file paths based on the ISISROOT environment variable.  Someone could alter their actual, installed AUTHORS.rst file and not the file in the git repo they mean to change for commit.

Yeah, the json module is part of the Python standard library, you should use it.

jessemapel (commented):
👍 to using the json module to read the zenodo file. Also, I would use argparse to get the path to the zenodo file and/or author file. Plus, you can also add a help message and docs that way.

isis/scripts/zenodo_to_authors.py
```diff
@@ -0,0 +1,50 @@
...
+authors_path = args.authors_output_file
+
+# Files to open
+if (("$ISISROOT" in args.zenodo_input_file) or ("$ISISROOT" in args.authors_output_file)):
+  if ("ISISROOT" in os.environ):
+    zenodo_path = os.path.expandvars(args.zenodo_input_file)
+    authors_path = os.path.expandvars(args.authors_output_file)
+  else:
+    print("Please set $ISISROOT and re-run this script or specify the input and output files.")
+    exit()
```
jessemapel: I do not like setting the defaults and then having to do this; especially,  because this could error. My rule of thumb is that a script with default arguments should be able to just run; if I have to set environment variables to run it with default arguments I may as well just pass the arguments. I would just make the arguments required.

isis/scripts/zenodo_to_authors.py
```diff
@@ -0,0 +1,50 @@
...
+
+# Read .zenodo.json file:
+lines = []
```
jessemapel: This looks like it's no longer needed

isis/scripts/zenodo_to_authors.py
```diff
@@ -0,0 +1,50 @@
...
+try: 
+  with open(zenodo_path, 'r') as zenodo_file:
+    parsed_json = json.load(zenodo_file)
+except: 
+  print("{} could not be opened as the input file.".format(zenodo_path))
+  exit()
```
jessemapel: You should re-throw the exception here so that users know why is failed.

isis/scripts/zenodo_to_authors.py
```diff
@@ -0,0 +1,50 @@
...
+  name = elt['name']
+  last, first = name.split(",")
+  output_lines.append("- {} {}".format(first.strip(),last.strip()))
+
+output_lines.append("\n-----")
+output_lines.append("This list was generated from the .zenodo.json file by running python zenodo_to_authors.py.")
+
+# Write to AUTHORS.rst file
+with open(authors_path, 'w') as authors_file:
+   authors_file.write('\n'.join(output_lines))
```
jessemapel: You should add similar exception logic for writing the output file as you did for reading the input file.

isis/scripts/zenodo_to_authors.py
krlberry: Can do. My initial thought was that the default use of this script would be automated (in the future!) rather than normally being something an individual user would run, so I was thinking "$ISISROOT" is set would be the default case. I see your point, though, and am happy to just remove the defaults and make the arguments required.

isis/scripts/zenodo_to_authors.py
krlberry: Oops. Good catch.

jessemapel (approved):

Issue #3363, opened 2019-07-30
THEMIS and MRO SSD Kernels in ISIS


jlaura:
**ISIS version(s) affected**: 3.8.0

**Description**  
The smithed THEMIS kernels and MRO SSD kernels are currently not included in ISIS. These kernels need to be shipped to ISIS users.

**How to reproduce**  
Download smithed kernels, use them manually with spiceinit and be amazed by how much better good pointing really is.

**Possible Solution**  
copy and update kernels.db files

**Additional context**

krlberry:
@jlaura Where are the smithed THEMIS kernels located?

jlaura:
@rfergason @lwellerastro Can you all aim us at the kernels to be included? I have seen some data/kernels under /work/projects, but I want to make sure we get the correct kernels.

jessemapel:
For the MRO SPKs, they can be found [here](https://naif.jpl.nasa.gov/pub/naif/MRO/kernels/spk/). Anything with `_ssd_` in its name is from the JPL SSD group. We should prefer the `mro110c` kernels, but if they are not available, then we can use the `mro95a` kernels.

rfergason:
The final smithed CK and SPK kernels that Lynn produced from the THEMIS control network solution are located in:  /archive/projects/themis_control/Kernels

There are separate CK and SPK kernels for the dayIR images and nightIR images, as daytime and nighttime IR images were not bundled in the same solution.

krlberry:
The kernels are now in and updated. Leaving this issue open until all related ISIS tests are passing.

jlaura:
Is this safe to close? I didn't check / see the CRON jobs - looks like the working Jenkins stuff is still good to go.

krlberry:
Kernels were updated and tests are passing

lwellerastro:
Hi - how do I access the updated Themis kernels?  I am working with a test image under isis3.10.2 (and under isis4.2.0) and it doesn't look like I'm picking up smithed kernels.  This is an image that was controlled and is included in the updated kernels.

```
thm2isis from=/pds_san/PDS_Archive/Mars_Odyssey/THEMIS/USA_NASA_PDS_ODTSDP_100XX/ODTSDP_10014/data/odtir1_0014/i147xxrdr/I14705003RDR.QUB to=I14705003RDR_smithed.cub

spiceinit from=I14705003RDR_smithed.cub cksmithed=true spksmithed=true
```
From the label after spiceinit:

[...]

PR #3997, opened 2020-09-02
Adds an optional parameter to kerneldbgen to time-pad smithed kernels.

krlberry:
##  Description
Adds a new optional parameter to kerneldbgen (the ISIS application that generates kernel databases) to add time-padding to database entries for smithed kernels. This was added in addition to the global startoffset/endoffset parameters because for #3363, just the smithed kernel entries needed to have their start/stop times extended by a fraction of a second.

## Related Issue
#3363 

## Motivation and Context
Fixes issue where smithed kernels would not get picked up by spiceinit because of a fraction of a second time-offset from the start time of the image, even though they work with every pixel of an image. 

## How Has This Been Tested?
Updated `makedb` scripts for themis to use updated kerneldbgen, ran with 0.5 second padding added to the smithed themis SPKs and CKs, and now spiceinit will pick up the smithed kernels. 

Also used phocube and stats to make sure that pointing information was available for every pixel of the image.

## Types of changes

## Checklist:

## Licensing


isis/src/system/apps/kerneldbgen/kerneldbgen.xml
```diff
@@ -429,7 +432,21 @@
           <item>SMITHEDFILTER</item>
         </exclusions>
       </parameter>
-    </group>
+      <parameter name="SMITHEDPAD">
+        <type>double</type>
+        <brief>
+          Seconds to subtract from the start time and add to the stop time in each smithed kernel.
```
jessemapel: This is added to each segment of each kernel, not the entire start and stop time of the kernel.

jessemapel (commented):
Looks good, but needs a changelog entry

isis/src/system/apps/kerneldbgen/kerneldbgen.xml
krlberry: Thanks for catching this. I'll update the documentation.

jessemapel (approved):

jessemapel:
@kberryUSGS This can be merged once CI successfully builds.

scsides (changes requested):
Each smithed kernel filter pattern usually comes from a single source. I think it would be reasonable to have a different padding for each filter. It might also be reasonable to add the padding for reconstructed too

krlberry:
This didn't actually end up being useful for the THEMIS kernel problem I was working on. @scsides Do you think it's still worth me updating this and getting it merged? I was considering just closing this PR because I'm not sure it's actually useful.

scsides:
This would only be useful if there were different sources of smithed kernels that needed different padding. If we don't need it now, I suggest closing. We can always add the feature in later when it is needed.

jessemapel:
@scsides The other time I've run into this, I had to do what Kristin is doing here. It was for Cassini ISS images that all had radically different exposure times. I ended up writing a script to pad for each individual image.

PR #3922, opened 2020-06-15
Revert "Changed histogram binning logic"

jessemapel:
Reverts USGS-Astrogeology/ISIS3#3918

This is being reverted because it caused significant change in the app test outputs and more time it needed to check that it did not break anything. When, all of the changes are signed off on, it can be merged back in.

Issue #3882, opened 2020-05-26
hist application does not use the user specified nbin parameter correctly.


lavalaz:
**ISIS version(s) affected**: x.y.z
isis4.1.0 (and earlier)
**Description**  

I carefully picked the min/max values and NBINS to put the bin boundaries at very specific DN values. But the application did not honor those user input limits and chose a min value larger than the user specified min and a max value less than the user specified max.  Therefore the bin boundaries are not where the user intended.  
**How to reproduce**  

Pick any typical image with I/F values between 0 and 1.
_hist from=[input.cub] minimum=0.0 maximum=1.0 nbin=10_
This should but the bin boundaries at 0.0-0.1, 0.1-0.2, 0.2-0.3, 0.3-0.4, 0.4-0.5, 0.5-0.6, 0.6-0.7, 0.7-0.8, 0.8-0.9, 0.9-1.0.  Count them. Ten (10) bins.  Hist produced 10 bins with "DN" of 0,0.11111,0.222222,0.333333,0.444444...  ...0.88888,1.  Ten bins alright, but not at 0.1 spacing.  I guess I was wrong and 1/10=0.111111.  My bad.  So what DN's are in those bins?  The output only specifies one DN value and I can't tell if its the max or the min of the bin.  Since it starts with "0", clearly its the minimum value and the first bin has DN values of 0.0-0.111111.  But since the last bin is "1", it must have DN values of 1 to ???  Oh, must be my mistake, the DN value must be the for the top of the bin so the last bin is for DN values of 0.88888-1.0.  So the first bin has DNs of ??? to zero? But if I ask for 11 bins, I get bins of "0,0.1,0.2,0.3...1".  1/11=0.1, now I get it! 
**Possible Solution**  

Part 1: You need to give the values for minimum and maximum of the range of DNs that are going into each bin, not just a single DN value.  You should also clearly document which side of the bin is a greater than/less than versus greater-than-or-equal-to/less-than-or-equal-to.  
Part 2: You should also have a bin for "less-than-the-minimum" and "greater-than-the-maximum".  So 12, clearly labeled, bins when the user says nbin=10.  
**Additional context**  

Just for historical context, I know this issue was raised about 15 years ago.  The user in question, not myself, who had a degree in mathematics, was told that they did not understand the mathematics of histograms and this behavior was correct.  That is why you are seeing extra snarkiness in this ticket.

jessemapel:
[...]

PR #4004, opened 2020-09-08
Revert reverted histogram PR and handle conflicts.

krlberry:
This PR, along with updated tests will close out #3882. I couldn't revert the reverted PR using GitHub, so this PR does so.

## Description
The actual work for this PR was completed by @kaitlyndlee and @AustinSanders during the last support sprint. It needed to be reverted due to failing tests, but is now being re-submitted.

See previous PRs: #3918 and #3922 

## Related Issue
#3882 

## Motivation and Context
See #3882 or #3922. Summary from #3922: "Makes hist and cnethist behave in the way users expect them to."

## How Has This Been Tested?
Testing right now using Jenkins.

## Types of changes

## Checklist:

## Licensing


AustinSanders (dismissed):

kaitlyndlee (dismissed):

acpaquette:
Regarding these hist changes. The histogram tool used to use the minimum and maximum DNs of a cube as the middle of the first and last bin to represent the 0.5% - 99.5% range. The new changes place the minimum and maximum at the ends of the hist, meaning that the 0.5% - 99.5% ranges are now some value that is a little larger than the minimum DN and a little smaller than the maximum DN.

This directly conflicts with the histograms use as a stretch tool. When ProcessExport is outputing using Reals (32bit) numbers the max and min are determined through the histogram by getting the middle of the first and last bin. This conflicts with the behavior of the new histogram as the middle of the first and last bin are no longer your minimum and maximum DN. I think this can be changed by just setting the minimum and maximum through stats rather than asking the histogram for the minimum and maximum.

Consequently we would also be working with 0% - 100% of the data in the cube rather than the .5% - 99.5% dictated in some export apps like isis2raw, where minpercent is .5 and maxpercent is 99.5.

I think fixing the minimum and maximum issue currently present will help but there will be a few other things that need to be fixed along with it.

jessemapel:
I'm not sure if it's correct to do 0-100 stretches or 0.5 to 99.5 stretches. Hopefully some folks more familiar with SOP for this can chime in. I am inclined to reproduce the old behavior.

For reproducing the old data, we could switch to using Histogram::Percent instead of the min and max.

jlaura:
@thareUSGS @lavalaz @blandoplanet @kaitlyndlee   - I believe that you three commented on the original issue. Care to chime in on the comment from @acpaquette above?

acpaquette:
@jessemapel Unfortunately Histogram::Percent is what is causing this problem. It will only return the middle value of the bin. So if you ask for 0.0 or 0.5, you'll get the middle value of the bin that contains that percentage. I think this is the larger issue in Histogram class as the Percent method gives you the closet middle bin value and not the actual value at a given percentile

jessemapel:
I would like to request a "deeply saddened" emoji reaction for me to use on Adam's comment.

lavalaz:
For what its worth, I can explain what I, as a user, expect when I ask for a 0-100% stretch versus a 0.5-99.5% stretch.  0-100% stretch means that the minimum valid DN value and the maximum valid DN value of the input and the output are the same.  0.5-99.5% stretch means the highest 0.5% and lowest 0.5% of the DN values of the input get crammed into the top and bottom DN values of the output.  This is sometimes referred to as "clipping" the data.  It is common for image data to have a small fraction of the pixels with anomalously high and low values (e.g., because of noise in the detectors) and thus a small amount of clipping can help use the dynamic range of the DN values on scientifically useful features. But sometimes it is important to have no clipping (e.g., finding a few pixels of bright snow on dark sand).  So the difference can be very important to the user.

kaitlyndlee:
When I was originally working on the code, I wasn't sure why the bin intervals were being set they way they were. Now it makes more sense. I am not sure which is the best approach. From a person who does not use hist, the changes in this PR make sense to me, but a user may think differently. If we decide to revert again (lol), I think adding documentation to explain why the bins are being created in this way would be extremely helpful.

lavalaz:
I should clarify, that my explanation was for when there are no special pixels being generated.  Including pixel values for high and low saturation, 0-100% stretch means no pixels get converted to high or low saturation values.  But for 0.5-99.5% stretch, the bottom and top 0.5% of the pixels get converted to low and high saturation values.

acpaquette:
@lavalaz I think the changes that Austin and Kaitlyn made still facilitate these features. What you are asking for as a user is what I kinda figured the percentage selection was for but the histogram class hasn't been producing the right percentages for a while now. 

When you ask for output as Reals you can't set any invalid pixels without setting percentages. Unfortunately the Histogram class hasn't been actually trying to get detailed percentages. Rather it simply finds the bin that contains that percentage and returns the middle of that bin. 

So if your first bin contains the first 15% of data (which would be weird) and you ask for everything above the first 0.5% the class would return whatever the middle of that 15% is. So really you would chop the bottom 7.5% of your DNs out as special pixels rather than only the first 0.5%.

lavalaz:
From my perspective, that is...  ..."undesirable" if the bins do not correspond to individual DN values.  If 15% of the pixels have a DN value only 1 above low saturation, then having them all go to low saturation when I ask for the bottom 0.5% to be clipped would be reasonable.  I really don't understand the details of how the math is being done so I can't give you a more specific suggestion at this point.  But if it helps, I can describe what I would expect from various pathologic (but plausible) data sets...

thareUSGS:
I agree with @lavalaz for the proper use case. Doesn't this come down to an acceptable histogram bin size. It seems for 16bit and higher data, ISIS uses "nbins = 65536;" which appears to be very sufficient for slicing data to return a good percentage value.  Thus I wonder how much of impact in practice this might really be? Better to get it correct, but I assume with that number of bins, it would be very close if not generally the same answer (by taking the middle of the bin). But a good test case would help.

https://github.com/USGS-Astrogeology/ISIS3/blob/ad037e654f1e4b7a2e1cd0d31a10b56f78c24bef/isis/src/base/objs/Histogram/Histogram.cpp#L298

lavalaz:
Sorry, Trent's link to the code confused me.  For 32-bit real DN values, when isis collects statistics via _hist_, it seems to be compressing the DN values down to 16-bit?  The trick of having a bin for every possible DN value was a very clever trick to reduce the computational expense of generating statistics.  But for 32-bit real values, you need 4.29 billion bins, not 65536?  

But returning to Trent's first point, 65536 bins seems very adequate in most cases.  But there are situations, that actually happen, where this is not the case.  When an image is over- or under-exposed, the histogram is very skewed.  And much of the scientifically valuable data (and a large fraction of the valid pixels) are crammed right up against the high or low saturation limit.  The other case is when a high-dynamic-range imager takes images of a sunlit scene with interesting features in the deep shadows of a crater. There can be extremely interesting information in just a few DN values barely above low saturation.  I am not suggesting any specific fix, but I want to point out real cases where "very close to correct" numerically is a long ways from "correct" for the user.  I appreciate the efforts to sort this out and will definitely be someone taking advantage of even subtle improvements.  Thanks!

If it helps, I can dig up some specific images that could be used to test what I would call "realistic edge cases" as opposed to hypothetical pathologies...

jessemapel:
@lavalaz The default is 65536 because it's good for most cases. The situations were users want higher precision than that, such as the examples you gave, are handled by allowing users to specify the number of bins they want. I don't think we need special test cases for that.

lavalaz:
Agreed for _hist_.  I have lost lock on what _stretch_ is doing behind the scenes, but am sure you all will sort it out great!

isis/src/base/apps/gaussstretch/main.cpp
```diff
@@ -17,10 +17,10 @@ void IsisMain() {
...
 
   for(int i = 0; i < icube->bandCount(); i++) {
-    Histogram hist = *(icube->histogram(i + 1));
-    double mean = (hist.Maximum() + hist.Minimum()) / 2.0;
-    double stdev = (hist.Maximum() - hist.Minimum()) / (2.0 * gsigma);
-    stretch.push_back(new GaussianStretch(hist, mean, stdev));
+    Histogram *hist = icube->histogram(i + 1);
+    double mean = (hist->Maximum() + hist->Minimum()) / 2.0;
+    double stdev = (hist->Maximum() - hist->Minimum()) / (2.0 * gsigma);
+    stretch.push_back(new GaussianStretch(*hist, mean, stdev));
```
jessemapel, line 23: This looks like it used to be, and still is a memory leak.

isis/src/base/objs/GaussianStretch/GaussianStretch.h
```diff
@@ -25,6 +25,7 @@
 
 #include "Statistics.h"
 #include "Histogram.h"
+#include "Histogram.h"
```
jessemapel: Double include isn't needed

isis/src/base/objs/Histogram/Histogram.cpp
```diff
@@ -138,9 +137,6 @@ namespace Isis {
 
     //set the ranges
     rangesFromNet(net,statFunc);
-
-    //add all the data to the now setup histogram
-    addMeasureDataFromNet(net,statFunc);
```
jessemapel, line 143: Why was this removed? This looks like it doesn't add the data to the histogram anymore.

isis/src/base/objs/Histogram/Histogram.cpp
```diff
@@ -163,32 +159,15 @@ namespace Isis {
...
-   //SetValidRange(binWidth*floor(this->ValidMinimum()/binWidth),
-   //              binWidth*ceil(this->ValidMaximum()/binWidth));
-
    //from the domain of the data and the requested bin width calculate the number of bins
    double domain = this->ValidMaximum() - this->ValidMinimum();
    int nBins = int ( ceil(domain/binWidth) );
    SetBins(nBins);
-
-   //add all the data to the now setup histogram
-   addMeasureDataFromNet(net,statFunc);
```
jessemapel, line 191: Same as other control network constructor. This doesn't actually add the control network data.

isis/src/base/objs/Histogram/Histogram.h
```diff
@@ -127,16 +132,19 @@ namespace Isis {
       void SetValidRange(const double minimum = Isis::ValidMinimum,
                                        const double maximum = Isis::ValidMaximum);
 
+      void addMeasureDataFromNet(ControlNet &net, double(ControlMeasure::*statFunc)() const);
```
jessemapel: Why did this get promoted to public? I'd image it goes to protected.

isis/src/base/objs/Histogram/unitTest.cpp
```diff
@@ -28,7 +29,7 @@ int main(int argc, char *argv[]) {
     // Old unit test begins here
 
     Isis::Preference::Preferences(true);
-    Isis::Histogram h(-10.0, 10.0, 21);
+    Isis::ImageHistogram h(-10.0, 10.0, 21);
```
jessemapel: The Histogram unit test using the child class and not itself is incorrect.

isis/src/base/objs/ImageHistogram/ImageHistogram.cpp
```diff
@@ -0,0 +1,230 @@
...
+   * @param statFunc: Pointer to a ControlMeasure acessor, the returns of this
+   *    function call will be used to build up the network.
+   * @param bins:  The number of bins to divide the histogram into.
+   * @throws The number of Histogram Bins must be greater than 0.
+   */
+  ImageHistogram::ImageHistogram(ControlNet &net, double(ControlMeasure::*statFunc)() const, int bins) :
+  Histogram(net, statFunc, bins) {
+
+    //add all the data to the now setup histogram
+    addMeasureDataFromNet(net, statFunc);
```
jessemapel: Is this why the control network constructor doesn't add the data in the Histogram class anymore?

isis/src/base/objs/ImageHistogram/ImageHistogram.h
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @ingroup Statistics
+   *
+   * @author 2002-05-13 Jeff Anderson
+   *
+   * @internal
+   *   @todo This class needs an example.
+   *   @history 2002-05-22 Jeff Anderson moved Reset, AddData, and RemoveData
+   *                            methods into public space.
+   */
```
jessemapel, line 55: These docs look incorrect for this class.

isis/src/control/apps/cnetwinnow/cnetwinnow.cpp
```diff
@@ -90,7 +90,7 @@ namespace Isis {
       }
     }
     //build a Histogram of the residual Magnitudes from the ControlNet
-    Histogram hist(net, &ControlMeasure::GetResidualMagnitude, ui.GetDouble("BIN_WIDTH"));
+    ImageHistogram hist(net, &ControlMeasure::GetResidualMagnitude, ui.GetDouble("BIN_WIDTH"));
```
jessemapel: This should use the new Histogram class because it is working with residuals, not pixel values.

jessemapel (dismissed):

jessemapel:
@acpaquette This looks ready to go. Merge when you're ready to checkin updated app test data

jessemapel (approved):

thareUSGS:
Don't let this change anything here, just ran across some interesting code where the number of bins for data types above 16bit default to 10000000 bins.

https://github.com/gina-alaska/dans-gdal-scripts/blob/5d933296941a92d9d0c8f54e4dfdf4fb0e357550/src/gdal_contrast_stretch.cc#L299

PR #4010, opened 2020-09-11
Adds slope calculations to phocube

### LINKED ISSUE #3645: Phocube missing essential output options (local phase angle, slope, slope azimuth)
**Description**  

The name of the phocube application suggests it is for doing photometric analyses which is the comparison of phase angle versus reflectivity.  The most important photometric parameter is the phase angle, corrected for topography.  This would be called "LOCALPHASE" in the parlance of phocube.  Phocube has LOCAL_EMISSION and LOCAL_INCIDENCE which means all the math to calculate local phase angle is being done, but there is no option to output the one parameter that really matters. It is essential to also output the local incidence and local emission angle values to allow proper error analysis of the local phase angles. 

The request is to add three (3) new output options in phocube:

- LOCALPHASE (phase angle computed for each image pixel, using the DEM. Same math as calculating the phase angle without the DEM, but using the local incidence and local emission angle)

- SLOPE (slope from the DEM that is used to calculate LOCALEMISSION, LOCALINCIDENCE, and LOCALPHASE; this is already being computed for each pixel, just asking for an option to output it as a band)

- SLOPEAZIMUTH (azimuth of the downslope direction from the DEM that is used to calculate LOCALEMISSION, LOCALINCIDENCE, and LOCALPHASE; this is already being computed for each pixel, just asking for an option to output it as a band)

**Example**  

phocube from=EW0131773041G_cal.cub to EW0131773041G_cal.pho.cub dn=yes localphase=yes slope=yes slopeazimuth=yes

jessemapel:
## Description

This adds computation of the slope and slope azimuth/aspect to phocube. The slope computation is sufficiently complex that I put it in Camera similar to the local photometric angle computations. The slope azimuth computation is fairly simple, so I left it in phocube like the other ground azimuth computations.

## Related Issue

Fixes #3645 

## Motivation and Context

ISIS supports slope computations in level2, projected, images via the slpmap application. This adds the ability to do these computations in level1 images. Without this, users have to use map2cam to convert slpmap output into a level1 image and that introduces interpolation. These values are of particular interest for photometric analysis.

## How Has This Been Tested?

All tests passed still. The new output was compared against slpmap output here: https://github.com/USGS-Astrogeology/ISIS3/issues/3645#issuecomment-690787489

See issue comment

## Types of changes

## Checklist:

- [x] I have added myself to the [AUTHORS.rst](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/AUTHORS.rst) document.

## Licensing

### DIFF
--- a/isis/src/base/apps/phocube/main.cpp
+++ b/isis/src/base/apps/phocube/main.cpp
@@ -3,12 +3,16 @@
 #include "Angle.h"
 #include "Camera.h"
 #include "Cube.h"
+#include "Displacement.h"
 #include "FileName.h"
 #include "IException.h"
+#include "Latitude.h"
+#include "Longitude.h"
 #include "ProjectionFactory.h"
 #include "ProcessByBrick.h"
 #include "ProcessByLine.h"
 #include "SpecialPixel.h"
+#include "SurfacePoint.h"
 #include "TProjection.h"
 
 #include <cmath>
@@ -90,6 +94,8 @@ void IsisMain() {
   bool sunAzimuth = false;
   bool spacecraftAzimuth = false;
   bool offnadirAngle = false;
+  bool slope = false;
+  bool slopeAzimuth = false;
   bool subSpacecraftGroundAzimuth = false;
   bool subSolarGroundAzimuth = false;
   bool morphologyRank = false;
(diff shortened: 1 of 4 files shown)


jessemapel:
This is currently a draft as it's missing:

1. A regression test. I am still validating the test data for this.
1. A changelog entry. I think this needs to be rebased onto dev before I do this to avoid merge conflicts.

jessemapel:
I am closing this PR for right now because it will need to be picked up in the next project sprint. Here's the current status:

1. The slope calculation is working and is ready to go.
1. The slope azimuth calculation is having issues with relatively flat areas. There is significant variation that is not seen in the slpmap output. I attempted to recreate the logic in Camera::GroundAzimuth without using lat/lon and just using rectangular coordinates, but that resulted in the same issue. I think we'll need to actually compute the projection of the North and local normal vectors onto the plane tangent to the ellipsoid at the ground point and then compute their angle.

PR #4021, opened 2020-09-16
Hyb2pds4gen tests

Kelvinrr:
## Description

Adds tests to Hyb2pds4gen app. 

Also fixes bug where Additional PvlObjects other than IsisCube were not bing added correctly in `Cube::fromLabel` method. 

## Motivation and Context

Get pds4gen tested. 

## How Has This Been Tested?

Test pass locally. 

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -98,6 +98,10 @@ namespace Isis {
       putGroup(*grpIt);
     }
 
+    for (int i = 1; i < label.objects(); i++) {
```
scsides, line 101: I understand why this is needed for this app. but this is concerning: At this point why don't we just copy the whole input PVL? Where else is this function used? What other tests will this influence?

isis/tests/Fixtures.cpp
```diff
@@ -202,4 +201,96 @@ namespace Isis {
     delete threeImageOverlapFile; 
     delete twoImageOverlapFile; 
   }
+
```
scsides, line 204: Should we find a way to better organize these one-off fixtures? This file is going to get crowded if we keep adding 60+ engest programs to it, plus any others that come along.

isis/tests/FunctionalTestsHyb2Pds4Gen.cpp
```diff
@@ -0,0 +1,119 @@
...
+  testCube = new Cube(fileName, "rw");
+
+  QVector<QString> args = {"to=/tmp/output", "PDS4LOGICALIDENTIFIER=Whatever"};
+  UserInterface options(APP_XML, args);
+  
+  hyb2pds4gen(testCube, options);
+  
+  QDomDocument pds4lab;
+
+  QFile f("/tmp/output.xml");
```
scsides, line 50: Won't these collide if we have more than one test running?

isis/tests/data/defaultImage/defaultCube.pvl
```diff
@@ -89,6 +89,16 @@ Object = IsisCube
     Status   = Nominal
   End_Group
 End_Object
+Object = History
```
scsides, line 92: I know why hyb2 needs these, but why does the default cube need these? They are not required objects.

scsides (changes requested):
Mostly questions/thoughts, but one race condition that needs looking into.

isis/src/base/objs/Cube/Cube.cpp
Kelvinrr: I thought the same, but it seems that things from the IsisCube PVL Object need to be set in a particular way? Almost like the init process creates a simple IsisCube Pvl Object so replacing it directly seems unsafe.

isis/tests/Fixtures.cpp
Kelvinrr: I was thinking there is a way to make the mission specific camera stuff more generic, but after looking at options it seems we will inevitably end up with a lot of fixtures for different missions. @kaitlyndlee and I briefly talked about this, I personally don't mind letting Fixtures.h get long, but if we wanted a `MissionFixtures.h/.cpp` I wouldn't be against it.

isis/tests/FunctionalTestsHyb2Pds4Gen.cpp
Kelvinrr: oof, I made that local for testing purposes and forgot to revert, good catch, my bad.

isis/tests/data/defaultImage/defaultCube.pvl
Kelvinrr: hyb2pds4gen needed these as it expect them to be there as part of the translation. Since it didn't seem like a good idea to change the functionality of `ProcessExportPds4` by making them optional, so I added them to the cube. I imagine all Pds4 related tests will have this issue so I made the changes to defaultcube fixture so someone else won't run into the same issue I did (also doesn't negatively impact other tests).

PR #4022, opened 2020-09-17
Fixed FunctionalTestGetsnAllTrue by creating non-static APP_XML

AustinSanders:
Fixed FunctionalTestGetsnAllTrue test by instantiating a non-static APP_XML instance

## Description
FunctionalTestGetsnAllTrue passed locally, but failed in CI.

This is a super weird issue that I really don't understand.  The idea for trying this fix came from a discussion of "leaky tests" found [here](https://objectpartners.com/2018/09/18/why-your-tests-may-pass-locally-but-fail-in-jenkins/).  The post mentioned potential issues regarding persistent data, and it went on to talk about how static classes and static variables sometimes result in data carryover among tests.

This doesn't answer the question of why processing resulted in a "child aborted" exception.  It seems like the carryover from static variables would just result in munged values, not full-on process breaking exceptions.  Maybe this resulted in the right change for the wrong reasons?

## Related Issue

getsnAllTrue test failure

## Motivation and Context

Gotta have passing tests!

## How Has This Been Tested?

Since the failures were CI-side only, this was painfully tested by pushing new commits to a draft PR and waiting for CI to pick up, build, and test the changes. Yikes.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/FunctionalTestsGetsn.cpp
+++ b/isis/tests/FunctionalTestsGetsn.cpp
@@ -22,18 +22,18 @@ static QString APP_XML = FileName("$ISISROOT/bin/xml/getsn.xml").expanded();
 
 // check for all correct outputs
 TEST_F(DefaultCube, FunctionalTestGetsnAllTrue) {
+  QString d_APP_XML = FileName("$ISISROOT/bin/xml/getsn.xml").expanded();
   QString expectedSN = "Viking1/VISB/33322515";
   QString expectedON = "Viking1/VISB/33322515";
-  QVector<QString> args = {
-			               "FILE=TRUE",
+  QVector<QString> args = { "FILE=TRUE",
                            "SN=TRUE",
                            "OBSERVATION=TRUE"};
-  UserInterface options(APP_XML, args);
+  UserInterface options(d_APP_XML, args);
   Pvl appLog;
 
   getsn( testCube, options, &appLog );
-  PvlGroup results = appLog.findGroup("Results");  
-  
+  PvlGroup results = appLog.findGroup("Results");
+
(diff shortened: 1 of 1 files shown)


isis/tests/FunctionalTestsGetsn.cpp
```diff
@@ -22,18 +22,18 @@ static QString APP_XML = FileName("$ISISROOT/bin/xml/getsn.xml").expanded();
 
 // check for all correct outputs
 TEST_F(DefaultCube, FunctionalTestGetsnAllTrue) {
+  QString d_APP_XML = FileName("$ISISROOT/bin/xml/getsn.xml").expanded();
```
jessemapel: If static variables are an issue in test cases, we should get rid of the static variable and switch all of the test cases over to using this. Another option would be to sub-class the DefaultCube and put the app xml into the fixture.

Kelvinrr: I don't understand how a static var could cause the issue though? It should make it local to the file, if anything being non static global would cause the behavior where the var is leaking into other tests. Even within the same test module, the variable isn't being changed. The post Austin linked to talked about static classes which could have statefulness across tests, but that's not what is happening here.  

And why does this test need the var changed but the other do not? 

Lastly, putting it into the fixture would get messy especially if a set of app tests rely on different fixtures. 

I'm wondering if using something like `#DEFINE GETSN_XML_PATH` would work as an alternative to defining the string once but reusing it in a way that can't cause the var to be accidentally manipulated in any way.

Kelvinrr (approved):

PR #4035, opened 2020-09-24
Resolve Test Failure: ThreeImageNetwork.FunctionalTestCnetcheckCamera (SEGFAULT)

tgiroux:
## Description

I was able to resolve the test fail by adding a condition to protect for null values within cnetcheck.cpp noLatLonCheck(). This simply prevents a call to Camera::SetImage(const double sample, const double line) when there are null values for sample or line, which previously caused a segfault way down a nest of function calls. This small change to cnetcheck.cpp does not seem to affect the other cnet tests or break any existing functionality.
I'm unsure if Camera::SetImage() should be made safe for null parameters. After some investigating it seems to me that the changes required to do so would be far outside of the scope of this PR. I'd appreciate input from any reviewers (should an issue be raised?).

## Motivation and Context

Fixes a test failure so that we can be closer to clean builds before Oct. 1.

## How Has This Been Tested?

Rebuilt and ran `ctest -R ThreeImageNetwork` `ctest -R Cnet` to see tests that may be impacted. The only difference after the changes was that CnetcheckCamera passes instead of failing with SEGFAULT.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/control/apps/cnetcheck/cnetcheck.cpp
+++ b/isis/src/control/apps/cnetcheck/cnetcheck.cpp
@@ -635,7 +635,7 @@ namespace Isis {
 
           // Check the exact measure location
           bool setCamera = false;
-          if (createdCamera) {
+          if (createdCamera && !IsSpecial(measure->GetSample()) && !IsSpecial(measure->GetLine())) {
             setCamera = cam->SetImage(measure->GetSample(), measure->GetLine());
           }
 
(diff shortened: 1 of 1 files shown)


jessemapel (approved):
This is sufficient to resolve the issue 👍 

I would expect Camera::SetImage to fail on Null samples/lines, but it should return a False instead of seg faulting.

PR #4073, opened 2020-10-15
Overlapstats test

tgiroux:
## Description

Added gtests for overlapstats.All are working except FunctionalTestOverlapstatsErrors.

FunctionalTestOverlapstatsErrors is meant to trigger overlapstats to have a specific problem which causes it to write to the user-specified error file. This error has proved to be difficult to trigger for various reason. It's unclear how it was done within the makefile tests.

Earlier this week @jessemapel said "If this is something that's so convoluted to get, is it worth testing?" I'm starting to think FunctionalTestOverlapstatsErrors may be worth skipping here. Any input would be appreciated

Note: @Kelvinrr mentioned that within within the tests for any given app, we are not supposed to use any other apps. In this case, a list of overlaps is needed as input for overlapstats. This is best attained from using the ThreeImageNetwork with findimageoverlaps. If that's not ok here, I could instead write each overlaps.lis from a QString literal of the desired input.

## Motivation and Context

372Mb

## Types of changes

## Checklist:

## Licensing


Kelvinrr:
Right now, I am thinking that trying to get these cranked out is more important than keeping 100% coverage, especially for the sake of pacing. I would argue to for making an issue for the missed test and un-draft to get this reviewed/merged.

isis/src/base/apps/overlapstats/overlapstats.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include "PvlGroup.h"
+#include "Pvl.h"
+#include "Target.h"
+#include "TProjection.h"
+#include "ProjectionFactory.h"
+#include "PolygonTools.h"
+#include "Progress.h"
+
+using namespace std;
+using namespace Isis;
```
Kelvinrr: This is kinda everywhere and we should probably remove it. No need for using namespace if everything is in the same namespace.

isis/tests/FunctionalTestsOverlapstats.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  QString cubeListPath = tempDir.path() + "/cubes.lis";
+  QString overlapsPath = tempDir.path() + "/overlaps.lis";
+  cubes.write(cubeListPath);
+
+  // get overlaps for overlapstats input
+  QString FIO_APP_XML = FileName("$ISISROOT/bin/xml/findimageoverlaps.xml").expanded();
+  QVector<QString> fio_args = { "OVERLAPLIST=" + overlapsPath };
+  UserInterface fioOptions(FIO_APP_XML, fio_args);
+  findimageoverlaps(cubes, fioOptions);
```
Kelvinrr, line 133: I would try to get these calls out of here as it's seems cleaner to not make tests depend on multiple apps, assuming replacing it with a pre-baked overlap file isn't too much of a hassle.

isis/tests/FunctionalTestsOverlapstats.cpp
```diff
@@ -0,0 +1,259 @@
...
+  Blob pvlBlob("Footprint", "Polygon");
+  Pvl pvl;
+  PvlObject polyObject = PvlObject("Polygon");
+  polyObject.addKeyword(PvlKeyword("Name", "Footprint"));
+  polyObject.addKeyword(PvlKeyword("StartByte", "1"));
+  polyObject.addKeyword(PvlKeyword("Bytes", toString(polyStrSize)));
+  pvl.addObject(polyObject);
+
+  pvlBlob.Read(pvl, polyStream);
+  cube3->write(pvlBlob);
```
Kelvinrr: Would it be easy enough to replace this with the cleaner version in the fixture? 

e.g: 
```C++
ImagePolygon poly;
    coords = {{30, 0},
              {30, 10},
              {35, 10},
              {35, 0},
              {30, 0}};
poly.Create(coords);
cube1->write(poly);
cube1->reopen("rw");
```

Kelvinrr (commented):
Overall looks fine, just some nitpicky stuff.

isis/tests/FunctionalTestsOverlapstats.cpp
tgiroux: Not sure where you're seeing this, & I'm unable to get such a thing to work since ImagePolygon::Create() takes much different parameters. What I'm doing here seems to be the convention based on whats in the fixture and other tests using ThreeImageNetwork.

Kelvinrr: I'm seeing it on the fixtures for three image adjacency: https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/tests/Fixtures.cpp#L186 

If it's not easily do-able you can not worry about it

tgiroux: I stand corrected, that's some of adams work that was merged last week. Looks like I could totally do that and it's way cleaner.

Kelvinrr (approved):
I think once Jenkins comes back okay this is good to merge.

isis/tests/FunctionalTestsOverlapstats.cpp
```diff
@@ -0,0 +1,199 @@
...
+  cubes.append(FileName(cube1->fileName()));
+  cubes.write(badCubeList);
+  cubes.append(FileName(cube2->fileName()));
+  cube1->close();
+  cube2->close();
+
+  QString FIO_APP_XML = FileName("$ISISROOT/bin/xml/findimageoverlaps.xml").expanded();
+  QVector<QString> fio_args = { "OVERLAPLIST=" + overlapsPath };
+  UserInterface fioOptions(FIO_APP_XML, fio_args);
+  findimageoverlaps(cubes, fioOptions, false, nullptr);
```
acpaquette, line 32: These calls to findimageoverlaps need to be replaced using the overlap.lis file that findimageoverlaps produces. I'm fine with creating the overlap files then adding them to the code base. Like @Kelvinrr mentioned before we don't want these gtest app tests to be dependent on other apps.

acpaquette (changes requested):

isis/tests/FunctionalTestsOverlapstats.cpp
Kelvinrr: In the name of keeping momentum, we agreed that smaller issues like this can be punted and dealt with later when the size of the data area is smaller. Since Tim has had issues costing him a few days already (trying to get a overlap file working included), I think it's best to punt on the problem for now.

acpaquette (approved):

PR #4080, opened 2020-10-23
Adds apollo fixture

krlberry:
## Description
Adds apollo fixture for jigsaw tests.

## Types of changes

## Checklist:

## Licensing


isis/tests/Fixtures.cpp
```diff
@@ -122,6 +122,7 @@ namespace Isis {
     TempTestingFiles::SetUp();
 
     std::ifstream isdFile("data/defaultImage/defaultCube.isd");
+    std::ifstream isdFile1("data/defaultImage/defaultCube.isd");
```
kaitlyndlee: I don't see an isdFile1 being used in the DefaultCube. Do we need this here?

krlberry: Good catch! This doesn't need to be in there. I'll update it.

isis/tests/Fixtures.cpp
```diff
@@ -253,7 +255,110 @@ namespace Isis {
...
+    isdFile6 = new FileName("data/apolloNetwork/Sub4-AS15-M-1423.isd");
+    isdFile7 = new FileName("data/apolloNetwork/Sub4-AS15-M-1537.isd");
+
+    label1 = new FileName("data/apolloNetwork/Sub4-AS15-M-0583_msk.pvl");
+    label2 = new FileName("data/apolloNetwork/Sub4-AS15-M-0584_msk.pvl");
+    label3 = new FileName("data/apolloNetwork/Sub4-AS15-M-0585_msk.pvl");
+    label4 = new FileName("data/apolloNetwork/Sub4-AS15-M-0586_msk.pvl");
+    label5 = new FileName("data/apolloNetwork/Sub4-AS15-M-0587_msk.pvl");
+    label6 = new FileName("data/apolloNetwork/Sub4-AS15-M-1423.pvl");
+    label7 = new FileName("data/apolloNetwork/Sub4-AS15-M-1537.pvl");
```
Kelvinrr: probably a dumb question: can we get away with less? If not, can we use an arrays?

isis/tests/Fixtures.h
```diff
@@ -121,6 +121,19 @@ namespace Isis {
       void TearDown() override;
   };
 
+  class ApolloNetwork: public TempTestingFiles {
+    protected:
+      Cube *cube1, *cube2, *cube3, *cube4, *cube5, *cube6, *cube7; 
+      FileName *isdFile1, *isdFile2, *isdFile3, *isdFile4, *isdFile5, *isdFile6, *isdFile7, 
+           *cubeLabel, *label1, *label2, *label3, *label4, *label5, *label6, *label7;
+
```
Kelvinrr: Arrays would probably clean some of this up.

isis/tests/Fixtures.cpp
```diff
@@ -253,7 +255,110 @@ namespace Isis {
     delete threeImageOverlapFile;
     delete twoImageOverlapFile;
   }
-  
+
+  void ApolloNetwork::SetUp() {
+    TempTestingFiles::SetUp();
+    
+    isdFile1 = new FileName("data/apolloNetwork/Sub4-AS15-M-0583_msk.isd");
```
Kelvinrr: usually names are generic or used to define the kind of images they represent (e.g. steropairL/steropairR or mroLabel.pvl or image1, image2, image3  if it doesn't matter). Could these simply be "apolloImage<n>.pvl/.isd? or something similar?

Kelvinrr (commented):
Looks good, just some questions about making things cleaner.

isis/tests/Fixtures.h
```diff
@@ -121,6 +121,19 @@ namespace Isis {
...
+      FileList *cubeList;
+      QString cubeListFile;
```
Kelvinrr, line 131: might want to add the control net path to this.

isis/tests/Fixtures.cpp
jessemapel: Someone could pull these up in qmos or qgis and look at the overlap. We want to have a few points with 3-4 measures but not all of them are needed. It's possible some of these could go away. I don't think an array will save any hassle here, each FileName still has to be initialized by itself.

isis/tests/Fixtures.cpp
```diff
@@ -253,7 +255,110 @@ namespace Isis {
...
+
+    delete cube1;
+    delete cube2;
+    delete cube3;
+    delete cube4;
+    delete cube5;
+    delete cube6;
+    delete cube7;
+
+    delete cubeList;
```
jessemapel: We really should make these fixtures more careful about just raw deleting things. Either we should move to smart pointers or we should wrap our deletes in if checks.

isis/tests/Fixtures.cpp
Kelvinrr: If the filenames are also programmatic it would fix that I think (like in my comment at the bottom). Something like labels[i] = new FileName("data/apolloImage" + i + ".pvl");

isis/tests/Fixtures.h
Kelvinrr: also, seems like the network and it's path is missing

isis/tests/Fixtures.cpp
krlberry: Yeah, they can just be apolloImage1 -> n.

isis/tests/Fixtures.cpp
krlberry: I feel like smart pointers is a better idea, but I just updated to wrap in "if checks" for now.

isis/tests/Fixtures.h
krlberry: Added the path for the control network. For the network, since it's opened by jigsaw from a file, I didn't include the ControlNetwork itself as a member variable in the Fixture class. Definitely let me know if I missed something here!

isis/tests/Fixtures.cpp
krlberry: On the "using an array" thing. I expected this to work, and thought it was a good idea, and then when I tried it, I ran into trouble. I nominate this specific thing for being a thing you update in a future PR if you it's not out of the way.

Kelvinrr (approved):

PR #4088, opened 2020-10-26
Converted jigsaw apollo test

krlberry:
## Description
Converts the jigsaw apollo test to gtest. 

Also includes:
* A helper function to compare csv lines for convenience.
* Updates to the `Cube::createFromIsd()` functions to include exception throws if the isd or pvl file cannot be opened. This 
was an issue I had and had difficulty debugging without this change because the error I received (for this issue) was a "json parse" exception.

## Related Issue
Part of test conversion.

## Motivation and Context
Part of test conversion issue. Decreases size of necessary test data. 

## How Has This Been Tested?
This is a test. 

## Types of changes

## Checklist:

## Licensing


isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+  QStringList compareMe = headerStr.split(",");
+  for (int i=initialIndex; i<compareMe.size(); i++) {
+    if (isNumeric(compareMe[i].trimmed())) {
+      EXPECT_NEAR(csvLine[i].toDouble(), compareMe[i].toDouble(), 0.000001);
+    }
+    else{
+      EXPECT_EQ(QString(csvLine[i]).toStdString(), compareMe[i].toStdString()); 
+    }
+  }
+};
```
Kelvinrr: Probably move this to Utilities.cpp?

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+                            "update=no",
+                            "bundleout_txt=no",
+                            "cnet="+tempDir.path()+"/apolloNet.net",
+                            "fromlist="+tempDir.path() + "/cubes.lis",
+                            "onet="+tempDir.path()+"/apollo_out.net",
+  "file_prefix="+tempDir.path()+"/"};
+
+  UserInterface ui1(APP_XML, args1);
+
+  jigsaw(ui1);
```
Kelvinrr: probably get rid of the "1" and just call it ui

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+    else{
+      EXPECT_EQ(QString(csvLine[i]).toStdString(), compareMe[i].toStdString()); 
+    }
+  }
+};
+
+
+TEST_F(ApolloNetwork, FunctionalTestJigsawFirst) {
+  QString controlNetFile = tempDir.path()+"/apollo.net";
```
Kelvinrr: this path should probably be saved as part of the fixture.

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+  QString imagesOutput = tempDir.path() + "/bundleout_images.csv";
+  QString residualsOutput = tempDir.path() + "/residuals.csv";
+  ControlNet outputNet(tempDir.path()+"/apollo_out.net");
+
+  // Check for the correct header output format and csv file structure for the points.csv file
+  CSVReader::CSVAxis csvLine;
+  CSVReader header = CSVReader(pointsOutput,
+                               false, 0, ',', false, true);
+
+  int numColumns = header.columns();
```
Kelvinrr: This process is definitely better for comparing headers than what I was doing. That being said, as jigsaw is kinda all over the place (in the sense of it relies on a number of objects). Try to check what parts of the CSV isn't already tested by one of these objects as part of a unittest (I think bundleobservation and bundleresults do a lot of the work? there is also bundleUtilities/unittest.cpp that does a lot of testing).

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+  compareCsvLine(csvLine, "Point,Image,Image,coordinate,coordinate,Sample,Line,residual,residual,Magnitude");
+  csvLine = header.getRow(2);
+  compareCsvLine(csvLine, "Label,Filename,Serial Number,(mm),(mm),(pixels),(pixels),(pixels),(pixels),(pixels),Rejected");
+
+  // Check line/sample and residuals 
+  // Check all measures
+  ControlMeasure* measure;
+  for (int i=3; i < numRows; i++) {
+    csvLine = header.getRow(i);
+    EXPECT_NO_THROW({
```
Kelvinrr, line 183: I wonder how necessary these EXPECT_THROW are? If nothing is catching them, can we not let them naturally fail?

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  // Spot check a few measures for hard-coded values:
+  compareCsvLine(header.getRow(14), "AS15_000031448,/tmp/qt_temp-l7wTTZ/cube1.cub,APOLLO15/METRIC/1971-07-31T14:00:53.547,    -24.91466687,     -8.24555718,   4109.77150653,   2450.19288272,-0.00343036,      0.70304341,      0.70305178", 2); 
+
+  compareCsvLine(header.getRow(142), "AS15_000032200,/tmp/qt_temp-l7wTTZ/cube2.cub,APOLLO15/METRIC/1971-07-31T14:01:16.947,    -25.59176645,    -10.57595225,   4143.71597937,   2333.56318790,     -0.00372340,      0.48459237,      0.48460667", 2);
+  compareCsvLine(header.getRow(424), "AS15_000055094,/tmp/qt_temp-l7wTTZ/cube1.cub,APOLLO15/METRIC/1971-07-31T14:00:53.547,     20.35945982,     34.23830188,   1844.18431849,   4576.36730130,      0.00691810,     -0.57578795,      0.57582951", 2);
+  compareCsvLine(header.getRow(970), "AS15_test01,/tmp/qt_temp-l7wTTZ/cube3.cub,APOLLO15/METRIC/1971-07-31T14:01:40.346,     -5.04180936,    -34.53366079,   3115.51026031,   1134.42313078,     -3.07166949,      1.44947401,      3.39648765", 2);
+
+  // Test output network size
+  ControlNet inputNet("$ISISROOT/../isis/tests/data/apolloNetwork/apollo.net");
```
Kelvinrr: I recall you having issues with the test data area but this should probably be a relative path or a path in fixture (as Jenkins doesn't seem to complain with relative paths).

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+using namespace Isis;
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/jigsaw.xml").expanded();
+
+// Check to see if a QString contains only numeric values.
+bool isNumeric(QString str){
+  QRegExp re("-*\\d*.*\\d*");  // a digit (\d), zero or more times (*)
+  return re.exactMatch(str);
+}
```
Kelvinrr: Also should probably go to utilities.cpp.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -133,11 +133,39 @@ namespace Isis {
     std::ifstream isdStream(isdFile.expanded().toStdString());
     std::ifstream labelStream(labelFile.expanded().toStdString());
 
-    Pvl label;
+    if (isdStream.fail()) {
+      QString msg = QString("failed to open isd stream: %1").arg(isdFile.expanded());
+      throw IException(IException::Io, msg,
+                 isdFile.baseName().toStdString().c_str(), 153);
+    }
```
Kelvinrr, line 140: this is good because I would sometimes mess up the path to the ISD and I get a bogus error from the JSON lib. Would this also trigger in the case of a file not existing? (presumably yes).

Kelvinrr (commented):
some questions in line. I would favor this over the other Apollo PR with only the fixtures so I vote for closing the other.

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: Good catch, didn't realize I'd left this in.

isis/src/base/objs/Cube/Cube.cpp
krlberry: Yeah, it also triggers if the file doesn't exist, which was the issue I was running into!

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: My intent here was to indicate that the this part for the test resulting in an exception would be a failure of the test, like _explicitly_, i.e. "not throwing is something being tested" and this was my attempt to indicate that.  

There would be no technical problem with just letting them fail, and I'm not sure how much extra this adds. Maybe I'll add one of those "custom error messages" to supply more context.

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: Can do. Didn't do initially because I wasn't sure how useful they would be elsewhere.

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: I did this. Let me know what you think.

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: Good point. 

I did some grepping, and it looks like the header values for the any of the csv files aren't being tested anywhere, so I think it makes sense to leave the header comparisons in for each of the output csv files, too, and just test them in this _one_ jigsaw test. I don't think the headers need to be tested for any other jigsaw tests, since they're covered here.

The other csv line comparisons just "spot-check" a few points to make sure their exact values match the expected values in the current output, though I have no particular preference for doing this as opposed to taking the statics approach you took. I'm tempted to leave all of this in. What do you think?

krlberry:
I'm going to wait until the apollo fixture is merged (PR #4080) and then go ahead and merge dev into this branch to fix merge conflicts.

Kelvinrr (dismissed):
if jenkins comes back okay I think this is good to merge

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -13,6 +34,200 @@ using namespace Isis;
...
+    SurfacePoint sp = outputPoint->GetAdjustedSurfacePoint();
+    SurfacePoint originalSP = outputPoint->GetAprioriSurfacePoint();
+    EXPECT_NEAR(sp.GetLatitude().planetocentric(Angle::Degrees), csvLine[5].toDouble(), 0.000001);
+    EXPECT_NEAR(sp.GetLongitude().positiveEast(Angle::Degrees), csvLine[6].toDouble(), 0.000001);
+    EXPECT_NEAR(sp.GetLocalRadius().kilometers(), csvLine[7].toDouble(), 0.000001);
+    EXPECT_NEAR(sp.GetLatSigmaDistance().meters(), csvLine[8].toDouble(), 0.000001);
+    EXPECT_NEAR(sp.GetLonSigmaDistance().meters(), csvLine[9].toDouble(), 0.000001);
+    EXPECT_NEAR(sp.GetLocalRadiusSigma().meters(), csvLine[10].toDouble(), 0.000001);
+
+    EXPECT_NEAR(sp.LatitudeToMeters(sp.GetLatitude().planetocentric(Angle::Radians)) -
```
krlberry: @jessemapel This was my attempt to compare the corrected latitude and longitude in meters from the control network to the csv output, but the comparison I did here didn't quite work out. (See: 1000.0 as the tolerance.) Do you know what the appropriate numbers to compare would be and if they are even available at the ControlNetwork level? (The information that is written directly to the csv uses the correction from the bundle directly.)

Kelvinrr (approved):

PR #4096, opened 2020-10-30
chan1m32isis gtest

AustinSanders:
Converts chan1m32isis to .cpp/.h and converts makefile tests to gtest

## Description
Converts chan1m32isis to .cpp/.h and converts makefile tests to gtest

## Related Issue
Test conversion

## Motivation and Context
Eliminates 1176Mb of test data

## How Has This Been Tested?
6 Gtests

## Types of changes

## Checklist:

## Licensing


isis/src/chandrayaan1/apps/chan1m32isis/chan1m32isis.cpp
```diff
@@ -0,0 +1,480 @@
...
+  void flipUtcTable(Table &utcTable);
+  Cube *g_oCube;
+  Brick *g_oBuff;
+  int g_totalLinesAdded;
+  double g_expectedLineRate;
+  Table *g_utcTable;
+  PvlGroup g_results("Results");
+
+  void chan1m32isis(UserInterface &ui) {
+    //g_results.clear();
```
AustinSanders: This line results in a segfault during testing.  This needs to be fixed, but I wanted to get the PR up ASAP.  Any ideas?

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,550 @@
...
+  ASSERT_DOUBLE_EQ(hist->Average(), 15.58169779027077);
+  ASSERT_DOUBLE_EQ(hist->Sum(), 47368.361282423139);
+  ASSERT_EQ(hist->ValidPixels(), 3040);
+  ASSERT_DOUBLE_EQ(hist->StandardDeviation(), 2.2696592481066249);
+
+  delete hist;
+}
+
+
+TEST(Chan1m32Isis, Chan1m32IsisTestFowardDescending) {
```
AustinSanders, line 113: I'm not sure if we need to test all of the same keywords + DNs for each of the forward/reverse ascending/descending, but I wrote the tests anyway.  We can trim this down as needed.

isis/src/chandrayaan1/apps/chan1m32isis/chan1m32isis.cpp
kaitlyndlee: UserInterface inherits from IsisAml and the `WasEntered` method called in `importImage` is inherited from IsisAml. This could be relating to the problem and the fact that in the tests, we are not using the Application class anymore.

isis/tests/data/chan1m32isis/badFile/bad.file
```diff
@@ -0,0 +1 @@
+>:(
```
kaitlyndlee: Is this needed or could we use an ofstream to create it programmatically? An example of this is in the campt tests.

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,550 @@
...
+
+
+TEST(Chan1m32Isis, Chan1m32IsisTestBadFile) {
+  QTemporaryDir prefix;
+  QString cubeFileName = prefix.path() + "/chan1m32isisTEMP.cub";
+  QVector<QString> args = {"from=data/chan1m32isis/badFile/bad.file",
+                           "to=" + cubeFileName };
+
+  UserInterface options(APP_XML, args);
+  ASSERT_THROW(chan1m32isis(options), IException);
```
kaitlyndlee: @Kelvinrr It seems like there are multiple ways we are checking errors. I think we should come up with a standard.

Kelvinrr: Good catch.

I think sticking with `try{app(); FAIL();}Catch(){e.what substring check}` is the way to go, it allows you to assert both the error type and the message. Especially for cases like this where the exception type is not enough to reasonably assume that the Exception was thrown for the right reason. Here, IException can be triggered by a number of things that may not be what you want. 

@AustinSanders I would recommend switching to the try/except approach instead. see: https://github.com/USGS-Astrogeology/ISIS3/blob/24b1fc7e1c2c1182f2b454c4ecb0d4394a424d14/isis/tests/FunctionalTestsJigsaw.cpp#L422

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+using ::testing::HasSubstr;
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/chan1m32isis.xml").expanded();
+
+TEST(Chan1m32Isis, Chan1m32IsisTestFowardAscending) {
+  QTemporaryDir prefix;
+  QString cubeFileName = prefix.path() + "/chan1m32isisTEMP.cub";
+  QVector<QString> args = {"from=data/chan1m32isis/forwardAscending/M3T20090630T083407_V03_L1B_cropped.LBL",
+                           "loc=data/chan1m32isis/forwardAscending/M3T20090630T083407_V03_LOC_cropped.IMG",
+                           "obs=data/chan1m32isis/forwardAscending/M3T20090630T083407_V03_OBS_cropped.IMG",
```
acpaquette: Could we copy these into the temporary testing directory that gets created and use the copied files here. If we don't there are a couple cubes that get created in the test data directory. Maybe using something like https://en.cppreference.com/w/cpp/filesystem/copy

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+  ASSERT_EQ(int(dimensions["Samples"]), 608);
+  ASSERT_EQ(int(dimensions["Lines"]), 5);
+  ASSERT_EQ(int(dimensions["Bands"]), 3);
+
+  // Pixels Group
+  PvlGroup &pixels = isisLabel->findGroup("Pixels", Pvl::Traverse);
+  ASSERT_EQ(pixels["Type"][0].toStdString(), "Real");
+  ASSERT_EQ(pixels["ByteOrder"][0].toStdString(), "Lsb");
+  ASSERT_EQ(double(pixels["Base"]), 0.0);
+  ASSERT_EQ(double(pixels["Multiplier"]), 1.0);
```
acpaquette: Another comment made on another 2isis PR, these values can be accessed directly from functions that exist on the cube. 

`sample/line/bandcount` for the number of samples, lines, and bands
`pixelType, bandOrder, base, and multiplier` for the latter group

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+
```
acpaquette: For subsequent tests outside of the "default" it maybe unnecessary to check these groups as it's often the same for images from the same instrument.

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][0].toDouble(), 1);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][21].toDouble(), 22);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][42].toDouble(), 43);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][84].toDouble(), 85);
+
+  // Kernels Group
+  PvlGroup &kernel = isisLabel->findGroup("Kernels", Pvl::Traverse);
+  ASSERT_EQ(int(kernel["NaifFrameCode"]), -86520);
+
+  Histogram *hist = cube.histogram();
```
acpaquette: Convert these to a smart pointer `std::unique_ptr<Histogram> hist (cube->histogram());` and the deletes can be removed.

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+  ASSERT_EQ(inst["SpacecraftName"][0].toStdString(), "CHANDRAYAAN-1");
+  ASSERT_EQ(inst["InstrumentId"][0].toStdString(), "M3" );
+  ASSERT_EQ(inst["TargetName"][0].toStdString(), "MOON" );
+  ASSERT_EQ(inst["SpacecraftClockStartCount"][0].toStdString(), "2/1531046.542" );
+  ASSERT_EQ(inst["SpacecraftClockStopCount"][0].toStdString(), "2/1531047.050" );
+  ASSERT_DOUBLE_EQ(double(inst["LineExposureDuration"]), 101.76);
+  ASSERT_EQ(inst["StartTime"][0].toStdString(), "2008-11-29T17:14:29.729807" );
+  ASSERT_EQ(inst["StopTime"][0].toStdString(), "2008-11-29T17:14:30.238607" );
+  ASSERT_EQ(inst["SpacecraftYawDirection"][0].toStdString(), "FORWARD" );
+  ASSERT_EQ(inst["OrbitLimbDirection"][0].toStdString(), "DESCENDING" );
```
acpaquette: Similarly outside of the default I don't know if it's worth checking every keyword here. This test really only seems to care that `SpacecraftYawDirection` is FORWARD and `OrbitLimbDirection` is DESCENDING. For tests that are explicitly looking to test those features I don't think it's super important to test all keywords in the instrument group.

isis/tests/FunctionalTestsChan1m32isis.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+  ASSERT_DOUBLE_EQ(bandbin["FilterNumber"][0].toDouble(), 5);
+  ASSERT_DOUBLE_EQ(bandbin["FilterNumber"][21].toDouble(), 57);
+  ASSERT_DOUBLE_EQ(bandbin["FilterNumber"][42].toDouble(), 99);
+  ASSERT_DOUBLE_EQ(bandbin["FilterNumber"][84].toDouble(), 253);
+
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][0].toDouble(), 1);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][21].toDouble(), 22);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][42].toDouble(), 43);
+  ASSERT_DOUBLE_EQ(bandbin["OriginalBand"][84].toDouble(), 85);
```
acpaquette: Similar thought with the bandbins group. I don't think it's super important to test outside of the default test but I would like to hear other opinions/points regarding this.

acpaquette (commented):
A few things to address regarding each test written for this instrument

acpaquette (dismissed):
Looks good!

acpaquette (approved):

PR #4107, opened 2020-11-05
Added text to description by Dave Humm. Closes #3860

### LINKED ISSUE #3860: lrowaccal outputs negative DN values
3.9.x and 4.0.0

**Description**  

Negative I/F values are physically impossible - they mean the surface is darker than a black hole. 

**How to reproduce**  

Take M102214493CE.IMG
ingest, run lrowaccal
you will find negative DN values

**Possible Solution**  

I suggest simply setting the negative values to 0.00000000 (they are small negative values)

**Additional context**

scsides:
Added documentation to describe why the calibrated image has negative values
## Description

Added documentation from Dave Humm to describe the reasons the output I/F image
may have negative DNs.

## Motivation and Context

The issue #3860 was created because negative DNs were seen in the output images. 
The LROC team, who wrote the calibration programs, supplied text to describe why those
exist.

## How Has This Been Tested?

Only a documentation change

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

## Licensing

### DIFF
--- a/isis/src/lro/apps/lronaccal/lronaccal.xml
+++ b/isis/src/lro/apps/lronaccal/lronaccal.xml
@@ -58,19 +58,39 @@
     </p>
 
     <p>
-      Let's look at the case of a calibrated image for which the true signal is zero, a dark 
-      image. In calibration, the mean bias and dark current are subtracted, but the random 
-      noise cannot be subtracted, because only the statistical distribution of the random 
-      noise can be known in advance, not the specific value for that pixel for that image. 
-      Because the random noise values are both positive and negative, when the true signal 
-      is zero the calibrated image should have both positive and negative values. If it had 
-      only positive values, the mean would be positive and therefore it would have a 
-      systematic error, since the true signal is zero. It is not possible to eliminate the 
-      random noise entirely, but it is defined as centered around zero to minimize the 
-      systematic error in the calibrated image. That forces calibrated dark images to have 
-      both positive and negative values.
+      The calibration equation is:
+      <pre>  reportedDN = ObservedDN - MeanBias - DarkCurrent </pre>
+
+        Where:
+        <pre>
+   ObservedDN = TrueDN + E
+   E is a randomly sampled value from (mu, sigma^2) and mu=0
+   TrueDN is the signal that would be reported in an idealized case of an instrument with zero noise.</pre>
+     </p>
+
+    <p>
+      Let's look at the case of a calibrated image for which the true signal
+      is zero, a dark image. In calibration the mean bias and dark current are
+      subtracted. The random noise term is then randomly sampled from a known
+      distribution with a mean of zero. Since the distribution has a mean of
+      zero, values for the random noise can be positive or negative.
+      Therefore, the addition of random noise to a pixel with true signal near
+      zero can result in negative DN values.
     </p>
 
+    <p>
+      Negative reported DNs are possible when E < -1 * TrueDN. These are
+      pixels in a very dark image that happen to have a strongly negative
+      random noise value.
+    </p>
+
+    <p>
+      Note: ObservedDN and TrueDN both must be greater than or equal to zero.
+      For ObservedDN, it's because the hardware is not able to report negative
+      DN values . For TrueDN, it's because radiance and reflectivity cannot be
+      negative. The dimmest target is one that is completely dark, and for
+      that target TrueDN = 0.
+    </p>
   </description>
 
   <history>
--- a/isis/src/lro/apps/lronaccal/lronaccal.xml
+++ b/isis/src/lro/apps/lronaccal/lronaccal.xml
@@ -8,41 +8,41 @@
 
   <description>
     <p>
-    lronaccal performs radiometric corrections to images acquired by the Narrow Angle
-    Camera aboard the Lunar Reconnaissance Orbiter spacecraft.
+      lronaccal performs radiometric corrections to images acquired by the Narrow Angle
+      Camera aboard the Lunar Reconnaissance Orbiter spacecraft.
     </p>
 
     <p>
-    The LRO NAC detector has a total of 5064 pixels, divided among an A
-    channel and a B channel.  The pixels alternate between the two channels:
-    ABABABAB, etc.  Images from LROC NAC may or may not include all pixels in the
-    acquired image.  There are special summing modes that are utilized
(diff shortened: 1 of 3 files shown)


isis/src/lro/apps/lronaccal/lronaccal.xml
```diff
@@ -46,6 +45,33 @@
...
+      image. In calibration, the mean bias and dark current are subtracted, but the random 
+      noise cannot be subtracted, because only the statistical distribution of the random 
+      noise can be known in advance, not the specific value for that pixel for that image. 
+      Because the random noise values are both positive and negative, when the true signal 
+      is zero the calibrated image should have both positive and negative values. If it had 
+      only positive values, the mean would be positive and therefore it would have a 
+      systematic error, since the true signal is zero. It is not possible to eliminate the 
+      random noise entirely, but it is defined as both additive and subtractive to the signal 
+      to minimize the systematic error in the calibrated image. That forces calibrated dark 
+      images to have both positive and negative values.
```
jlaura: I am concerned that this paragraph is not very clear. From what I am reading, the random noise is additive, with a mean of zero (and some variance, σ^2). Since the variance is not equal to zero, a sample taken from what sounds like a normal distribution can have a negative value. Is that what the above is saying?

Then the mean bias (MB) and dark current (DC) are additive as well, though they can have negative values? So, if the DN = 0, should we see:

DB + MB + DC = N(μ, σ^2), where N(μ, σ^2) is the error as sampled from a normal distribution.

Right now, the additive vs. subtractive seems to muddy the conversation. It seems that all the distinction does is flip the sign on the error term.

isis/src/lro/apps/lrowaccal/lrowaccal.xml
```diff
@@ -149,6 +149,34 @@
...
+      noise can be known in advance, not the specific value for that pixel for that image. 
+      Because the random noise values are both positive and negative, when the true signal 
+      is zero the calibrated image should have both positive and negative values. If it had 
+      only positive values, the mean would be positive and therefore it would have a 
+      systematic error, since the true signal is zero. It is not possible to eliminate the 
+      random noise entirely, but it is defined as both additive and subtractive to the signal 
+      to minimize the systematic error in the calibrated image. That forces calibrated dark 
+      images to have both positive and negative values.
+    </p>
+
```
jlaura, line 199: Same comment as above. This looks like a straight copy/paste?

jlaura (changes requested):

isis/src/lro/apps/lrowaccal/lrowaccal.xml
scsides: It is a copy. The exact same process is done for both instruments, and both have negative I/F DNs in dark places.

isis/src/lro/apps/lronaccal/lronaccal.xml
scsides: > I am concerned that this paragraph is not very clear. From what I am reading, the random noise is additive, with a mean of zero (and some variance, σ^2). Since the variance is not equal to zero, a sample taken from what sounds like a normal distribution can have a negative value. Is that what the above is saying?

Yes, that is what it is saying. Is it the last sentence of the first paragraph that is confusing? I believe Dave is restating the previous sentence in a different way. Trying to emphasize that zero +- the noise results in negative output when the noise is negative. Not that it makes a difference, but the noise is modeled as a poisson distribution.
> 
> Then the mean bias (MB) and dark current (DC) are additive as well, though they can have negative values? So, if the DN = 0, should we see:
> 
> DB + MB + DC = N(μ, σ^2), where N(μ, σ^2) is the error as sampled from a normal distribution.
> 
> Right now, the additive vs. subtractive seems to muddy the conversation. It seems that all the distinction does is flip the sign on the error term.

jlaura: It is these sentences in combination:

`The random noise is both additive and subtractive to the signal by definition.`
and
` In calibration, the mean bias and dark current are subtracted, but the random 
       noise cannot be subtracted, because only the statistical distribution of the random 
       noise can be known in advance, not the specific value for that pixel for that image.`

The idea that the values are added and subtracted does not make sense. I would guess that the values are added and that some of them can be negative. Can we just say that? If not, what are the rules are used to determine whether the sampled random error is added or subtracted. 

It also does not make sense that the 'the random noise cannot be subtracted' because the statistical distribution can be known. These two are completely orthogonal to one another as far as I know. Some value is sampled from a distribution. Literally anything could be done with that value.

isis/src/lro/apps/lronaccal/lronaccal.xml
```diff
@@ -8,42 +8,67 @@
...
+      scene, bias, and dark current with no systematic error. That implies the statistical 
+      distribution of the random noise has an average of zero, and therefore the random noise
+      has both positive and negative values, except for the trivial case of zero random noise.
+    </p>
+
+    <p>
+      Let's look at the case of a calibrated image for which the true signal is zero, a dark 
+      image. In calibration, the mean bias and dark current are subtracted, but the random 
+      noise cannot be subtracted, because only the statistical distribution of the random 
+      noise can be known in advance, not the specific value for that pixel for that image.
```
jlaura: `In calibration, the mean bias and dark current are subtracted, but the random 
       noise cannot be subtracted, because only the statistical distribution of the random 
       noise can be known in advance, not the specific value for that pixel for that image.` - This sentence still does not make sense.

Consider the following for the entire paragraph:
`Let's look at the case of a calibrated image for which the true signal is zero, a dark image. In calibration, the mean bias and dark current are subtracted. The random noise term is then randomly sampled from a known distribution with a mean of zero. Since the distribution has a mean of zero, values for the random noise can be positive or negative. Therefore, the addition of random noise to a pixel with true signal near zero can result in negative DN values.`

I would also note that this discussion would be significantly improved with the inclusion of a simple math equation. Something along the lines of:

reportedDN = ObservedDN - MeanBias - DarkCurrent + E, where E is a randomly sampled value from (mu, sigma^2) and mu=0.

Negative reported DNs are possible when E < ObserveredDN - MeanBias - DarkCurrent.

jlaura (approved):

scsides:
Had another pair of eyes look at this, and agreed the equations are ok.

PR #4128, opened 2020-11-18
added socetlinescankeywords tests

Kelvinrr:
## Description

After looking through the original tests I couldn't find a reason to have other instrument specific tests. So it's basically a default and error test. 

## Motivation and Context

1.8Gb

## How Has This Been Tested?

local tests passing

## Types of changes

## Checklist:

## Licensing


Kelvinrr:
Talking to @scsides about the other tests and his thoughts on how necessary they actually are, seems like this app is a candidate for depreciation? If so, should this even be merged?

isis/tests/Fixtures.cpp
```diff
@@ -475,15 +474,17 @@ namespace Isis {
...
-    nk["INS"+ikid.toStdString()+"_ITRANSS"] = {-1000.86, -0.0087, -83.333};
-    nk["INS"+ikid.toStdString()+"_ITRANSL"] = {7457.9, 83.3333, -0.0087};
+    nk["INS-74605_TRANSX"] = {-89.496, -1.0e-06, 0.012};
+    nk["INS-74605_TRANSY"] = {-12.001, -0.012, -1.0e-06};
+    nk["INS-74605_ITRANSS"] = {-1000.86, -0.0087, -83.333};
+    nk["INS-74605_ITRANSL"] = {7457.9, 83.3333, -0.0087};
     nk["INS"+ikid.toStdString()+"_OD_K"] = {-0.0048509, 2.41312e-07, -1.62369e-13};
     nk["BODY499_RADII"] = {3396.19, 3396.19, 3376.2};
-    nk["CLOCK_ET_-74999_847428572:52459_COMPUTED"] = "8ed6ae8930f3bd41";
+    nk["CLOCK_ET_-74999_895484264:57342_COMPUTED"] = "8ed6ae8930f3bd41";
```
acpaquette: This seems odd? Does this not break other tests that use the setInstrument method for the mro cube? Or is it just left over changes that should be reverted?

isis/tests/FunctionalTestsSocetLineScanKeywords.cpp
```diff
@@ -0,0 +1,153 @@
...
+using namespace testing; 
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/socetlinescankeywords.xml").expanded();
+
+TEST_F(MroCube, FunctionalTestSocetLineScanKeywordsHirise) {
+  setInstrument("-74999", "HIRISE", "MARS RECONNAISSANCE ORBITER");
+  
+  QTemporaryDir prefix;
+  testCube->reopen("rw");  
+  QString outFileName ="/tmp/outTEMP.txt";
```
acpaquette: Remove prefix variable and use the tempDir from the fixture in the outFileName. Looks like this needs to happen in all tests, unless there is something I am missing.

acpaquette (commented):
One question and a couple things to address in the tests.

isis/tests/Fixtures.cpp
Kelvinrr: I was gonna leave to Jenkins to tell me this, but the impression I have is that the original code was somewhat busted and it would essential re-spiceinit the image but skipping the table logic and hitting the Spice kernels because of this error and this test simply exposed the issue since it actually needing these variables. Does kinda break the set instrument paradigm so I might change this to it's own CTX fixture.

isis/tests/FunctionalTestsSocetLineScanKeywords.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "gmock/gmock.h"
+
+using namespace Isis;
+using namespace testing; 
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/socetlinescankeywords.xml").expanded();
+
+TEST_F(MroCtxCube, FunctionalTestSocetLineScanKeywordsHirise) {
+  
+  QTemporaryDir prefix;
```
acpaquette: Still unsure why we need another temporary directory here. The inheritance chain for the MroCtxCube should also inherit the tempDir temporary directory, so we shouldn't need to define antoher temporary directory.

Kelvinrr: I think there was a trend of tests having there own prefix but I do not know where that originated from so this was me following the formula. I can try without and see if it breaks anything. 

My memory tells me there was a reason we wanted to separate the input tempDir from output tempDir 🤔

acpaquette: Hmmmm, weird. I don't really know why we would want/need separate input and output directories?

Kelvinrr: Nothing failed locally, so I think it's fine 🤷‍♂️

acpaquette (approved):

PR #4148, opened 2020-11-30
CSM test plugin and one test sensor model.

krlberry:
## Description
Provides just the test csm plugin and model classes for testing csm in ISIS. 

This might be a bit less "clean" than I initially thought it would be because I'd started (but not finished) modifying the plugin to handle a second sensor model and not just return "true" for everything. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/TestCsmModel.h
+++ b/isis/tests/TestCsmModel.h
@@ -1,0 +1,34 @@
+#ifndef TestCsmModel_h
+#define TestCsmModel_h
+
+#include <string>
+
+#include "csm/Plugin.h"
+#include "csm/Model.h"
+#include "csm/Version.h"
+
+#include <nlohmann/json.hpp>
+
+class TestCsmModel : public csm::Model {
+ public:
+   TestCsmModel();
+   ~TestCsmModel();
+
+   std::string getFamily() const;
+   csm::Version getVersion() const;
+   std::string getModelName() const;
+   std::string getPedigree() const;
+   std::string getImageIdentifier() const;
+   void setImageIdentifier(const std::string& imageId,
+                           csm::WarningList* warnings = NULL);
+   std::string getSensorIdentifier() const;
+   std::string getPlatformIdentifier() const; 
+   std::string getCollectionIdentifier() const;
+   std::string getTrajectoryIdentifier() const;
+   std::string getSensorType() const;
+   std::string getSensorMode() const;
+   std::string getReferenceDateAndTime() const;
+   std::string getModelState() const;
+   void replaceModelState(const std::string& argState);
+};
+#endif
--- a/isis/tests/TestCsmPlugin.cpp
+++ b/isis/tests/TestCsmPlugin.cpp
@@ -1,0 +1,126 @@
+#include <iostream>
+
+#include "TestCsmPlugin.h"
+#include "TestCsmModel.h"
+
+// Static Instance of itself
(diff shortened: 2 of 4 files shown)


jessemapel (approved):

isis/tests/TestCsmModel.h
```diff
@@ -0,0 +1,34 @@
+#ifndef TestCsmModel_h
+#define TestCsmModel_h
+
+#include <string>
+
+#include "csm/Plugin.h"
+#include "csm/Model.h"
+#include "csm/Version.h"
+
+#include <nlohmann/json.hpp>
```
jessemapel, line 10: I don't think this is needed

isis/tests/TestCsmPlugin.cpp
```diff
@@ -0,0 +1,126 @@
...
+    return static_cast<bool>(model);
+  }
+  catch (std::exception &e) {
+    std::string msg = "Could not create model [";
+    msg += modelName;
+    msg += "] with error [";
+    msg += e.what();
+    msg += "]";
+    std::cout << msg << std::endl;
+  }
```
jessemapel, line 61: We shouldn't do this try/catch logic similar to what we have in usgscsm. I would instead check something in the ISD like the filename and then decide off that. So, if the filename matches "testisd" or something, then it can create the first model, if it matches "testisd2" then it can create the second model. This helps ensure that no tests accidentally use this plugin and model. Right now, this will always return True, which could be problematic.

isis/tests/TestCsmPlugin.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+csm::Version TestCsmPlugin::getModelVersion(const std::string& modelName) const { 
+  return csm::Version(3,0,3);
+}
+
+bool TestCsmPlugin::canModelBeConstructedFromState(const std::string& modelName,
+                                    const std::string& modelState,
+                                    csm::WarningList* warnings) const {
+  try {
+    csm::Model *model = constructModelFromState(modelState, warnings);
```
jessemapel, line 51: This is a memory leak, model is never deleted.

krlberry: You're right, thank you.

krlberry: Is this a memory leak in usgscsm, too?

isis/tests/TestCsmPlugin.cpp
krlberry: Yeah, this was one of the things that was "in progress" for modification to use the second model! I ended up doing something like this.

isis/tests/TestCsmPlugin.cpp
jessemapel: It would be, we should check that.

isis/tests/TestCsmPlugin.cpp
jessemapel: Okay, that's what I figured 👍

isis/tests/TestCsmPlugin.cpp
krlberry: As it turns out, if you delete the plugin pointer, but don't remove it from the plugin list, it causes a segfault

jessemapel: This isn't a plugin pointer though, this is a dangling model pointer?

PR #4190, opened 2020-12-09
Kaguyami App and Test Update

acpaquette:
App conversion for kaguyami2isis and test conversion to gtests

## Description
Updates the kaguyami2isis app format and converts all existing make tests into gtests.

## Related Issue
N/A

## Motivation and Context
For the push to convert apps and gtests.

## How Has This Been Tested?
These are tests.

## Types of changes

## Checklist:

## Licensing


isis/src/kaguya/apps/kaguyami2isis/kaguyami2isis.cpp
```diff
@@ -0,0 +1,159 @@
...
+        kern += PvlKeyword("NaifFrameCode", toString(-13133) + baseBand[2]);
+      }
+      kern += PvlKeyword("NaifCkCode", toString(-131330));
+    }
+    else if (lab.findKeyword("INSTRUMENT_ID")[0] == "MI-NIR") {
+      if (baseBand.contains("MN")) {
+        kern += PvlKeyword("NaifFrameCode", toString(-13134) + baseBand[2]);
+      }
+      kern += PvlKeyword("NaifCkCode", toString(-131340));
+    }
```
acpaquette, line 124: Significant change made here for how the `NaifFrameCode`  is generated. Changed from 5 part if/else blocks to single if that extracts the value of the baseBand number. For `MI-VIS` and `MI-NIR` the baseBand defines the `NaifFrameCode` where the digit in the baseBand is the last digit in the frame code.

Kelvinrr: Old tests pass just fine I assume?

isis/tests/FunctionalTestsKaguyami2isis.cpp
```diff
@@ -0,0 +1,388 @@
...
+  EXPECT_DOUBLE_EQ(hist->Average(), 29.395262812500022);
+  EXPECT_DOUBLE_EQ(hist->Sum(), 188129.68200000015);
+  EXPECT_EQ(hist->ValidPixels(), 6400);
+  EXPECT_DOUBLE_EQ(hist->StandardDeviation(), 2.8449125231835715);
+}
+
+TEST(kaguyatc2isisTest, FunctionalTestsKaguyami2isisProj) {
+  QTemporaryDir prefix;
+  QString cubeFileName = prefix.path() + "/kaguyatc2isisTEMP.cub";
+  QVector<QString> args = {"from=data/kaguyami2isis/MVA_2B2_01_01229N603E1343_cropped", "to="+cubeFileName};
```
Kelvinrr: the Vis tests is also a Vis 2b2 image, is this test redundant or am I missing something?

Kelvinrr (commented):
just some questions, otherwise looks good.

isis/src/kaguya/apps/kaguyami2isis/kaguyami2isis.cpp
acpaquette: Yes, the original tests were passing after this change

isis/tests/FunctionalTestsKaguyami2isis.cpp
acpaquette: No, that is a fair point. The original default test for the mi camera tested three separate images and this was one of them. To be honest I am not sure why this image was special compared to the NIR and VIS images that were also being tested.

Kelvinrr: It seems like the projected test is supposed to be 3C5 file, not 2B2, and it is supposed to fail as an unsupported file type. It seems like the thing to do is replace the proj test with a failure check without a cropped image and instead write a PVL blob to disk with a PVL group named `IMAGE_MAP_PROJECTION` to trigger the error. 

As far as I can see, there is no difference between the 2b2 tests in default/2b2, so I think the rest of the tests are okay.

acpaquette: I pulled the label from the projected cube that was in the projected test. Since it's a pds label and not an ISIS label I am not sure about the mechanics off adding to the label in the test rather than just using an existing label. I could spend some time trying to figure that out but this seemed like the faster solution

Kelvinrr: that's fine enough, I think the thing to do was create a PVL object from a string literal and write that to disk in the test itself.

Kelvinrr:
If tests come back green I think this is good, unfortunately things are backed up at the moment.

Kelvinrr (approved):
looks like test are almost done, whoever checks it first can merge this.

PR #4191, opened 2020-12-10
Csm branch clean up

jessemapel:
- improved csminit docs
 - removed csmpt
 - added a changelog entry
 - removed debug couts
 - tweaked tests to remove more hardcoded names


CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ update the Unreleased link so that it compares against the latest release tag.
 
 ## [Unreleased]
 
+### Added
+
+ - Added thew new csminit application and CSM Library loading to the IsisPreferences file. Together these allow users to get CSM state strings from ISD files. Once CSM camera model support is added, these will be used to setup a Cube to use a CSM camera model.
```
krlberry: thew -> the

isis/src/base/apps/csminit/csminit.xml
```diff
@@ -2,17 +2,32 @@
...
 
   <description>
     <p>
-      This program takes an ISD or state string and attempts to create
-      a valid CSM Model from the loaded CSM Libraries. Then, it attaches the
-      state string from the CSM Model to the cube for later use. If there are
-      multiple valid CSM Models that can be instantiated, then this program
-      will error and the user will have to re-run it with a specific model
-      and/or plugin name.
+      This program takes an ISD and attempts to create a valid CSM model from
```
krlberry: Expand first instance of potentially user-unfamiliar acronym "ISD" like "Instrument Support Data (ISD)" Potentially do the same with CSM!

isis/src/base/apps/csminit/csminit.xml
```diff
@@ -2,17 +2,32 @@
...
+      the loaded CSM Libraries. If a single model can be created, its state
+      string is attached to the cube in the CSMState BLOB for later use. If
+      there are multiple valid CSM Models that can be instantiated, then this
+      program will error and return a list of possible models. The user will
+      then have to re-run it using the PLUGINNAME and/or MODELNAME parameters
+      to select a specific model.
+    </p>
+    <p>
+      Various pieces of information about the CSM model will also be added to
```
krlberry: I would eliminate "various pieces of" and just say "Information about the CSM model..." Additionally, I don't think this is necessary at this time, but potentially in the future when we get the CSM camera models working with ISIS and expect this app to be getting active use, it might be nice to have an example `CsmInfo` group as part of the docs, or add an example which includes the output CsmInfo group.

krlberry (commented):
Looks good! One typo and a couple of documentation suggestions. Code looks good.

isis/src/base/apps/csminit/csminit.xml
jessemapel: I think a more robust example would be good here. It would allow us to also provide an example of how to get a CSM Library loaded. I don't want to get into that right now, but it may be a good option to create one from a demo later on.

krlberry (approved):
:+1:

PR #4197, opened 2020-12-11
Added fix to half pixel issue in Camera.cpp - ISIS Issue 4018 

victoronline:
In the detector map, the cast to an integer is rounding:

int framelet = (int)((line - 0.5) / actualFrameletHeight) + 1;

In the pixels on the edge of the framelet, this is rounding incorrectly and mapping to the next/previous framelet.

ISIS [issue 4018]


victoronline:
Jumped the gun a bit. Will resubmit.

PR #4198, opened 2020-12-14
Isis2pds App and Gtests

amystamile-usgs:
Converted isis2pds app and added gtests.

## Description
Converted isis2pds/main.cpp to isis2pds/isis2pds.cpp and isis2pds/isis2pds.h. Added 10 gtests.

## Related Issue
Test conversion.

## Motivation and Context
Frees 120MBs.

## How Has This Been Tested?
10 gtests.

## Types of changes

## Checklist:

## Licensing


amystamile-usgs:
I was not sure on the unit conversion Makefile test. Looks like it was testing a cube with no units and a cube with units per the file names. But when comparing the two cubes, they were the same. Please clarify the unit conversion and I can add an additional test if needed.

isis/src/base/apps/isis2pds/isis2pds.cpp
```diff
@@ -0,0 +1,244 @@
...
+  void setRangeAndPixels(UserInterface &ui, ProcessExport &p,
+                         double &min, double &max, Pixtype ptype);
+  void isis2pds(UserInterface &ui, Pvl *log) {
+
+    // Check if input file is indeed, a cube
+    if (ui.GetFileName("FROM").right(3) != "cub") {
+      QString msg = "Input file [" + ui.GetFileName("FROM") +
+                  "] does not appear to be a cube";
+      throw  IException(IException::User, msg, _FILEINFO_);
+    }
```
kaitlyndlee: I know you didn't add these lines, but I think they are unreachable. If I remember correctly, the Cube class should throw an error if the input isn't a cube. Is this right @scsides?

scsides: I understand what you mean, but the code is reachable because the Aml does not check the file name or open it and make sure it is a cube prior to calling IsisMain.

Along the same lines, it is an odd test. Why not just wait until the file is opened with "open" or "setInputCube"? Either of those will fail gracefully and accurately if it is not a cube. It may be because isis2std was running too long before failing, or possibly creating the output PDS IMG. In either case, a file ending in .cub does not a cube make.

kaitlyndlee: I see what you mean, thank you. It looks like in the cube class, it will try to append the cub extension to the input file. With these lines in isis2pds, we remove that additional functionality in the cube class.

amystamile-usgs: So it sounds like overall this code block is not necessary?

isis/tests/FunctionalTestsIsis2pds.cpp
```diff
@@ -0,0 +1,435 @@
...
+  EXPECT_EQ(int(imageObject["CORE_LOW_INSTR_SATURATION"]), 2);
+  EXPECT_EQ(int(imageObject["CORE_HIGH_REPR_SATURATION"]), 65535);
+  EXPECT_EQ(int(imageObject["CORE_HIGH_INSTR_SATURATION"]), 65534);
+
+}
+
+TEST(isis2pdsTest, FunctionalTestIsis2pdsOffset) {
+  QTemporaryDir prefix;
+  QString outputFile = prefix.path() + "/outTemp.img";
+  QVector<QString> args = {"from=data/isis2pds/ldem_4_reduced.cub",
```
kaitlyndlee, line 330: I might've missed a discussion about this at the standups, but do we want to programmatically create the input cubes with a fixture or add these cub files to the test data like you are doing now?

isis/tests/FunctionalTestsIsis2pds.cpp
```diff
@@ -0,0 +1,435 @@
...
+  EXPECT_DOUBLE_EQ(double(projectionObject["WESTERNMOST_LONGITUDE"]), 0.0);
+  EXPECT_DOUBLE_EQ(double(projectionObject["LINE_PROJECTION_OFFSET"]), 119.5);
+  EXPECT_DOUBLE_EQ(double(projectionObject["SAMPLE_PROJECTION_OFFSET"]), 239.5);
+
+}
+
+TEST(isis2pdsTest, FunctionalTestIsis2pdsPds4) {
+  QTemporaryDir prefix;
+  QString outputFile = prefix.path() + "/outTemp.img";
+  QVector<QString> args = {"from=data/isis2pds/dawnEqui1.cub",
```
kaitlyndlee, line 401: Same question as above here.

isis/tests/FunctionalTestsIsis2pds.cpp
amystamile-usgs: My understanding from being trained is that if the fixtures do not handle the specific instances that we can alternatively reduce the image that was originally being tested with. For this, it mentioned testing an offset but I was not sure how to test offset with a fixture cube. Therefore I just reduced the original tested cube.

isis/src/base/apps/isis2pds/isis2pds.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include "isis2pds.h"
+
+using namespace std;
+namespace Isis{
+
+  enum Pixtype { NONE, NEG, BOTH };
+
+  void setRangeAndPixels(UserInterface &ui, ProcessExport &p,
+                         double &min, double &max, Pixtype ptype);
+  void isis2pds(UserInterface &ui, Pvl *log) {
```
kaitlyndlee: We should add a second isis2pds function with a cube parameter. Follow the steps on the [testing page](https://github.com/USGS-Astrogeology/ISIS3/wiki/Writing-ISIS3-Tests-Using-Gtest-and-Ctest#creating-a-basic-callable-function) or check out the [grid app](https://github.com/USGS-Astrogeology/ISIS3/blob/1fc23d07df1539d73a3aa118d748bf1d3f6ec67c/isis/src/base/apps/grid/grid.cpp#L32) for an example.

isis/tests/FunctionalTestsIsis2pds.cpp
amystamile-usgs: This one I could probably use a fixture cube instead. Good call.

amystamile-usgs: Ah. I remember the reason I switched to the cub file. When using the fixture, It gives me an error "Unable to find a target in input cube". When I reverted to the cub file, it no longer gave me the error.

isis/tests/FunctionalTestsIsis2pds.cpp
kaitlyndlee: I think the offset is referring to the offset keyword in the output image. You could create a new fixture in a similar way to the [Apollo cube](https://github.com/USGS-Astrogeology/ISIS3/blob/936e6fcfa91ace80414c7ac89be1fecf5c52e163/isis/tests/Fixtures.cpp#L706). @Kelvinrr What are your thoughts?

isis/src/base/apps/isis2pds/isis2pds.cpp
amystamile-usgs: Push updated code with added function. The only potential issue that I came across when comparing my test suite with the grid tests is that when I try calling isis2pds(testCube, options), it will interchangeably say segfault or child aborted. Tests pass if I keep the call as isis2pds(options). In order to prevent the segfault within isis2pds.cpp, I have to revert the code p.SetInputCube(icube) back to p.SetInputCube(ui.GetFileName("FROM"), inputAtt).

isis/tests/FunctionalTestsIsis2pds.cpp
amystamile-usgs: I am still having a difficulty finding the correlation between the output offset keywords from the original cub file. It does not seem like the data is being directly pulled from the cube label, therefore I am not sure how to add the correct information to a new fixture.

AustinSanders: We've pushed small .cubs to the repo for hicubeit, demprep, and ringsautomos.  We've also programmatically created cubes with pvl/isds in footprintinit, apolloNetwork, etc.  It seems like there's no clear policy related to this issue, so I'm of the opinion that this is not a merge-blocking issue.

I think that this is something that we should discuss during standup.

kaitlyndlee: I agree. There seems to be multiple ways of testing and we should come up with a standard.

isis/src/base/apps/isis2pds/isis2pds.cpp
```diff
@@ -0,0 +1,246 @@
...
+  enum Pixtype { NONE, NEG, BOTH };
+
+  void setRangeAndPixels(UserInterface &ui, ProcessExport &p,
+                         double &min, double &max, Pixtype ptype);
+
+  void isis2pds(UserInterface &ui, Pvl *log){
+    Process p;
+
+    CubeAttributeInput cai;
+    Cube *icube = p.SetInputCube(ui.GetFileName("FROM"), cai, ReadWrite);
```
kaitlyndlee: You should copy over the `setVirtualBands()` call from another converted app. This makes it so that the newly created cube `icube` will have the correct number of bands when the user runs the program from the command line. We also don't need to call `SetInputCube` and `endProcess` here since  they are called in the second isis2pds function.

isis/src/base/apps/isis2pds/isis2pds.cpp
kaitlyndlee: I think this is because the cube isn't "open" (function [here](https://github.com/USGS-Astrogeology/ISIS3/blob/87c8432b2c21f60610204b41a378f5c24464b6ca/isis/src/base/objs/Cube/Cube.cpp#L636)) before trying to set it as the input. To be able to read from a cube, it firsts needs to be opened with the `open` function. The `p.SetInputCube(ui.GetFileName("FROM"), inputAtt)` call opens the cube for you.

amystamile-usgs: When I try opening the cube before setting as input, I get all failed tests stating the the cube is already open.

amystamile-usgs: Discussed this issue in standup. I will look into this more today regarding what is causing the segfault. Since I will not be working the rest of the week, we decided that if it is not resolved by the end of today, to just revert it back to not having a function with the cube parameter.

amystamile-usgs: Resolved segfault by switching tests from SmallCube to DefaultCube.

AustinSanders (approved):
I think this is good to go.  I'm still curious as to why changing from small cube to default cube resolved the segfaults.

Issue #4018, opened 2020-09-14
Camera.cpp is not correctly calculating the local normal.


AaronBoyd:
The following code in Camera.cpp is not working as intended:
For:
`void Camera::GetLocalNormal(double normal[3]) {`
The following lines are not working as intended:
```
      QList< QPair< double, double > > surroundingPoints;
      surroundingPoints.append(qMakePair(samp, line - 0.5));
      surroundingPoints.append(qMakePair(samp, line + 0.5 - DBL_MIN));
      surroundingPoints.append(qMakePair(samp - 0.5, line));
      surroundingPoints.append(qMakePair(samp + 0.5 - DBL_MIN, line));
```

Currently the DBL_MIN is not operating as expected. The functionality that is intended is:
Calculate normal by grabbing elevation values at the edges of the pixel in question.

What is actually happening:
The values gathered are still not from the same pixel. It appears as though the DBL_MIN is not subtracting more than the rounding error, and this is essentially returning the same value as if the code was:
```
      QList< QPair< double, double > > surroundingPoints;
      surroundingPoints.append(qMakePair(samp, line - 0.5));
      surroundingPoints.append(qMakePair(samp, line + 0.5));
      surroundingPoints.append(qMakePair(samp - 0.5, line));
      surroundingPoints.append(qMakePair(samp + 0.5, line));
```

This is detrimental when calculating photometric angles for push-broom and push-frame cameras, because 1.0 could be the pixel center, 0.5 is the edge of pixel 1, but 1.5 is the edge of pixel 2. This does not work well when a push-broom imager is flying in a reverse direction, or for frame boundaries of a push-frame camera. The calculation is using a completely wrong location on the body.

This is a high priority for almost all missions currently using ISIS.

jessemapel:
@AaronBoyd Do you have a good example to reproduce this or is this happening in all images?

AaronBoyd:
This happens noticeably in every push-frame image, but here are a couple where the photometric angle difference can be seen at the frame boundaries:
http://wms.lroc.asu.edu/lroc/view_lroc/LRO-L-LROC-2-EDR-V1.0/M1099201504ME
http://wms.lroc.asu.edu/lroc/view_lroc/LRO-L-LROC-2-EDR-V1.0/M1229850147ME


lrowac2isis from= image.IMG to=image.cub
spiceinit from= image.vis.even.cub 
[...]

PR #4214, opened 2020-12-18
Added fix to Camera.cpp not correctly calculating the local norm 

victoronline:
Added fix to half pixel issue in Camera.cpp - ISIS Issue [#4018](https://github.com/USGS-Astrogeology/ISIS3/issues/4018), whereas DBL_MIN is not operating as expected, causing an error later in the code, when integer rounding occurs. This erroneously returns the next/previous framelet. 

## Description
Currently the DBL_MIN is not operating as expected. The functionality that is intended is:
Calculate normal by grabbing elevation values at the edges of the pixel in question.

What is actually happening:
The values gathered are still not from the same pixel. It appears as though the DBL_MIN is not subtracting more than the rounding error, and this is essentially returning the same value as if the code was:

      QList< QPair< double, double > > surroundingPoints;
      surroundingPoints.append(qMakePair(samp, line - 0.5));
      surroundingPoints.append(qMakePair(samp, line + 0.5));
      surroundingPoints.append(qMakePair(samp - 0.5, line));
      surroundingPoints.append(qMakePair(samp + 0.5, line));

This change was made to version 3.10.2. 

## Motivation and Context
This is detrimental when calculating photometric angles for push-broom and push-frame cameras, because 1.0 could be the pixel center, 0.5 is the edge of pixel 1, but 1.5 is the edge of pixel 2. This does not work well when a push-broom imager is flying in a reverse direction, or for frame boundaries of a push-frame camera. The calculation is using a completely wrong location on the body.

## How Has This Been Tested?
The LROC team has tested the fix. Will provide details in comments below but a test (GTest) has not been created for this yet.

Please see issue. Several screenshots and examples have been included.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/Camera/Camera.cpp
+++ b/isis/src/base/objs/Camera/Camera.cpp
@@ -1540,9 +1540,10 @@ namespace Isis {
       // order of points in vector is top, bottom, left, right
       QList< QPair< double, double > > surroundingPoints;
       surroundingPoints.append(qMakePair(samp, line - 0.5));
-      surroundingPoints.append(qMakePair(samp, line + 0.5 - DBL_MIN));
+      //surroundingPoints.append(qMakePair(samp, line + 0.5 - DBL_MIN));
+			surroundingPoints.append(qMakePair(samp, std::nexttoward(line + 0.5 - DBL_MIN)));
       surroundingPoints.append(qMakePair(samp - 0.5, line));
-      surroundingPoints.append(qMakePair(samp + 0.5 - DBL_MIN, line));
+      surroundingPoints.append(qMakePair(std::nexttoward(samp + 0.5 - DBL_MIN, line)));
 
       // save input state to be restored on return
       double originalSample = samp;
--- a/isis/src/base/objs/Camera/Camera.cpp
+++ b/isis/src/base/objs/Camera/Camera.cpp
@@ -1540,9 +1540,9 @@ namespace Isis {
       // order of points in vector is top, bottom, left, right
       QList< QPair< double, double > > surroundingPoints;
       surroundingPoints.append(qMakePair(samp, line - 0.5));
-			surroundingPoints.append(qMakePair(samp, std::nexttoward(line + 0.5 - DBL_MIN)));
+			surroundingPoints.append(qMakePair(samp, std::nexttoward(line + 0.5, line)));
       surroundingPoints.append(qMakePair(samp - 0.5, line));
-      surroundingPoints.append(qMakePair(std::nexttoward(samp + 0.5 - DBL_MIN, line)));
+      surroundingPoints.append(qMakePair(std::nexttoward(samp + 0.5, samp), line));
 
       // save input state to be restored on return
       double originalSample = samp;
--- a/isis/src/base/objs/Camera/Camera.cpp
+++ b/isis/src/base/objs/Camera/Camera.cpp
@@ -1539,9 +1539,9 @@ namespace Isis {
 
       // order of points in vector is top, bottom, left, right
       QList< QPair< double, double > > surroundingPoints;
-      surroundingPoints.append(qMakePair(samp, line - 0.5));
-			surroundingPoints.append(qMakePair(samp, std::nexttoward(line + 0.5, line)));
-      surroundingPoints.append(qMakePair(samp - 0.5, line));
+      surroundingPoints.append(qMakePair(samp, std::nexttoward(line - 0.5, line)));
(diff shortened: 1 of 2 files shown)


isis/src/base/objs/Camera/Camera.cpp
```diff
@@ -1540,9 +1540,9 @@ namespace Isis {
       // order of points in vector is top, bottom, left, right
       QList< QPair< double, double > > surroundingPoints;
       surroundingPoints.append(qMakePair(samp, line - 0.5));
-      surroundingPoints.append(qMakePair(samp, line + 0.5 - DBL_MIN));
+			surroundingPoints.append(qMakePair(samp, std::nexttoward(line + 0.5, line)));
```
jessemapel: Looks like a weird spacing issue is happening here. Check your tabs vs spaces in your editor.

acpaquette:
@victoronline This should be picked up by a unit test so I don't think a gtest will be necessary. I'm concerned that this will have a pretty big impact since this is a change to the base camera

jessemapel (commented):
Change looks good. I'm trying to think of the best way to test this. We could make a camera model that just records its framelet and test this function on the edge of a framelet and then check which framelet it sets to.

jessemapel:
Fixes #4018

jlaura:
@victoronline Awesome to see the PR come in! 👍 🎉 

@Kelvinrr @AustinSanders @acpaquette @kaitlyndlee Can we get this slotted into the support sprint this week or next to assist getting a GTest written?

What is everyone thinking for getting this released post merge? We should be able to cherry pick this into dev after a merge into 3.10 and it will then be released during the normal release process next month (4.3.0_RC and then 4.3.0). Victor, are you all running this change on your own 3.10 custom build?

isis/src/base/objs/Camera/Camera.cpp
```diff
@@ -1540,9 +1540,9 @@ namespace Isis {
...
       surroundingPoints.append(qMakePair(samp - 0.5, line));
-      surroundingPoints.append(qMakePair(samp + 0.5 - DBL_MIN, line));
+      surroundingPoints.append(qMakePair(std::nexttoward(samp + 0.5, samp), line));
```
scsides, line 1545: These two changes make sense, and will fix the problem for phocube. If you want a slightly more accurate answer, you should adjust line 1542 and 1544 in a similar way as you did for 1543 and 1545. This will keep the requested pixel (line/sample) centered between the four surrounding pixels.

Although this does fix this function when phocube uses it with integer sample/line values, it does not fix it when other applications use non-integer sample/line values. We have created a new issue #4256 to address what we believe to be a fix for the other cases.

As far as testing this change goes: We do not believe it is reasonable to test this change under the current unittest environment. This would be much more reasonable once the camera test has been converted to gtest. Given that, this PR should be mergeable after addressing the current comments.

scsides (changes requested):

jessemapel:
@victoronline Any update on this? I'll echo Stuart and say that this is mergable into dev once his comments have been addressed.

#4256 can be addressed in a separate PR to fix this for non-integer pixel values.

isis/src/base/objs/Camera/Camera.cpp
victoronline: Working to implement changes. Thanks.

victoronline (commented):
Changes made.

jessemapel:
@victoronline 3.10 is not an active branch any more, can you change this PR to go into dev?

scsides (changes requested):
All the code changes look good.

The gtest submodule changes (266 files) needs to be fixed. We don't want to include them in this PR. Can you try to re-checkout gtest to revert these changes?

victoronline:
Since 3.10 is no longer an active branch, but the version we need changes made to, what does that mean if putting into dev? Will this change only be reflected in the latest ISIS version?

jessemapel:
Yes, this change will only be available in the newest release. You can also cherry-pick the PR from this commit back onto the 4.4 branch and then a 4.4.1 release can be pulled without some of the other new functionality, but we only go back 1 feature release right now.

victoronline:
Reopening due to removing local branch in error.

victoronline:
Submitted change to dev branch via PR [#4340](https://github.com/USGS-Astrogeology/ISIS3/pull/4340) since branch for this PR is no longer active (3.10.2). Closing this PR.

scsides (commented):
Woops, missed the camera.h history comment.

PR #4220, opened 2020-12-23
Jigsaw Target Radii Warning

### LINKED ISSUE #4132: Improved documentation for Jigsaw's Target Body solution
**Description**
The Jigsaw application includes the capability to solve for target body parameters for global networks (pole RA/Dec, orientation, average/triaxial radii). This option is discussed briefly on the ISIS applications page (https://isis.astrogeology.usgs.gov/Application/presentation/Tabbed/jigsaw/jigsaw.html), which also points to a template for a pvl input file. However, I can't find a description of how jigsaw actually solves for these parameters. This results in at least three problems. 1) It's unclear what values several parameters should be in the pvl (especially the "sigma" values). 2) It's unclear what the uncertainties output by jigsaw mean (see below). 3) It's difficult to report the output in the peer-reviewed literature because the application is essentially a "black box".

**Suggested Edit**
Without knowing what the code is doing, I can't suggest specifics; however, I would like to see the following documented somewhere. 1) At least a qualitative description of how target body parameters are solved for. Presumably it's minimizing the residuals between the tie point lat/lon/radii values and an arbitrary ellipse/sphere, but given that there are potentially 6 unknowns (pole Ra/Dec, 3 radii, Wo) it would be nice to know the specifics. It's also not clear how the sigma values in the pvl file are used. Do these constrain the solution to be near the initial values? This would help the user better understand what values to use (in some cases it seems to work better to let sigma=0). Finally, I'd like to see a description of what the "Final Accuracy" reported by jigsaw actually means. For our Enceladus shape model, the final accuracy reported on the triaxial radii was ~25 cm. This clearly isn't the uncertainty in the shape. Perhaps it's just a simple sqrt(N) calculation? Otherwise it may be a bug.

Thanks!

### LINKED ISSUE #4193: Jigsaw's solution for target body parameters (triaxial shape and mean radius) are inconsistent with radii triangulated during the bundle
**ISIS version(s) affected**: 3.10.* (haven't tried older/newer versions)

**Description**  
Brief Summary: 
When we solve for target body parameters (in this cases for a global Europa control network) in jigsaw the resulting triaxial shape (a, b, c radii) and the mean radius are inconsistent with the triangulated radii of the tie points derived during bundle adjustment of the same control network.

Specifics: 
Before solving for target body, we bundle adjust the final Europa control network (Galileo and Voyager images) including solving for radius. The result is a set of lat/lon/radius values for each of our 53,000+ points. See /work/projects/EuropaBasemap/lweller/GLL_VGR_Network/TargetBody_FinalSolutions/2020EUROPA_PM/RadAngTwist_GlobalMerged_2020_CilixFree_NewPM_bundleout_points.csv
This bundle has an updated prime meridian and the images have updated (though not quit final) SPICE. The image list can be found here: /work/projects/EuropaBasemap/lweller/GLL_VGR_Network/TargetBody_FinalSolutions/2020EUROPA_PM/
NewPM_GalileoVoyager_Europa_Merged_2020.lis 

These radii are "good." A quick and dirty dem shows that the point cloud forms a triaxial shape with the long axis near 180 deg, and the short axis aligned with the pole. Here I just shoved the point cloud through ASP point2dem. Nothing fancy. The cloud is very sparse in some areas, thus a lot of no-data. The point is simply that the cloud is triaxial (yellows high, green/light blue intermediate, dark blue low). 

image

An interrogation of the points themselves indicates that they are consistent with IAU values. Only ~30 points out of 53,000+ have corrections larger than 500 m. The largest point values cluster around 1562.6 km, which is the IAU/pck a-axis and are all near 180 deg. The smallest values cluster at 1559.5 km (again consistent with the IAU/pck) and are all at high latitudes, as expected. If we just take the mean (a terrible way to calculate the mean radius) we get 1560.99 km, which is pretty darn close to the IAU mean radius of 1560.8 km - especially considering a lot of our points happen to be near the longest axis (very few points near the poles). The largest radii is 1563.37 km, and this appears to be a bit of an outlier (the next largest is 1562.97 km). So a quick look at the data suggests the points are well behaved and consistent with IAU values. Finally the difference between a-axis and c-axis (which can be interpreted geophysically) is about 3.1 km. 

Now, we take the same network and use it to solve for the triaxial shape (one jigsaw run) and the mean radius (a second jigsaw run). These solutions can be found here: /work/projects/EuropaBasemap/lweller/GLL_VGR_Network/TargetBody_FinalSolutions/2020EUROPA_TRIRADIUS/ 
The results are radii that are significantly larger than IAU values: e.g., 1563.9 km, 1562.8 km, and 1561.6 km for a, b, and c respectively (using no constraints). Note that in the target body solution we can't actually solve for radii to have a direct comparison. A few red flags. 1) the value for the a-axis is larger than even the largest outlier from the points themselves (1563.9 vs 1563.4 km). 2) The mean radius of 1562.7 km is 1.7 km larger than the simple average of the points, and similar to the largest values in our cloud (i.e., the a-axis). Recall that our "average" is already biased toward points on the long axis. 3) the short axis of 1561.6 is 0.8 km longer then where most of the points on the c-axis seem to be. 4) a-c is only 2.3 km. So it isn't just a shift from the IAU, the relative length of the axis has changed.

@lwellerastro spent a lot of effort assessing whether the sigma values used made any difference. They do tweak the result, but not in a meaningful way. These results can be found in the dir mentioned above. Each "sigma" value has its own network and bundleout (labeling should be clear). The pvl files are also in that directory.

**Additional context**  
We used jigsaw to solve for target body parameters for Enceladus following an identical method. In that case the triaxial solution fall right in the radii point cloud from the bundle solution (I just double checked!). The values we found are consistent with the IAU values, and the mean radius we derived was reasonable.

**Possible Solution**  
It's not entirely clear that this is a bug. I'll note that there is essentially no documentation for this application (see https://github.com/USGS-Astrogeology/ISIS3/issues/4111) so it's possible that it is pilot error. However, at this point that seems unlikely to me. We've used this successfully before. It's also possible that the point cloud is just too sparse for whatever method is being used. If that's the case, the uncertainty values reported should be much larger (again, it's not clear how the uncertainty is determined). It's also possible that this is a bug (introduced between isis3.10 and isis3.8.0)

acpaquette:
Adds a warning about solving for the target body radii and updates the jigsaw docs.

## Description
Adds a new warning if the user is solving for the target body radii that it can negatively effect the bundle adjust. Updated the docs to describe how solving for the target body radii can cause problems and is likely redundant. Also updated the app xml to include the rand notebook that jigsaw was based on.

## Related Issue
Closes #4193
Closes #4132

## Motivation and Context
Jigsaws target body radii solving code has little to no documentation and the resulting radii output from a recent global Europa control network project are extremely fair off from their expected values. That said, there have been other projects in the past that have used this solve and gotten reasonable radii (maybe 100 - 500 meters off expected values).

## How Has This Been Tested?
Currently untested, but this warning could be checked in one of the radii tests.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/control/apps/jigsaw/jigsaw.xml
+++ b/isis/src/control/apps/jigsaw/jigsaw.xml
@@ -94,23 +94,26 @@
       </blockquote>
 
       <blockquote>
-      <b>Solving for the target body radii (triaxial or mean) is irrelavent and
+      <b>Solving for the target body radii (triaxial or mean) is NOT possible and
       likely increases error in the solve.</b>
       <p>
-        When solving for target body radii both the mean and triaxial solve options
-        are redundantly solving for the target body radii. The radii solve is part of
-        the sequence of equations that the <b>jigsaw</b> uses to compute various partial
-        derivatives to populate the solve matrix. The radii solve is being applied
-        to the target body separately and the partials from this separate application
-        are being added to the solve matrix. This is adding additional computation time
-        to <b>jigsaw</b> and creating additional uncertainty/error in the bundle adjust. We
-        advise this radii solve to be avoided.
+        With the current jigsaw implementation, it is NOT possible to individually
+        solve for either the mean or triaxial radii as seperate calculations in
+        the bundle adjustment. More specifically, the target body radii has no
+        effect when solving for individual points and thus cannot be solved for in the bundle.
+        A local radii solve is already part of the sequence of equations that <b>jigsaw</b>
+        uses to compute various partial derivatives to populate the solve matrix.
+        A seperate mean or triaxial radii solve can be applied to the target body
+        and the partials from this separate application are added to the
+        solve matrix. This option adds additional computation time to <b>jigsaw</b>
+        and creates additional uncertainty/error in the bundle adjust. We advise
+        the mean and triaxial radii solve be avoided.
       </p>
       <p>
-        If you are trying to generate a spheroid from a control network their are
+        If you are trying to generate a spheroid from a control network there are
         other programs that can do this for you. An easier but more naive method
-        is ingesting the <i>OUTPUT_CSV</i> from your network and generating radii
-        information from those points.
+        is ingesting the <i>OUTPUT_CSV</i> from your network, gather local radii
+        information from those points, then generate a sphereiod from those local radii.
       </p>
       </blockquote>
 
--- a/isis/src/control/apps/jigsaw/jigsaw.xml
+++ b/isis/src/control/apps/jigsaw/jigsaw.xml
@@ -64,6 +64,12 @@
      pages.
     </p>
 
+    <h4>Relevant Documentation</h4>
+    <p>
+      For information on what the original <b>jigsaw</b> code was based on checkout
+      <a href="https://www.rand.org/pubs/notes/N3330.html">Rand Notebook</a>
+    </p>
+
     <h4>Known Issues</h4>
       <blockquote>
       <b>Running jigsaw with a control net containing <i>JigsawRejected</i>
@@ -84,6 +90,27 @@
         clear the <i>JigsawRejected</i> flags.</li>
         </ol>
         </blockquote>
+      </p>
+      </blockquote>
+
+      <blockquote>
+      <b>Solving for the target body radii (triaxial or mean) is irrelavent and
+      likely increases error in the solve.</b>
+      <p>
+        When solving for target body radii both the mean and triaxial solve options
+        are redundantly solving for the target body radii. The radii solve is part of
+        the sequence of equations that the <b>jigsaw</b> uses to compute various partial
+        derivatives to populate the solve matrix. The radii solve is being applied
+        to the target body separately and the partials from this separate application
+        are being added to the solve matrix. This is adding additional computation time
+        to <b>jigsaw</b> and creating additional uncertainty/error in the bundle adjust. We
+        advise this radii solve to be avoided.
+      </p>
+      <p>
+        If you are trying to generate a spheroid from a control network their are
+        other programs that can do this for you. An easier but more naive method
+        is ingesting the <i>OUTPUT_CSV</i> from your network and generating radii
+        information from those points.
       </p>
       </blockquote>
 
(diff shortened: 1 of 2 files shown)


isis/src/control/apps/jigsaw/jigsaw.xml
```diff
@@ -87,6 +93,27 @@
...
+        are redundantly solving for the target body radii. The radii solve is part of
+        the sequence of equations that the <b>jigsaw</b> uses to compute various partial
+        derivatives to populate the solve matrix. The radii solve is being applied
+        to the target body separately and the partials from this separate application
+        are being added to the solve matrix. This is adding additional computation time
+        to <b>jigsaw</b> and creating additional uncertainty/error in the bundle adjust. We
+        advise this radii solve to be avoided.
+      </p>
+      <p>
+        If you are trying to generate a spheroid from a control network their are
```
AustinSanders: their -> there

AustinSanders (commented):
Just one quick typo fix

isis/src/control/apps/jigsaw/jigsaw.xml
```diff
@@ -87,6 +93,27 @@
       </p>
       </blockquote>
 
+      <blockquote>
+      <b>Solving for the target body radii (triaxial or mean) is irrelavent and
+      likely increases error in the solve.</b>
+      <p>
+        When solving for target body radii both the mean and triaxial solve options
+        are redundantly solving for the target body radii. The radii solve is part of
```
jessemapel: This is not so much redundant as not possible. There is no dependence of the back projection upon the target radii. It instead depends upon the local point radii which can then be collectively evaluated, as described later, to fit target radii.

jessemapel: Be very careful about terminology here. Target radii == the radii of an ellipsoid fit to the body, while local radii == the local radius of each control point. The two are not the same, but they are closely related. Each point radius is a local approximation of the global target radii.

acpaquette: Reworded and added to this section to mention that the local point radii is being solved for rather than just some magic radii

AustinSanders (approved):

jessemapel (approved):

PR #4223, opened 2020-12-30
Topds4 real parameters

scsides:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have read and agree to abide by the [Code of Conduct](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/Code-Of-Conduct.md)
- [ ] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.
- [ ] I have added any user impacting changes to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [X] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #4226, opened 2020-12-30
Adds PDS4 template callback functions

jessemapel:
This adds 3 callbacks:

* The current time
* The name of the export img file
* The MD5 hash of the cube file

and tests for each of them to topds4. Because the template reading is not in the branch yet, I hardcoded the templates and would uncomment each to check them. Once #4223 is merged I will get this merged with that and the tests should work properly.

### DIFF
--- a/isis/src/base/apps/topds4/topds4.cpp
+++ b/isis/src/base/apps/topds4/topds4.cpp
@@ -1,6 +1,9 @@
 #include <iostream>
+#include <time.h>
 
 #include <inja/inja.hpp>
+
+#include "md5wrapper.h"
 
 #include "topds4.h"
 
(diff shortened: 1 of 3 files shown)


isis/src/base/apps/topds4/topds4.cpp
```diff
@@ -26,17 +29,47 @@ namespace Isis {
...
+    // Call back functions
+    /**
+     * Renders to the current UTC time formatted as YYYY-MM-DDTHH:MM:SS
+     */
+    renderEnv.add_callback("currentTime", 0, [](Arguments& args) {
+      time_t startTime = time(NULL);
+      struct tm *tmbuf = gmtime(&startTime);
+      char timestr[80];
+      strftime(timestr, 80, "%Y-%m-%dT%H:%M:%S", tmbuf);
+      return string(timestr);
```
jessemapel, line 80: Lifted from mrf2pds

isis/tests/FunctionalTestsTopds4.cpp
```diff
@@ -0,0 +1,85 @@
...
+  char yearstr[80];
+  // Just check that the year is correct
+  strftime(yearstr, 80, "%Y", tmbuf);
+  EXPECT_EQ(yearstr, line.substr(0, 4));
+  // Check that the rest is the correct format
+  // YYYY-MM-DDTHH:MM:SS
+  QRegExp timeRegex("\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}");
+  EXPECT_TRUE(QString::fromStdString(line).contains(timeRegex)) << "String [" << line << "] "
+                                                                << "does not match the time format "
+                                                                << "[YYYY-MM-DDTHH:MM:SS].";
```
jessemapel, line 45: Here's an example output for a failure, this one I had the ouput file contain the MD5 hash,

```
1729: /Users/jmapel/ISIS3/isis/tests/FunctionalTestsTopds4.cpp:43: Failure
1729: Value of: QString::fromStdString(line).contains(timeRegex)
1729:   Actual: false
1729: Expected: true
1729: String [7ca9d00ad2b259d3c4bf95b6e5b477fd] does not match the time format [YYYY-MM-DDTHH:MM:SS].
```

isis/src/base/apps/topds4/topds4.cpp
```diff
@@ -58,13 +60,41 @@ namespace Isis {
...
+      QString cubeFilename = icube->fileName().split("/").back();
+      return (cubeFilename.split(".")[0] + ".img").toStdString();
+    });
+
+    /**
+     * Renders to the MD5 hash for the input cube
+     */
+    env.add_callback("md5Hash", 0, [icube](Arguments& args) {
+      md5wrapper md5;
+      return md5.getHashFromFile(icube->fileName()).toStdString();
```
scsides, line 96: Is this for the case where a pds4 label points to an ISIS cube? If so, then should we change the name to reflect that? I'm thinking there may be other md4Hash functions (e.g., The xml may point to a DN only file). If not then when would we use this specific callback?

scsides (approved):

isis/src/base/apps/topds4/topds4.cpp
jessemapel: I just put this in to see if we could call stuff from other places in the callbacks. This comes from lronac2pds where it is the checksum of the output file. Once we are properly exporting the output file then I'll change it to that.

PR #4232, opened 2021-01-01
automate copying of init scrips

### LINKED ISSUE #4231: Installation procedure: Copy conda activation/deactivation scripts to other envs
**Description**  

The instructions describe how one can  manually copy the init scripts to other conda envs.
This can be cumbersome and scary for new terminal users.
It would be good if this could be scripted.

michaelaye:
## Description

* scripted identification of existing conda envs
* UI to ask user which envs should be ISIS-initialized
* copying of newly created act/deact scripts to user-provided conda envs
* add `echo "ISIS version info"` at activation time to reduce user confusion
* improve README

## Related Issue

fixes #4231 

## Motivation and Context

Installation hurdles are one of the biggest stoppers of more user involvement and willingness of installing/updating new versions. 
The less cumbersome the install appears, the less apprehensive users will be for updating.

## How Has This Been Tested?

Ran script locally and had scripts installed into 3 other envs:

```bash
$ python isisVarInit.py -d ~/big_drive/isisdata                                                                                                      (isis) 
-- ISIS Data Directories --
/home/maye/big_drive/isisdata exists, don't need to create.
/home/maye/miniconda3/envs/isis/testData exists, don't need to create.
/home/maye/miniconda3/envs/isis/aleData exists, don't need to create.
-- Conda activation and deactivation scripts --
/home/maye/miniconda3/envs/isis/etc/conda/activate.d exists, don't need to create.
/home/maye/miniconda3/envs/isis/etc/conda/deactivate.d exists, don't need to create.
Searching for conda envs in /home/maye/miniconda3/envs ...
Found the following conda envs:
0 py38
1 jlab3
2 py37
3 lsp
4 isis
Which envs should receive the ISIS path initialization?
(If you don't want to copy the init scripts to other envs, just press RETURN.)
Comma-separated list of env indexes (e.g. 1,2,4):0,1,2
Wrote /home/maye/miniconda3/envs/isis/etc/conda/activate.d/isis-activate.sh
Wrote /home/maye/miniconda3/envs/isis/etc/conda/deactivate.d/isis-deactivate.sh
Wrote /home/maye/miniconda3/envs/py38/etc/conda/activate.d/isis-activate.sh
Wrote /home/maye/miniconda3/envs/py38/etc/conda/deactivate.d/isis-deactivate.sh
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/activate.d/isis-activate.sh
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/deactivate.d/isis-deactivate.sh
Wrote /home/maye/miniconda3/envs/py37/etc/conda/activate.d/isis-activate.sh
Wrote /home/maye/miniconda3/envs/py37/etc/conda/deactivate.d/isis-deactivate.sh
Wrote /home/maye/miniconda3/envs/isis/etc/conda/activate.d/isis-activate.csh
Wrote /home/maye/miniconda3/envs/isis/etc/conda/deactivate.d/isis-deactivate.csh
Wrote /home/maye/miniconda3/envs/py38/etc/conda/activate.d/isis-activate.csh
Wrote /home/maye/miniconda3/envs/py38/etc/conda/deactivate.d/isis-deactivate.csh
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/activate.d/isis-activate.csh
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/deactivate.d/isis-deactivate.csh
Wrote /home/maye/miniconda3/envs/py37/etc/conda/activate.d/isis-activate.csh
Wrote /home/maye/miniconda3/envs/py37/etc/conda/deactivate.d/isis-deactivate.csh
Wrote /home/maye/miniconda3/envs/isis/etc/conda/activate.d/isis-activate.fish
Wrote /home/maye/miniconda3/envs/isis/etc/conda/deactivate.d/isis-deactivate.fish
Wrote /home/maye/miniconda3/envs/py38/etc/conda/activate.d/isis-activate.fish
Wrote /home/maye/miniconda3/envs/py38/etc/conda/deactivate.d/isis-deactivate.fish
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/activate.d/isis-activate.fish
Wrote /home/maye/miniconda3/envs/jlab3/etc/conda/deactivate.d/isis-deactivate.fish
Wrote /home/maye/miniconda3/envs/py37/etc/conda/activate.d/isis-activate.fish
Wrote /home/maye/miniconda3/envs/py37/etc/conda/deactivate.d/isis-deactivate.fish
```

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -25,6 +25,11 @@
       "name": "Austin, Christopher"
     },
     {
+      "affiliation": "Laboratory for Atmospheric and Space Physic (LASP), University of Colorado Boulder",
+      "name": "Aye, Michael Aye",
+      "orcid": "https://orcid.org/0000-0002-4088-1928"
+    },
+    {
       "name": "Backer, Jeanie"
     },
     {
@@ -40,7 +45,7 @@
     {
       "name": "Beyer, Ross",
       "affiliation": "SETI Institute and NASA Ames Research Center",
-      "orcid": "0000-0003-4503-3335"
+      "orcid": "https://orcid.org/0000-0003-4503-3335"
     },
     {
       "name": "Bonn, John"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ update the Unreleased link so that it compares against the latest release tag.
 
 ## [Unreleased]
 
+- added new switch `--init-other-envs` to scripts/isisVarInit.py for copying the needed init-files to other conda envs.
+
 ### Added
 
 - Added warning to ocams2isis about the model being out of date. [#4200](https://github.com/USGS-Astrogeology/ISIS3/issues/4200)
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
         #Activate the environment
         #Depending on your version of Anaconda use one of the following:
 
-        #Anaconda 3.4 and up:
+        #Anaconda 3.4 and up (conda>=4.6):
         conda activate isis
 
-        #Prior to Anaconda 3.4:
+        #Prior to Anaconda 3.4 (conda<4.6):
         source activate isis
 
         #Add the following channels to the environment
@@ -77,15 +77,21 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
 
         conda config --env --add channels usgs-astrogeology
 
+        # If you have followed conda-forges advice to set `channel_priority=strict`,
+        # you need to set it to flexible for the install:
+        conda config --set channel_priority flexible
+
 
 6.  The environment is now ready to download ISIS and its dependencies:
 
-        conda install -c usgs-astrogeology isis
+        # To install e.g. version 4.3
+        conda install isis=4.3
 
     If you would like to work with our latest ISIS version 3, rather than updating
     to ISIS 4, instead run:
 
-	    conda install -c usgs-astrogeology isis=3.10.0
+	    # This will find the latest 3.x.y release
+        conda install isis=3
 
 
 7.  Finally, setup the environment variables:
@@ -110,7 +116,14 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
 
         python $CONDA_PREFIX/scripts/isisVarInit.py --data-dir=[path to data directory]  --test-dir=[path to test data directory]
 
-    More information about the ISISDATA environment variable and the ISIS Data Area can be found [here]("#The-ISIS-Data-Area"). Now everytime the isis environment is activated, $ISISROOT, $ISISDATA, and $ISISTESTDATA will be set to the values passed to isisVarInit.py. This does not happen retroactively, so re-activate the isis envionment with one of the following commands:
+
+    For the newer version we also support automatic copying of the init files to other conda environment. If you require this, call the script with the optional switch `--init-other-envs`:
+
+        python @CONDA_PREFIX/scripts/isisVarInit.py --init-other-envs
+
+    If you rather want to copy the files manually, see the instructions in the next section.
+
+    More information about the ISISDATA environment variable and the ISIS Data Area can be found [here]("#The-ISIS-Data-Area"). Now everytime the ISIS environment is activated, $ISISROOT, $ISISDATA, and $ISISTESTDATA will be set to the values passed to isisVarInit.py. This does not happen retroactively, so re-activate the isis envionment with one of the following commands:
 
         for Anaconda 3.4 and up - conda activate isis
         prior to Anaconda 3.4 - source activate isis
--- a/isis/scripts/isisVarInit.py
+++ b/isis/scripts/isisVarInit.py
@@ -10,7 +10,9 @@ not specified, they are placed in $ISISROOT.
 
 import argparse
 import os
+import shutil
 from pathlib import Path
+from typing import List, Tuple
 
 # This is free and unencumbered software released into the public domain.
 #
@@ -20,12 +22,50 @@ from pathlib import Path
 #
 # SPDX-License-Identifier: CC0-1.0
 
+# This is the full path to this environment:
+CONDA_PREFIX = os.environ["CONDA_PREFIX"]
+ENVS_DIR = Path(CONDA_PREFIX).parent
+THIS_ENV = Path(CONDA_PREFIX).name
+
+
+def get_conda_envs() -> List[Path]:
+    """Returns a list of paths to the found conda envs.
+
+    This method is used because
+    (1) there is no API for `conda env`,
+    (2) `subprocess.check_output` and `sh.conda` are not producing the same output.
+    """
+    print(f"Searching for other conda envs in {ENVS_DIR} ...")
+    paths = [d for d in ENVS_DIR.iterdir() if d.is_dir()]
+    return [p.name for p in paths if p.name != THIS_ENV]
+
+
+def get_envs_to_init() -> List[Path]:
+    """Find conda envs and ask user which ones should initialize ISIS."""
+    env_names = get_conda_envs()
+    print("Found the following other conda envs:")
+    for i, env in enumerate(env_names):
+        print(i, env)
+    print("Which envs should receive the ISIS path initialization?")
+    print("(Press RETURN for skipping.)")
+    answer = ""
+    answer = input("Comma-separated list of env indexes (e.g. 1,2,4):")
+    if not answer:
+        return []
+    try:
+        envs_to_init = [env_names[int(i)] for i in answer.split(",")]
+    except ValueError:
+        raise ValueError(
+            "Could not parse your answer.\nDid you use a comma-separated list?"
+        )
+    return envs_to_init
+
 
 def mkdir(p: Path) -> str:
-    """Returns a string with a message about the creation or existance
-    of the path, *p*."""
+    """Create directory `p` if not exists, and notify user when created."""
+
     if p.exists():
-        return f"{p} exists, don't need to create."
+        return p
     else:
         p.mkdir(parents=False)
         return f"Created {p}"
@@ -42,6 +82,8 @@ def activate_text(shell: dict, env_vars: dict) -> str:
     if shell["activate_extra"] != "":
         lines.append(shell["activate_extra"])
 
+    lines.append(shell['path_update'])
+    lines.append("echo 'ISIS version info:'")
     lines.append("cat $ISISROOT/version")
 
     return "\n".join(lines)
@@ -58,20 +100,33 @@ def deactivate_text(shell: dict, env_vars: dict) -> str:
     return "\n".join(lines)
 
 
+def get_paths_make_dirs(env: str) -> Tuple[Path]:
+    "Create act/deact paths, create the dirs, and return paths."
+
+    # Create the conda activation and deactivation directories:
+    activate_dir = ENVS_DIR / f"{env}/etc/conda/activate.d"
+    deactivate_dir = ENVS_DIR / f"{env}/etc/conda/deactivate.d"
+
+    mkdir(activate_dir)
+    mkdir(deactivate_dir)
+    return activate_dir, deactivate_dir
+
+
 # Set up and then parse the command line:
 parser = argparse.ArgumentParser(description=__doc__)
+
 
 parser.add_argument(
     "-d",
     "--data-dir",
-    default=os.environ["CONDA_PREFIX"] + "/data",
+    default=CONDA_PREFIX + "/data",
     type=Path,
     help="ISIS Data Directory, default: %(default)s",
 )
 parser.add_argument(
     "-t",
     "--test-dir",
-    default=os.environ["CONDA_PREFIX"] + "/testData",
+    default=CONDA_PREFIX + "/testData",
     type=Path,
     help="ISIS Test Data Directory, default: %(default)s",
 )
@@ -79,10 +134,17 @@ parser.add_argument(
 parser.add_argument(
     "-a",
     "--ale-dir",
-    default=os.environ["CONDA_PREFIX"] + "/aleData",
+    default=CONDA_PREFIX + "/aleData",
     type=Path,
     help="ISIS Ale Data Directory, default: %(default)s",
 )
+
+parser.add_argument(
+    "--init-other-envs",
+    action="store_true",
+    help="Launch dialog to find other conda environments to be initialized.",
+)
+
 args = parser.parse_args()
 
 print("-- ISIS Data Directories --")
@@ -90,57 +152,72 @@ print("-- ISIS Data Directories --")
 print(mkdir(args.data_dir))
 print(mkdir(args.test_dir))
 print(mkdir(args.ale_dir))
+print()
 
-print("-- Conda activation and deactivation scripts --")
-# Create the conda activation and deactivation directories:
-activate_dir = Path(os.environ["CONDA_PREFIX"]) / "etc/conda/activate.d"
-deactivate_dir = Path(os.environ["CONDA_PREFIX"]) / "etc/conda/deactivate.d"
+# always init this env:
+envs_to_init = [THIS_ENV]
 
-print(mkdir(activate_dir))
-print(mkdir(deactivate_dir))
+# if wanted, ask user which other conda envs should get ISIS-initialized
+if args.init_other_envs:
+    envs_to_init += get_envs_to_init()
 
 # Set the environment variables to manage
 env_vars = dict(
-    ISISROOT=os.environ["CONDA_PREFIX"],
+    ISISROOT=CONDA_PREFIX,
     ISISDATA=args.data_dir,
     ISISTESTDATA=args.test_dir,
-    ALESPICEROOT=args.ale_dir
+    ALESPICEROOT=args.ale_dir,
 )
 
 # These dicts define the unique aspects of the shell languages
 # for setting and unsetting the environment variables.
 
 sh = dict(  # this covers bash and zsh
+    name="bash/zsh",
     extension=".sh",
     shebang="#!/usr/bin/env sh",
     activate="export {}={}",
     activate_extra="",
-    deactivate="unset {}"
+    deactivate="unset {}",
+    path_update="export PATH=$PATH:$ISISROOT/bin"
 )
 
 csh = dict(
+    name="csh",
     extension=".csh",
     shebang="#!/usr/bin/env csh",
     activate="setenv {} {}",
     activate_extra="source $CONDA_PREFIX/scripts/tabCompletion.csh",
-    deactivate="unsetenv {}"
+    deactivate="unsetenv {}",
+    path_update="setenv PATH $PATH:$ISISROOT/bin"
 )
 
 fish = dict(
+    name="fish",
     extension=".fish",
     shebang="#!/usr/bin/env fish",
     activate="set -gx {} {}",
     activate_extra="",
-    deactivate="set -e {}"
+    deactivate="set -e {}",
+    path_update="set -agx PATH $ISISROOT/bin"
 )
 
 # Loop over the shell types and create the correct scripts associated with
 # each:
-for shell in (sh, csh, fish):
-    a_path = activate_dir / ("isis-activate" + shell["extension"])
-    a_path.write_text(activate_text(shell, env_vars))
-    print(f"Wrote {a_path}")
+print("\nInit files will be written for the following conda environments:")
+print(envs_to_init)
 
-    d_path = deactivate_dir / ("isis-deactivate" + shell["extension"])
-    d_path.write_text(deactivate_text(shell, env_vars))
-    print(f"Wrote {d_path}")
+for env in envs_to_init:
+    print(f"\n-- Initializing {env} conda env --\n")
+    a_dir, d_dir = get_paths_make_dirs(env)
+    print(f"Activation in {a_dir}")
+    print(f"Deactivaton in {d_dir}")
+    for shell in (sh, csh, fish):
+        print(f"{shell['name']} shell:")
+        a_path = a_dir / ("isis-activate" + shell["extension"])
+        a_path.write_text(activate_text(shell, env_vars))
+        print(f"Wrote {a_path.name}")
+
+        d_path = d_dir / ("isis-deactivate" + shell["extension"])
+        d_path.write_text(deactivate_text(shell, env_vars))
+        print(f"Wrote {d_path.name}")


rbeyer:
Don't add author names to individual files, that's not a practice that we want to encourage in this Public Domain software repo.  Instead, at a minimum, please add your name to the [.zenodo.json file](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json), and add it as part of the PR, as the [CONTRIBUTING.md document indicates](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md#contributions).

On the one hand, the intent for this script is to stay simple, and so there's an argument for putting this functionality in *another* script.  However, the functionality is so similar, that including it here might be good.  Not sure.  At any rate, here are my thoughts about the functionality:
 
- It is confusing that the "isis" environment is presented as an option, but will be installed by default, even if the user didn't select it.

- I feel like this new functionality you are presenting should be an optional capability, and not the default thing that happens every time this runs.  I suggest adding a switch that enables this functionality.  If the user doesn't ask for it, then this should run as it did before (only considering *this* environment and not others and not presenting any input prompts to the user).  Otherwise, you'd have to add a whole bunch more functionality and switches to allow this to be run without user interaction in some sort of CI environment (which is already a pain/challenge for ISIS because you have to potentially download some ISISDATA and set these env vars: [the horror](https://github.com/rbeyer/kalasiris/blob/main/.github/workflows/python-test.yml)).

Here are some user stories / thoughts:
- If someone knows nothing about other conda environments, and only has the one ISIS conda env, then the script shouldn't complicate that interaction (right now, it would present a single option, and would carry out the install one time if no item was selected, and two times if it was--which is also kind of a bug).
- If someone already has the current "isis" environment installed, they should be able to come back to the program, and its "default" behavior would overwrite *this* repo's scripts, but when this optional argument is set, then it would   only change *other* repos (potentially resulting in different values of -d and -t for the other repos, perhaps defaulting to those of *this* repo, without changing the activate scripts for *this* repo).
- If someone has multiple ISIS conda environments installed, this also has the potential ability to mangle the ISIS activate scripts of other ISIS conda environments.  One could argue that this is an "active" choice by the user, the same as `rm *` is, so I'm not sure this needs to be particularly considered.

michaelaye:
> Don't add author names to individual files, that's not a practice that we want to encourage in this Public Domain software repo. Instead, at a minimum, please add your name to the [.zenodo.json file](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json), and add it as part of the PR, as the [CONTRIBUTING.md document indicates](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md#contributions).

Okay. I wanted to easily identify who wrote something, but I guess we have `git annotate` for that.

> 
> On the one hand, the intent for this script is to stay simple, and so there's an argument for putting this functionality in _another_ script. However, the functionality is so similar, that including it here might be good. Not sure. At any rate, here are my thoughts about the functionality:
> 
> * It is confusing that the "isis" environment is presented as an option, but will be installed by default, even if the user didn't select it.

Yep, that's a bug.

> * I feel like this new functionality you are presenting should be an optional capability, and not the default thing that happens every time this runs.  

I would agree on grounds of continuity, but I'm not sure you have a good case on user base/story grounds, see below.

> I suggest adding a switch that enables this functionality.  If the user doesn't ask for it, then this should run as it did before (only considering _this_ environment and not others and not presenting any input prompts to the user).  Otherwise, you'd have to add a whole bunch more functionality and switches to allow this to be run without user interaction in some sort of CI environment (which is already a pain/challenge for ISIS because you have to potentially download some ISISDATA and set these env vars: [the horror](https://github.com/rbeyer/kalasiris/blob/main/.github/workflows/python-test.yml)).

There's always pyautogui for testing UIs/CLIs (and other tools), but I agree that if we really insist on testing this script (can't see any so far), that a no-UI workflow is obviously less effort on our side. Not sure that it'd serve the majority of user stories though.

> 
> Here are some user stories / thoughts:
> 
> * If someone knows nothing about other conda environments, and only has the one ISIS conda env, then the script shouldn't complicate that interaction (right now, it would present a single option, and would carry out the install one time if no item was selected, and two times if it was--which is also kind of a bug).

If this were the most frequent case, I'd agree, however, due to Anaconda being the [world’s most popular Python distribution platform with over 20 million users worldwide](https://www.anaconda.com/products/individual) I am claiming that this is the more frequent user story:

Someone has installed Anaconda, because it's the most popular Python distribution. They are not even aware that they have a conda environment, they use the GUI for managing it and launching Jupyter and thusly are not familiar with working in the terminal. (This is the most likely user scenario during my Python intro classes). For working with ISIS they are now forced to install another conda environment and they are very overwhelmed by all the terminal interaction required.

> * If someone already has the current "isis" environment installed, they should be able to come back to the program, 

with "the program" you mean this script I am working on?

> and its "default" behavior would overwrite _this_ repo's scripts, but when this optional argument is set, then it would   only change _other_ repos (potentially resulting in different values of -d and -t for the other repos, perhaps defaulting to those of _this_ repo, without changing the activate scripts for _this_ repo).

Sorry, I'm lost, to many `this` I can't decipher what they refer to.

> * If someone has multiple ISIS conda environments installed, this also has the potential ability to mangle the ISIS activate scripts of other ISIS conda environments.  One could argue that this is an "active" choice by the user, the same as `rm *` is, so I'm not sure this needs to be particularly considered.

We could add a safety question if the file exists?

rbeyer:
> > * I feel like this new functionality you are presenting should be an optional capability, and not the default thing that happens every time this runs.
> 
> I would agree on grounds of continuity, 

Glad you agree.


> > I suggest adding a switch that enables this functionality.  If the user doesn't ask for it, then this should run as it did before (only considering _this_ environment and not others and not presenting any input prompts to the user).  Otherwise, you'd have to add a whole bunch more functionality and switches to allow this to be run without user interaction in some sort of CI environment (which is already a pain/challenge for ISIS because you have to potentially download some ISISDATA and set these env vars: [the horror](https://github.com/rbeyer/kalasiris/blob/main/.github/workflows/python-test.yml)).
> 
> There's always pyautogui for testing UIs/CLIs (and other tools), but I agree that if we really insist on testing this script (can't see any so far), that a no-UI workflow is obviously less effort on our side. Not sure that it'd serve the majority of user stories though.

Perhaps I wasn't clear, I meant CI (continuous integration), not CLI (command line interface).  I'm talking about things that have ISIS as a dependency that must install ISIS (and thus must also run this program) as part of "ISIS Installation."  In order to be able to do that in a continuous integration environment, this program should not present a prompt (unless it also provides a `-y` option or something similar to `conda install` and many others).  At any rate, making this "create or overwrite activate/deactivate scripts in other environments" functionality an option is the way to go, and then this isn't a problem.

> > Here are some user stories / thoughts:
> > 
> > * If someone knows nothing about other conda environments, and only has the one ISIS conda env, then the script shouldn't complicate that interaction (right now, it would present a single option, and would carry out the install one time if no item was selected, and two times if it was--which is also kind of a bug).
> 
> If this were the most frequent case, I'd agree, however, due to Anaconda being the [world’s most popular Python distribution platform with over 20 million users worldwide](https://www.anaconda.com/products/individual) I am claiming that this is the more frequent user story:
> 
> Someone has installed Anaconda, because it's the most popular Python distribution. They are not even aware that they have a conda environment, they use the GUI for managing it and launching Jupyter and thusly are not familiar with working in the terminal. (This is the most likely user scenario during my Python intro classes). For working with ISIS they are now forced to install another conda environment and they are very overwhelmed by all the terminal interaction required.

It may very well be the most frequent that you have encountered.  The most frequent that I have encountered is "what is this conda thing that I now have to install?"  I'm not entirely sure what we're in conflict about here.  it seems to me that the default functionality for this program should be to install the scripts in the conda environment in which it lives, and it should have the option to to ease installation in other environments.

> > * If someone already has the current "isis" environment installed, they should be able to come back to the program,
> 
> with "the program" you mean this script I am working on?

Yes.

> 
> > and its "default" behavior would overwrite _this_ repo's scripts, but when this optional argument is set, then it would   only change _other_ repos (potentially resulting in different values of -d and -t for the other repos, perhaps defaulting to those of _this_ repo, without changing the activate scripts for _this_ repo).
> 
> Sorry, I'm lost, to many `this` I can't decipher what they refer to.

They have the same reference each time: the repo in which the script was installed (i.e. the one with *isis* installed in it, typically named *isis*).

I wouldn't worry about it.  I look forward to the next iteration, I think you're headed in a good direction, and we're splitting hairs until you make some more changes.

> > * If someone has multiple ISIS conda environments installed, this also has the potential ability to mangle the ISIS activate scripts of other ISIS conda environments.  One could argue that this is an "active" choice by the user, the same as `rm *` is, so I'm not sure this needs to be particularly considered.
> 
> We could add a safety question if the file exists?

You could, but it gets complicated with older versions of ISIS and `isis3VarInit.py` which installed activation/deactivation scripts with a different name, and then you could have two activate scripts which both get run, and it could be a whole nightmare, but the chance is probably low, and its probably fine.

michaelaye:
I just noticed that the default docs are apparently not advising on setting up the PATH and the script wouldn't do that either, can that be true?

michaelaye:
Note that I don't consider hacking the PATH variable to remove a path entry as necessary as closing the current shell will simply achieve the same. If you want, I can add an "echo" warning to the deactivation to remind the user about that?

README.md
```diff
@@ -53,10 +53,10 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
         #Activate the environment
         #Depending on your version of Anaconda use one of the following:
 
-        #Anaconda 3.4 and up:
+        #Anaconda 3.4 and up (conda>=4.6):
         conda activate isis
 
-        #Prior to Anaconda 3.4:
+        #Prior to Anaconda 3.4 (conda<4.6):
```
rbeyer, line 59: I'm confused here, are you adding these parenthetical statements because the `source activate` command is still allowable for conda until conda 4.6?  If so, I wouldn't include this.  It doesn't particularly clarify the instructions for novices, and conda experts like yourself already know these historical caveats.

README.md
```diff
@@ -77,15 +77,21 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
...
-        conda install -c usgs-astrogeology isis
+        # To install e.g. version 4.3
+        conda install isis=4.3
 
     If you would like to work with our latest ISIS version 3, rather than updating
     to ISIS 4, instead run:
 
-	    conda install -c usgs-astrogeology isis=3.10.0
+	    # This will find the latest 3.x.y release
+        conda install isis=3
```
rbeyer, line 94: I don't think this will work.  Did you test it?  I mean, it would if the conda isis package had been developed as a conda package for many years, and versions and such were coherent, but as far as I know you must do `conda install isis3` for the latest ISIS 3.* or `conda install isis3=3.6.1` to get a specific version.

michaelaye:
I was thinking of miniconda users that are completely unaware of Anaconda versions.

--
Michael Aye
LASP / CU Boulder

README.md
michaelaye: I was thinking of miniconda users that don't know Anaconda versions at all.

rbeyer: Oh ... Anaconda and miniconda are not version-locked? Did not know. In that case, I'd explicitly say "miniconda" instead of "conda" in your parenthetical. "conda" is often used as the generic thing, and could mean either (and that's how I was parsing it!)

isis/scripts/isisVarInit.py
```diff
@@ -20,12 +22,50 @@
...
+
 
 def mkdir(p: Path) -> str:
-    """Returns a string with a message about the creation or existance
-    of the path, *p*."""
+    """Create directory `p` if not exists, and notify user when created."""
+
     if p.exists():
-        return f"{p} exists, don't need to create."
+        return p
```
rbeyer, line 68: This function now sometimes returns a string, and sometimes returns a Path, which is weird (and violates the return signature at the very least).  It should always return a string (like it did) or always return a Path.  Pick one.

isis/scripts/isisVarInit.py
```diff
@@ -42,6 +82,8 @@ def activate_text(shell: dict, env_vars: dict) -> str:
     if shell["activate_extra"] != "":
         lines.append(shell["activate_extra"])
 
+    lines.append(shell['path_update'])
```
rbeyer, line 85: This is where things get dicey, and I'm worried about where this is going.  The default version of this program, and the activate scripts did not need to set the path, because the ISIS programs were simply in $CONDA_PREFIX/bin, and all of the path setting nonsense was taken care of by conda.  Putting this functionality into the activate script fundamentally circumvents all of the path wrangling that conda does, and I'm concerned that there is a high liklihood of downstream complications.  This should not be done in a script like this that we expect non-experts to use.

Explicitly doing this will place the $ISISROOT/bin into the PATH twice (once when conda sets up the path, and once when you run the activate script) when run in the current environment.  The fact that any "other" environment that you write these scripts into will get a path add, but not a path removal on conda deactivate seems like a real bad idea.  I appreciate that you can just "terminate the shell," but that requires that you know what you're doing.  I suspect that for less savvy users, this could result in a kind of memory leak situation, in which someone blithely uses the same shell and activates and deactivates one of these environments over and over again, with their PATH variable growing and growing and growing.

.zenodo.json
```diff
@@ -40,7 +45,7 @@
     {
       "name": "Beyer, Ross",
       "affiliation": "SETI Institute and NASA Ames Research Center",
-      "orcid": "0000-0003-4503-3335"
+      "orcid": "https://orcid.org/0000-0003-4503-3335"
```
rbeyer, line 48: The .zenodo.json file "standard" is a hot mess as far as I can tell.  I thought that "orcid" here was just supposed to be the id, where are you seeing that it should be a URL?

README.md
```diff
@@ -110,7 +116,14 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
 
         python $CONDA_PREFIX/scripts/isisVarInit.py --data-dir=[path to data directory]  --test-dir=[path to test data directory]
 
-    More information about the ISISDATA environment variable and the ISIS Data Area can be found [here]("#The-ISIS-Data-Area"). Now everytime the isis environment is activated, $ISISROOT, $ISISDATA, and $ISISTESTDATA will be set to the values passed to isisVarInit.py. This does not happen retroactively, so re-activate the isis envionment with one of the following commands:
+
+    For the newer version we also support automatic copying of the init files to other conda environment. If you require this, call the script with the optional switch `--init-other-envs`:
```
jlaura, line 120: s/environment/environments

README.md
```diff
@@ -110,7 +116,14 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
...
+
+        python @CONDA_PREFIX/scripts/isisVarInit.py --init-other-envs
+
```
jlaura, line 123: Does this init all other envs?

isis/scripts/isisVarInit.py
```diff
@@ -20,12 +22,50 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
+# This is the full path to this environment:
+CONDA_PREFIX = os.environ["CONDA_PREFIX"]
```
jlaura, line 26: If no conda env is activated this will throw right? Should we have a os.environ.get('CONDA_PREFIX', None) and then throw an error?

isis/scripts/isisVarInit.py
```diff
@@ -20,12 +22,50 @@
...
+ENVS_DIR = Path(CONDA_PREFIX).parent
+THIS_ENV = Path(CONDA_PREFIX).name
```
jlaura, line 28: These path lib changes are quite nice aren't they!

isis/scripts/isisVarInit.py
```diff
@@ -20,12 +22,50 @@
...
+    answer = input("Comma-separated list of env indexes (e.g. 1,2,4):")
+    if not answer:
+        return []
+    try:
+        envs_to_init = [env_names[int(i)] for i in answer.split(",")]
+    except ValueError:
+        raise ValueError(
+            "Could not parse your answer.\nDid you use a comma-separated list?"
+        )
+    return envs_to_init
```
jlaura, line 61: If I were to run this, I would have to y/n through 60+ environments. Perhaps it is better to take an optinal list of envs as arguments to the script?

I realize that the counter argument to this is that maybe I do want ISIS in all 60. So impossible to make everyone happy. Maybe the solution is to use either a list or an all flag.

isis/scripts/isisVarInit.py
```diff
@@ -58,89 +100,124 @@ def deactivate_text(shell: dict, env_vars: dict) -> str:
...
+)
+
 args = parser.parse_args()
 
 print("-- ISIS Data Directories --")
 # Create the data directories:
 print(mkdir(args.data_dir))
 print(mkdir(args.test_dir))
 print(mkdir(args.ale_dir))
+print()
```
jlaura, line 155: ? so the output is prettier?

jlaura (changes requested):

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ update the Unreleased link so that it compares against the latest release tag.
 
 ## [Unreleased]
 
+- added new switch `--init-other-envs` to scripts/isisVarInit.py for copying the needed init-files to other conda envs.
```
jessemapel, line 33: This needs to go under the `### Added` header as it's a new addition.

README.md
jessemapel: Can we just axe the source activate part? It's quite a bit out of date at this point.

README.md
```diff
@@ -77,15 +77,21 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
 
         conda config --env --add channels usgs-astrogeology
 
+        # If you have followed conda-forges advice to set `channel_priority=strict`,
+        # you need to set it to flexible for the install:
+        conda config --set channel_priority flexible
```
jessemapel, line 82: It's probably better to just set this for the environment to avoid impacting user's other environments.

README.md
jessemapel: The old isis3 package is defunct. Users should use the new isis package regardless, isis=3 should install 3.10.2

README.md
```diff
@@ -77,15 +77,21 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
...
+
 
 6.  The environment is now ready to download ISIS and its dependencies:
 
-        conda install -c usgs-astrogeology isis
+        # To install e.g. version 4.3
```
jessemapel, line 87: I'd move this up into the text above.

The environment is now ready to download ISIS and its dependencies. To install the latest release, version 4.3, run:

    conda install isis=4.3

README.md
```diff
@@ -77,15 +77,21 @@ This installation guide is for ISIS users interested in installing ISIS (3.6.0)+
...
+        conda install isis=4.3
 
     If you would like to work with our latest ISIS version 3, rather than updating
     to ISIS 4, instead run:
 
-	    conda install -c usgs-astrogeology isis=3.10.0
+	    # This will find the latest 3.x.y release
```
jessemapel, line 93: This is redundant.

jessemapel (commented):
I want to echo Ross' concerns about editing the path. We really should not be touching the PATH, particularly in other environments.

.zenodo.json
michaelaye: An ORCID is the full thing only:

https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier

> The ORCID iD is an https URI with a 16-digit number that is compatible with the ISO Standard (ISO 27729), also known as the International Standard Name Identifier (ISNI), e.g. https://orcid.org/0000-0001-2345-6789

isis/scripts/isisVarInit.py
michaelaye: Yes. It's otherwise information density overload. IMHO.

README.md
michaelaye: Don't understand. My version has the advantage of staying correct even when you'd release 3.10.1 and more?

michaelaye:
In general I am perplexed about the opposition to manipulate the PATH variable? The current instructions advise to do so? I obviously could improve on it, to avoid double-setting etc., but in general that is what needs to happen to set up ISIS completely?
However, I disagree that a written guidance to close the current shell is more difficult to apprehend than the currently given instructions or indeed, e.g. correctly using ISIS?

.zenodo.json
rbeyer: I appreciate that is what the orcid.org people think, but as far as I can tell, Zenodo just wants the number without the URL prefix in their .zenodo.json files (e.g. https://developers.zenodo.org/, https://github.com/nipy/nipype/blob/master/.zenodo.json, https://github.com/perakakis/CoVidAffect/blob/master/.zenodo.json).

README.md
jessemapel: We're going to release any more 3.x versions. 3.10.2 was the final ISIS 3 release.

jessemapel:
My issue with the path modification is that this change will modify the path for all users and not just people trying to get access to ISIS across environments. By adding the path modification to the activate scripts running `conda activate my_isis_env && conda deactivate` will result in a modified path at the end that is not cleaned up, even when you are not trying to stack environments.

README.md
jessemapel: Sorry, my comment wasn't about the new command, it is 👍 .

I do not like having text that says something and then immediately after it having an in-line comment that says the same thing. So, I think you should remove the comment line.

rbeyer:
> In general I am perplexed about the opposition to manipulate the PATH variable? The current instructions advise to do so? 

Let's define "current instructions."  

The default installation instructions for installing ISIS do not require manual manipulation of PATH, nor does isisVarInit.py manipulate it.  The act of "conda install isis" puts the ISIS binaries into the environment's bin/ and that path is automatically managed by conda on "activate" and "deactivate."

The alternate instructions for [Practical Usage with other conda packages](https://github.com/USGS-Astrogeology/ISIS3#practical-usage-with-other-conda-packages) defines two modes: (1) Easy Mode only uses conda's `--stack` argument and also doesn't require manipulation of the PATH variable.  (2) The Hard Mode describes what you are trying to script, it does include path setting in its activation script, and path removal in its deactivation script (but it is a bit fragile).

The difference between Hard Mode and what you are suggesting be incorporated into isisVarInit, is that the Hard Mode instructions require someone do this manually themselves.  This acts as a kind of barrier.  The "average user" who might not be very savvy can't do this "accidentally" and I think that is the source of the opposition: without a large number of safeguards, incorporating this into isisVarInit might have unintended consequences.

> I obviously could improve on it, to avoid double-setting etc., but in general that is what needs to happen to set up ISIS completely?

To get it to run from another environment?  Yes, that's right.  Again, that could be accomplished via `--stack`, it could be accomplished by something in a user's `~/.*rc` file completely outside of the conda context, it could also be accomplished via conda activate and deactivate scripts.  Lots of options.

> However, I disagree that a written guidance to close the current shell is more difficult to apprehend than the currently given instructions or indeed, e.g. correctly using ISIS?

It isn't to someone who knows what they're doing.  I have seen a lot of weird set-ups out there (even with conda) which are based on people just fundamentally not knowing how shells or environment variables work, and the point here is that whatever you "do" in a conda activate script, you should completely "undo" in a conda deactivate script.  If you can't do that, then another solution should be sought.

README.md
michaelaye: confirmed. that's what happpens (with `channel_priority=flexible`)

isis/scripts/isisVarInit.py
michaelaye: it's not per env, you provide a comma-separated list of the printed indexes in the UI dialog.
However, I think taking the actual env names as comma-separated lists would be better and less error prone than using indexes. More work for the user, but it would allow you to paste in a list of envs, yes.

michaelaye:
It should be noted that stacking doesn't work for jupyter notebooks, which is the most used [Python "IDE"](https://www.kdnuggets.com/2020/10/most-popular-python-ides-editors.html), so I'm not sure it's okay to fall back on that as a solution. While ISIS is foremost a set of terminal application, this precludes the use of pysis/kalasiris within Jupyter, something that should be considered, IMHO.

rbeyer:
Sure, that's a fine example where `--stack` won't work, and you'd have to use some other method.

I'll note that as far as `kalasiris` goes, the PATH environment variable doesn't need to be set, you just need ISISROOT and ISISDATA set.  You could even do it in python before you import kalasiris, although this isn't great practice:
```
import os

my_isisroot = os.path.join(os.eviron['HOME'], 'anaconda3' ,'envs', 'isis')
os.environ['ISISROOT'] = my_isisroot
os.environ['ISISDATA'] = os.path.join(my_isisroot, 'data')

import kalasiris
```

PR #4239, opened 2021-01-05
Added Extra data source parameters to topds4

jessemapel:
Added extra pvl, xml, and json data source parameters. They can be a single file or multiple files. In the case of multiple files, duplicate keys/elements in later files overwrite previous files. If this happens, a user warning is logged like this:

```
Group = Warning
  Duplicate = "Duplicate key [TestValue] in extra Pvl file
               [/var/folders/s2/r22kzkt91ms_t4j5b90g5k1w002dkh/T/qt_temp-bLxZd-
               t/extra2.pvl]. Previous value [{"Value":"b"}] will be
               overwritten."
End_Group
End
```

I also re-arranged the included in topds4.cpp and added the xml to json header because I needed it for the extra xml data.

### DIFF
--- a/isis/src/base/apps/topds4/topds4.cpp
+++ b/isis/src/base/apps/topds4/topds4.cpp
@@ -69,7 +69,7 @@ namespace Isis {
               PvlGroup duplicateWarnings("Warning");
               QString message = "Duplicate key [" + QString::fromStdString(element.key())
                               + "] in extra Pvl file [" + pvlFile + "]. "
-                              + "Previous value [" + QString::fromStdString(element.value().dump())
+                              + "Previous value [" + QString::fromStdString(dataSource["ExtraPvl"][element.key()].dump())
                               + "] will be overwritten.";
               duplicateWarnings += PvlKeyword("Duplicate", message);
               log->addGroup(duplicateWarnings);
@@ -92,7 +92,7 @@ namespace Isis {
               PvlGroup duplicateWarnings("Warning");
               QString message = "Duplicate element [" + QString::fromStdString(element.key())
                               + "] in extra xml file [" + xmlFile + "]. "
-                              + "Previous value [" + QString::fromStdString(element.value().dump())
+                              + "Previous value [" + QString::fromStdString(dataSource["ExtraXml"][element.key()].dump())
                               + "] will be overwritten.";
               duplicateWarnings += PvlKeyword("Duplicate", message);
               log->addGroup(duplicateWarnings);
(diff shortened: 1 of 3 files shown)


isis/src/base/apps/topds4/topds4.cpp
```diff
@@ -54,6 +55,78 @@ namespace Isis {
...
+                              + "Previous value [" + QString::fromStdString(element.value().dump())
+                              + "] will be overwritten.";
+              duplicateWarnings += PvlKeyword("Duplicate", message);
+              log->addGroup(duplicateWarnings);
+            }
+          }
+        }
+        dataSource["ExtraJson"].update(extraJson);
+      }
+    }
```
jessemapel, line 128: There is a lot of duplicate code between these three sections, but things are subtly different enough that I couldn't just pull it out into a function.

isis/src/base/apps/topds4/topds4.xml
```diff
@@ -68,7 +68,9 @@
           parameter by removing the last extension and adding .cub
         </description>
       </parameter>
+    </group>
 
+    <group name="Data">
```
jessemapel, line 73: I moved the DATA parameter into this group because it made sense to me, but I'm not 100% on it.

isis/tests/FunctionalTestsTopds4.cpp
```diff
@@ -81,6 +83,232 @@ TEST_F(SmallCube, FunctionalTestTopds4NoOriginalLabel) {
...
+
+  // The duplicate key should generate a warning
+  EXPECT_TRUE(log.hasGroup("Warning"));
+}
+
+TEST_F(SmallCube, FunctionalTestTopds4ExtraXml) {
+  QString xmlFile = tempDir.path()+"/extra.xml";
+  std::ofstream xmlStream;
+  xmlStream.open(xmlFile.toStdString());
+  xmlStream << "<TestValue>a</TestValue>";
```
jessemapel, line 242: Writing raw XML here because it's just easier than using the QT stuff.

isis/src/base/apps/topds4/topds4.xml
scsides: totally reasonable

scsides (approved):
The new tests are painful to look at, but great

jessemapel:
So I botched the warning and put the new value in the warning instead of the old value, here's the updated warning from the top comment:

```
Group = Warning
  Duplicate = "Duplicate key [TestValue] in extra Pvl file
               [/var/folders/s2/r22kzkt91ms_t4j5b90g5k1w002dkh/T/qt_temp-KxImP-
               j/extra2.pvl]. Previous value [{"Value":"a"}] will be
               overwritten."
End_Group
End
```

krlberry (approved):
:+1: Looks good!

scsides (approved):

PR #4241, opened 2021-01-06
Add writing a cube out to topds4

krlberry:
## Description
Converts `cubeatt` to a callable and then uses it in `topds4` to generate an output cube. 

## Related Issue
PDS4 work.

## Motivation and Context
So that `topds4` can output a cube in addition to a label.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/cubeatt/cubeatt.cpp
+++ b/isis/src/base/apps/cubeatt/cubeatt.cpp
@@ -19,6 +19,7 @@ namespace Isis {
   }
 
 
+  // Doesn't allow specification of input attributes
   void cubeatt(Cube *icube, UserInterface &ui) {
     bool propTables = ui.GetBoolean("PROPTABLES");
     QString outputFileName = ui.GetFileName("TO");
@@ -26,7 +27,7 @@ namespace Isis {
     cubeatt(icube, outputFileName, outputAttributes, propTables);
   }
 
-
+  // Doesn't allow specification of input attributes
   void cubeatt(Cube *icube, QString outputCubePath, CubeAttributeOutput outAttributes, bool propTables) {
     // We will be processing by line
     ProcessByLine p;
@@ -42,7 +43,6 @@ namespace Isis {
     p.StartProcess(cubeattProcess);
     p.EndProcess();
   }
-
 
   // Line processing routine
   void cubeattProcess(Buffer &in, Buffer &out) {
--- a/isis/src/base/apps/cubeatt/cubeatt.cpp
+++ b/isis/src/base/apps/cubeatt/cubeatt.cpp
@@ -6,7 +6,7 @@ using namespace std;
 using namespace Isis;
 
 namespace Isis {
-  void cubeatt_process(Buffer &in, Buffer &out);
+  void cubeattProcess(Buffer &in, Buffer &out);
 
   void cubeatt(UserInterface &ui) {
     Cube icube;
@@ -18,26 +18,36 @@ namespace Isis {
     cubeatt(&icube, ui);
   }
 
+
   void cubeatt(Cube *icube, UserInterface &ui) {
+    bool propTables = ui.GetBoolean("PROPTABLES");
+    QString outputFileName = ui.GetFileName("TO");
+    CubeAttributeOutput outputAttributes= ui.GetOutputAttribute("TO");
+    cubeatt(icube, outputFileName, outputAttributes, propTables);
+  }
+
+
+  void cubeatt(Cube *icube, QString outputCubePath, CubeAttributeOutput outAttributes, bool propTables) {
     // We will be processing by line
     ProcessByLine p;
 
     // Should we propagate tables
-    if(ui.GetBoolean("PROPTABLES")) {
+    if(propTables) {
       p.PropagateTables(false);
     }
 
     // Setup the input and output cubes
     p.SetInputCube(icube);
-    p.SetOutputCube(ui.GetFileName("TO"), ui.GetOutputAttribute("TO"), 
+    p.SetOutputCube(outputCubePath, outAttributes, 
                     icube->sampleCount(), icube->lineCount(), icube->bandCount());
 
-    p.StartProcess(cubeatt_process);
+    p.StartProcess(cubeattProcess);
     p.EndProcess();
   }
 
+
   // Line processing routine
-  void cubeatt_process(Buffer &in, Buffer &out) {
+  void cubeattProcess(Buffer &in, Buffer &out) {
     // Loop and copy pixels in the line.
     for(int i = 0; i < in.size(); i++) {
       out[i] = in[i];
--- a/isis/src/base/apps/cubeatt/cubeatt.cpp
+++ b/isis/src/base/apps/cubeatt/cubeatt.cpp
@@ -32,9 +32,7 @@ namespace Isis {
     ProcessByLine p;
 
     // Should we propagate tables
-    if(propTables) {
-      p.PropagateTables(false);
-    }
+    p.PropagateTables(propTables);
 
     // Setup the input and output cubes
     p.SetInputCube(icube);
--- a/isis/src/base/apps/cubeatt/cubeatt.cpp
+++ b/isis/src/base/apps/cubeatt/cubeatt.cpp
@@ -1,0 +1,46 @@
+#include "cubeatt.h"
+
+#include "ProcessByLine.h"
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+  void cubeatt_process(Buffer &in, Buffer &out);
+
+  void cubeatt(UserInterface &ui) {
+    Cube icube;
+    CubeAttributeInput inAtt = ui.GetInputAttribute("FROM");
+    if (inAtt.bands().size() != 0) {
+      icube.setVirtualBands(inAtt.bands());
+    }
+    icube.open(ui.GetFileName("FROM"));
+    cubeatt(&icube, ui);
+  }
+
+  void cubeatt(Cube *icube, UserInterface &ui) {
+    // We will be processing by line
+    ProcessByLine p;
+
+    // Should we propagate tables
+    if(ui.GetBoolean("PROPTABLES")) {
+      p.PropagateTables(false);
+    }
+
+    // Setup the input and output cubes
+    p.SetInputCube(icube);
+    p.SetOutputCube(ui.GetFileName("TO"), ui.GetOutputAttribute("TO"), 
+                    icube->sampleCount(), icube->lineCount(), icube->bandCount());
+
+    p.StartProcess(cubeatt_process);
+    p.EndProcess();
+  }
+
+  // Line processing routine
+  void cubeatt_process(Buffer &in, Buffer &out) {
+    // Loop and copy pixels in the line.
+    for(int i = 0; i < in.size(); i++) {
+      out[i] = in[i];
+    }
+  }
+}
--- a/isis/src/base/apps/topds4/topds4.cpp
+++ b/isis/src/base/apps/topds4/topds4.cpp
@@ -5,10 +5,12 @@
 #include <inja/inja.hpp>
 
 #include "md5wrapper.h"
-#include "QFile.h"
+#include <QFile>
 
 #include "topds4.h"
+#include "cubeatt.h"
 
+#include "CubeAttribute.h"
 #include "FileName.h"
 #include "OriginalLabel.h"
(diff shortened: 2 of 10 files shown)


isis/src/base/apps/topds4/topds4.cpp
```diff
@@ -31,9 +33,17 @@ namespace Isis {
     Process p;
     p.SetInputCube(icube);
 
-    // NEED TO WRITE AND CLOSE THE OUTPUT FILE BEFORE RENDERING SO THE FILE SIZE CALLBACK CAN GET THE FINAL FILE SIZE
     QString outputFile = ui.GetFileName("TO");
 
+    // NEED TO WRITE AND CLOSE THE OUTPUT FILE BEFORE RENDERING SO THE FILE SIZE CALLBACK CAN GET THE FINAL FILE SIZE
+
+    // Need a better solution for this.
+    QString outputCubePath = outputFile+".cub";
```
krlberry: How to handle one output filepath indicated by "TO" vs. two output files: the label and the cube itself. `tgocassisrdrgen` uses the one input for both and appends ".xml" or ".img" as needed. Thoughts for this application?

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+  // Test the label
+  Cube outputCube(cubePath);
+  Pvl *label = outputCube.label();
+  PvlGroup &pixelGroup = label->findObject("IsisCube").findObject("Core").findGroup("Pixels");
+  EXPECT_EQ(QString(pixelGroup["Type"]).toStdString(), "UnsignedByte");
+  EXPECT_NEAR(pixelGroup["Base"][0].toDouble(), -0.003952569, 0.00001);
+  EXPECT_NEAR(pixelGroup["Multiplier"][0].toDouble(), 0.003952569, 0.00001);
+
+  // Test the DNs
```
krlberry, line 31: I need to add this part of the tests still.

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "gmock/gmock.h"
+
+
+using namespace Isis;
+
+static QString APP_XML = FileName("$ISISROOT/bin/xml/cubeatt.xml").expanded();
+
+// Tests setting output attributes: bit type and range
+TEST_F(SmallCube, FunctionalTestCubeattBitttypeAndRange) {
+  QString cubePath = "/scratch/without.cub+8bit+0.0:1.0";
```
krlberry: Oops. I'm fixing this now.

isis/src/base/apps/topds4/topds4.cpp
jessemapel: I would have to specify the xml label and then strip that off to get the cube name. Ideally it would check if the extension is .xml, .cub, or neither. Then it will strip and add extensions as needed.

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,71 @@
...
+  UserInterface options(APP_XML, args);
+  cubeatt(options);
+
+  // Test the label
+  Cube outputCube(cubePath);
+  Pvl *label = outputCube.label();
+  PvlGroup &pixelGroup = label->findObject("IsisCube").findObject("Core").findGroup("Pixels");
+  EXPECT_EQ(QString(pixelGroup["Type"]).toStdString(), "UnsignedByte");
+  EXPECT_NEAR(pixelGroup["Base"][0].toDouble(), -0.003952569, 0.00001);
+  EXPECT_NEAR(pixelGroup["Multiplier"][0].toDouble(), 0.003952569, 0.00001);
```
jessemapel: Use the Cube accessors for these instead of text parsing the label. Also, I would not check the base and multiplier, just check the pixel type, setting the base and multiplier is the responsibility of Cube, not cubeatt.

isis/tests/FunctionalTestsCubeatt.cpp
jessemapel: I would pull the Cube statistics and check that the min/max values are between the specified values in the cubeatt call. The small cube fixture has values ranging from 0 to 1000, so I would pick something other than 0:1 for your range in this test. Maybe something like 100:460 that will cover about half of the pixels?

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  // Test the label
+  Cube outputCube(cubePath);
+  Pvl *label = outputCube.label();
+  PvlGroup &pixelGroup = label->findObject("IsisCube").findObject("Core").findGroup("Pixels");
+  EXPECT_EQ(QString(pixelGroup["Type"]).toStdString(), "Real");
+  EXPECT_EQ(pixelGroup["Base"][0].toDouble(), 0);
+  EXPECT_EQ(pixelGroup["Multiplier"][0].toDouble(), 1);
+
+  // Test that DNs match in the input and output cubes
```
jessemapel, line 55: I am confused about what this test is checking. It seems to be checking that if you don't specify any attributes that the output is the same as the input, but then it does a band selection on the input cube?

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,71 @@
...
+  UserInterface options(APP_XML, args);
+  cubeatt(options);
+
+  // Test the label
+  Cube outputCube(cubePath);
+  Pvl *label = outputCube.label();
+  PvlGroup &pixelGroup = label->findObject("IsisCube").findObject("Core").findGroup("Pixels");
+  EXPECT_EQ(QString(pixelGroup["Type"]).toStdString(), "Real");
+  EXPECT_EQ(pixelGroup["Base"][0].toDouble(), 0);
+  EXPECT_EQ(pixelGroup["Multiplier"][0].toDouble(), 1);
```
jessemapel: Again use the accessors, don't parse the label text. You should check all of the attributes of the cubes:

* pixel type
* storage format
* label format
* storage order
* pixel range
* bands

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,71 @@
...
+  UserInterface options(APP_XML, args);
+  cubeatt(options);
+
+  // Test the label
+  Cube outputCube(cubePath);
+  Pvl *label = outputCube.label();
+  PvlGroup &dimensionsGroup = label->findObject("IsisCube").findObject("Core").findGroup("Dimensions");
+
+  // Test the label
+  EXPECT_EQ(QString(dimensionsGroup["Bands"]).toInt(), 9);
```
jessemapel: Again just do outputCube.bandCount().

jessemapel (commented):
Conversion looks good. I added some comments about how to improve the tests and how to check DNs

isis/tests/FunctionalTestsCubeatt.cpp
krlberry: I always forget I can just do this. Thanks, @jessemapel

isis/tests/FunctionalTestsCubeatt.cpp
krlberry: Good point! The old test did both of these things, too, which I was confused by. I think to be actually testing what I think it needs to be testing, I'm going to take out the band selection on the input cube.

isis/src/base/apps/cubeatt/cubeatt.cpp
```diff
@@ -0,0 +1,56 @@
...
+    cubeatt(icube, outputFileName, outputAttributes, propTables);
+  }
+
+
+  void cubeatt(Cube *icube, QString outputCubePath, CubeAttributeOutput outAttributes, bool propTables) {
+    // We will be processing by line
+    ProcessByLine p;
+
+    // Should we propagate tables
+    if(propTables) {
```
scsides: This seems backwards.

isis/src/base/apps/topds4/topds4.cpp
scsides: Most if not all of the apps that deal with .img .xml .lbl, ... ask for the label file then change the extension to get at the other files.

isis/tests/FunctionalTestsCubeatt.cpp
krlberry: When set, the pixel range is just used by the cube class to set the base and multiplier, so checking those is a way to check that the min/max was set appropriately.

isis/src/base/apps/cubeatt/cubeatt.cpp
krlberry: Good catch! Also, I wrote this weirdly anyway. I've cleaned it up now to just do: `p.PropagateTables(propTables);`

isis/tests/FunctionalTestsCubeatt.cpp
krlberry: Checking base and multiplier is kind of a proxy for testing that the pixel range was set. I couldn't find a way to retrieve the min/max for the pixel range set besides checking the base and multiplier, since that's what the cube does when you call `Cube.setMinMax()`

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,83 @@
...
+  UserInterface options(APP_XML, args);
+  cubeatt(options);
+
+  Cube outputCube(cubePath);
+
+  // Check attributes: pixel type, storage format, label format, storage order, pixel range, bands
+  EXPECT_EQ(outputCube.pixelType(), PixelType::UnsignedByte);
+  EXPECT_EQ(outputCube.format(), Cube::Format::Tile);
+  EXPECT_TRUE(outputCube.labelsAttached());
+  EXPECT_EQ(outputCube.byteOrder(), ByteOrder::Lsb);
```
jessemapel: You shouldn't check every attribute in every test, you should only check the attributes that this test is working with. So, here I would not check format, attached label, or byte order.

isis/tests/FunctionalTestsCubeatt.cpp
```diff
@@ -0,0 +1,83 @@
...
+  Cube outputCube(cubePath);
+  // Check attributes: pixel type, storage format, label format, storage order, pixel range, bands
+  EXPECT_EQ(outputCube.pixelType(), PixelType::Real);
+  EXPECT_EQ(outputCube.format(), Cube::Format::Tile);
+  EXPECT_TRUE(outputCube.labelsAttached());
+  EXPECT_EQ(outputCube.byteOrder(), ByteOrder::Lsb);
+  // Setting the pixel range modifies the base/multiplier, so check those.
+  EXPECT_EQ(outputCube.base(), 0);
+  EXPECT_EQ(outputCube.multiplier(), 1);
+  EXPECT_EQ(outputCube.bandCount(), 9);
```
jessemapel, line 71: Same as the first test, only check what this test is testing for. In this case, only look at the bands in the output cube.

isis/tests/FunctionalTestsCubeatt.cpp
krlberry: ~~Sure, I can do that. My logic was that it was useful to test both the attributes that should change, and also the attributes that should remain the same, to make sure each is working as expected.~~

krlberry: Oops, the functionality I just mentioned is definitely already being tested by the `noChange` test, nevermined.

isis/tests/Fixtures.cpp
```diff
@@ -31,6 +32,25 @@ namespace Isis {
...
+    originalBand += "5";
+    originalBand += "6";
+    originalBand += "7";
+    originalBand += "8";
+    originalBand += "9";
+    originalBand += "10";
+    bandBin += originalBand;
+    cubeLabel.addGroup(bandBin);
+    testCube->close();
+    testCube->open(path);
```
jessemapel: This needs to be re-opened as read/write. It's causing an error when other tests try adding stuff to its label.

jessemapel (commented):
Looks good, just one issue that's causing other tests to fail.

jessemapel:
Also, when we do test conversion we are deleting the old test makefiles

isis/tests/Fixtures.cpp
krlberry: Thanks for catching this!

jessemapel (approved):

PR #4261, opened 2021-01-27
Added resolution and radii to CSM camera model

jessemapel:
Replaced hardcoded resolutions with proper resolution calculation using the CSM API. Also, added the proper ellipsoid stuff. It currently only works for models that are settable ellipsoids also. Do we just hardcode WGS84 radii for non-setttable ellipsoid models? That is what the API docs say should be assumed.

### DIFF
--- a/isis/src/base/objs/CSMCamera/CSMCamera.cpp
+++ b/isis/src/base/objs/CSMCamera/CSMCamera.cpp
@@ -23,17 +23,20 @@
 #include "IException.h"
 #include "IString.h"
 #include "iTime.h"
+#include "LinearAlgebra.h"
 #include "NaifStatus.h"
 #include "SpecialPixel.h"
 #include "StringBlob.h"
 
 #include "csm/Plugin.h"
+#include "csm/Ellipsoid.h"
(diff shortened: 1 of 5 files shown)


jessemapel:
From wikipedia, the WGS84 radii are:

major: 6,378,137 m
minor: 6,356,752.3142 m

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -116,32 +125,46 @@ namespace Isis {
...
+    return sqrt(imagePartials[1]*imagePartials[1] +
+                imagePartials[3]*imagePartials[3] +
+                imagePartials[5]*imagePartials[5]);
   }
 
- double CSMCamera::PixelResolution() {
-    // Camera version uses detectorMap
-    return m_pixelPitchY;
+  double CSMCamera::PixelResolution() {
+    // Redo the line and sample resolution calculations because it avoids
```
krlberry, line 145: :+1:

krlberry (approved):
:+1: Looks good! I'm going to merge this.

PR #4263, opened 2021-01-28
Changed pseudoinverse computation to armadillo SVD based

jessemapel:
Previous implementation did not allow you to compute this for all matrices. This should not work for everything at the cost of being a little slower.

### DIFF
--- a/isis/src/base/objs/LinearAlgebra/LinearAlgebra.cpp
+++ b/isis/src/base/objs/LinearAlgebra/LinearAlgebra.cpp
@@ -355,8 +355,8 @@ namespace Isis {
   LinearAlgebra::Matrix LinearAlgebra::pseudoinverse(const Matrix &matrix) {
     // Copy values into Armadillo matrix
     arma::mat arMat(matrix.size1(), matrix.size2());
-    for (int i = 0; i < matrix.size1(); i++) {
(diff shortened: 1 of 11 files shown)


isis/src/base/objs/LinearAlgebra/LinearAlgebra.cpp
```diff
@@ -343,36 +346,32 @@ namespace Isis {
...
-      // Otherwise, more columns than rows, compute a right inverse
-      // M^T * (M * M^T)^-1
-      else {
-        return multiply(transpose(matrix), inverse(multiply(matrix, transpose(matrix))));
+  LinearAlgebra::Matrix LinearAlgebra::pseudoinverse(const Matrix &matrix) {
+    // Copy values into Armadillo matrix
+    arma::mat arMat(matrix.size1(), matrix.size2());
+    for (int i = 0; i < matrix.size1(); i++) {
+      for (int j = 0; j < matrix.size2(); j++) {
+        arMat(i, j) = matrix(i, j);
```
jessemapel, line 360: There is likely an easier way to do these data copies with just accessing the raw data array but the boost storage format can be finicky and I don't feel like figuring out the code needed to keep it in line with Armadillo.

PR #4266, opened 2021-01-31
Modified CSM setImage to intersect the shape model

jessemapel:
Changed setImage to intersect the shape model. I also added some helpers to ensure that we properly handle CSM working in meters vs ISIS working in kilometers. This was causing the resolution calculations to fail because setImage wasn't scaling correctly.

### DIFF
--- a/isis/src/base/objs/CSMCamera/CSMCamera.cpp
+++ b/isis/src/base/objs/CSMCamera/CSMCamera.cpp
@@ -88,44 +88,50 @@ namespace Isis {
   }
 
 
+  // TODO make this an iterative technique that updates the locus each iteration
   bool CSMCamera::SetImage(const double sample, const double line) {
+    csm::ImageCoord imagePt;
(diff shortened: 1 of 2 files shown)


isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -88,46 +88,50 @@ namespace Isis {
...
 
-        // fill in whatever stuff ISIS needs from this
-//    }
-//    else {
-        // handle projected case
-//    }
+    // ISIS sensors work in Kilometers, CSM works in meters
+    std::vector<double> obsPosition = {imageLocus.point.x / 1000.0,
+                                       imageLocus.point.y / 1000.0,
+                                       imageLocus.point.z / 1000.0};
```
jessemapel, line 124: I would like to put this through the csm->ISIS conversion but I'm concerned about it being slow and it still requires making a surface point and specifying that it's in meters. So, it's still prone to errors for unit conversions.

krlberry (approved):
:+1: This looks good to me!

scsides (approved):

PR #4267, opened 2021-02-01
added initial CSMCamera test

jessemapel:
Created mock csm::RasterGM objects to test our API calls through Gmock. Also, added a new plugin that allows us to register the mocks so that the CSMCamera class can get them properly.

Wrote a simple test for SetImage. We ran out of time before we could get the Matcher stuff for csm::ImageCoord in the imaging locus call working. It will be in a follow on PR.

Done with @scsides

### DIFF
--- a/isis/src/base/objs/CSMCamera/CSMCamera.cpp
+++ b/isis/src/base/objs/CSMCamera/CSMCamera.cpp
@@ -48,6 +48,8 @@ namespace Isis {
     cube.read(stateString);
     PvlObject &blobLabel = stateString.Label();
     QString pluginName = blobLabel.findKeyword("PluginName")[0];
+    // TODO check that we can actually make a model with the appropriate name from the model state
+    // before doing it
     const csm::Plugin *plugin = csm::Plugin::findPlugin(pluginName.toStdString());
     m_model = dynamic_cast<csm::RasterGM*>(plugin->constructModelFromState(stateString.string()));
     // If the dynamic cast failed, raise an exception
@@ -61,12 +63,7 @@ namespace Isis {
     m_spacecraftNameLong = QString::fromStdString(m_model->getPlatformIdentifier());
     m_spacecraftNameShort = QString::fromStdString(m_model->getPlatformIdentifier());
 
-    if (m_target) {
-      delete m_target;
-      m_target = nullptr;
-    }
-
-    m_target = setTarget(*cube.label());
+    setTarget(*cube.label());
 
     std::cout << "Target status" << std::endl;
     if (target()) {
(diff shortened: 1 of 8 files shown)


jessemapel:
I've added more tests and some matchers to match the CSM structs. This should cover the existing code.

isis/tests/CSMCameraTests.cpp
```diff
@@ -0,0 +1,311 @@
...
+  //
+  // is
+  // -17  8
+  //  -2  2  *  1/18
+  //  13 -4
+  EXPECT_CALL(mockModel, computeGroundPartials)
+      .Times(6)
+      .WillRepeatedly(::testing::Return(std::vector<double>{1, 2, 3, 4, 5, 6}));
+
+  // We also have to set the mock up for setimage
```
jessemapel, line 289: If we end up having to duplicate this code a lot to get a camera with a set image/ground point then we can move it into a new fixture.

scsides (approved):

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -383,28 +380,28 @@ namespace Isis {
...
-    if (!ellipsoidModel) {
-      // TODO is there a fallback we can do here?
-      QString msg = "Failed to get ellipsoid from CSM Model.";
-      throw IException(IException::Programmer, msg, _FILEINFO_);
-    }
-    csm::Ellipsoid targetEllipsoid = ellipsoidModel->getEllipsoid();
-    std::vector<Distance> radii {Distance(targetEllipsoid.getSemiMajorRadius(), Distance::Meters),
-                                 Distance(targetEllipsoid.getSemiMajorRadius(), Distance::Meters),
-                                 Distance(targetEllipsoid.getSemiMinorRadius(), Distance::Meters)};
+    csm::Ellipsoid targetEllipsoid = csm::SettableEllipsoid::getEllipsoid(m_model);
```
krlberry, line 390: Does this throw its own error if it fails and that's why the ISIS exception could be removed?

jessemapel: This returns the WGS84 ellipsoid if it fails to convert the model to a settableellipsoid. So, this won't raise an exception.

isis/tests/MockCsmPlugin.cpp
```diff
@@ -0,0 +1,261 @@
...
+ * @param modelName the model name
+ * @param modelState the model state
+ * @param warnings the warning list
+ *
+ * @return bool if the model can be constructed from the state
+ */
+bool MockCsmPlugin::canModelBeConstructedFromState(const std::string& modelName,
+                                    const std::string& modelState,
+                                    csm::WarningList* warnings) const {
+  return m_registeredModels.contains(modelState);
```
krlberry, line 130: I'm a little confused about this. `contains(modelName)` would make more sense to me here.

jessemapel: The modelState is the modelName in this case. We want to check the modelState, though because that is what actually contains the data.

kberryUSGS: Could you add a comment that the model state is just the model name here?

krlberry (approved):
I had a couple of questions, but looks good!

isis/tests/MockCsmPlugin.cpp
krlberry: Isn't `m_registeredPlugin` just a map of model name, to actual model object, though? So how would contains work with a whole modelState here (if it weren't the model name)? I might just be missing the point, though, because it is the model name here, and as such, this is a good test to see if the model can be constructed from the state.

PR #4282, opened 2021-02-10
Updated CSMCamera doc strings

jessemapel:
Also documented the other new functions we added elsewhere.

### DIFF
--- a/isis/src/base/objs/CSMCamera/CSMCamera.cpp
+++ b/isis/src/base/objs/CSMCamera/CSMCamera.cpp
@@ -44,8 +44,11 @@ using namespace std;
 
 namespace Isis {
   /**
-   * Constructor for the USGS CSM Camera Model inside ISIS.
+   * Constructor for an ISIS Camera model that uses a Community Sensor Model (CSM)
+   * for the principal transformations.
    *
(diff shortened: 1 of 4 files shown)


isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -603,41 +830,85 @@ namespace Isis {
...
+   * @param[out] p The position of the sun
+   */
   void CSMCamera::sunPosition(double p[3]) const {
     QString msg = "Sun position is not supported for CSM camera models";
     throw IException(IException::Programmer, msg, _FILEINFO_);
   }
 
 
+  /**
+   * Get the SpicePosition object the contains the state information for the sun in J2000.
```
krlberry: Single typo: Potentially update the first "the" to "that".

krlberry (approved):
:+1: Great docs! Just found one typo in all of this.

isis/src/base/objs/CSMCamera/CSMCamera.cpp
jessemapel: Fixed, thanks for catching that.

scsides (approved):

PR #4291, opened 2021-02-16
Adds CSM camera model

jessemapel:
## Description

The main thrust of this feature branch is to add a camera model that is a wrapper around a CSM::RasterGM. The main class for this is CSMCamera. The primary difficulties with this are:

* The ISIS camera model expects to have access to the interior orientation (E.G. pixel size, focal length) of the sensor. CSM, does not expect this because it supports things like replacement sensor models that are designed to specifically obscure the internals of the sensor for security purposes.
* The ISIS camera model expects all of the sensor states and orientations to be relative to the universal J2000 reference frame. CSM only operates in the target body reference frame. This means, it cannot be related to the universal J2000 reference frame or to the internal sensor reference frame.

We worked around these two issues by aggressively overriding methods from the camera model hierarchy to either work around or throw an exception when other code attempts to use the camera model in these ways. We discussed making further changes to the existing camera model hierarchy to make it more flexible, but we decided that was too risky and instead isolated the code in the CSMCamera class.

This means that CSM sensor models cannot be used in all ISIS applications, but they can be used in many of them. Specifically, we targeted cam2map and campt. We expect that more modifications similar to the cmapt modifications will be required to make the CSM sensor models fully supported in ISIS.

All of this took many changes in several different parts of the code, here they are broken down by section:

### CSMCamera
* As stated previously, we promoted many parts of the Spice, Sensor, and Camera classes to protected or virtual so that we could override them.
* We had to modify the Spice and Target constructors to allow for a sensor model without any SPICE data at all. In order to avoid modifying existing execution paths, we "sprouted" a new initialization path for CSMCamera and any future sensor models that lack SPICE data.
* We had to add a pseudoinverse method to compute the pixel resolution. This can be found in the LinearAlgebra file.
* CSM time uses ISO 8601, so we updated the iTime class to better support it in iTime::fromUtc.

### CameraFactory
* We made the CameraFactory's method that initializes both the CSM plugins and the ISIS camera model plugins public and moved it towards being a singleton instead of a set of static functions.
* We added a new branch in CameraFactory::Create for CSMCamera because it does not use the normal ISIS Camera model plugin scheme.
* We extended the existing Cube::hasTable method to a more generic Cube::hasBlob method so that we could check for the CSMState BLOB.

### csminit and spiceinit
* We had to add some extra information to csminit that the camera model hierarchy expects, target name and instrument ID.
* We tweaked csminit and spiceinit so that they can be run one after another and they will clean-up each-other's changes.
* We updated csminit to revert any changes it has made on failure.
* We added the ability to pass a state string instead of an ISD to csminit.

### Testing
* We updated the existing csminit test models to be fully CSM::RasterGM subclasses instead of just CSM::GeometricModel subclasses.
* We added a new test CSM plugin class that allows us to register specific CSM::RasterGM objects and access them later.
* We added a CSM::RasterGM Mock.

## Related Issue

CSM in ISIS project

## Motivation and Context

This allows for much greater interoperability with other software packages that support CSM (E.G. Socet GXP, ASP).

## How Has This Been Tested?

We tested with our USGSCSM library and also added automated tests to test all new functionality.

## Types of changes

## Checklist:

## Licensing


jessemapel:
There is 1 expected test failure: phocubes mosaic errors. This is caused by the update to CameraFactory::Create's error message. We have new truth data ready to be checked in once this is merged.

~There is also 1 more commit coming with the CHANGELOG entry and a small tweak to csminit.~

isis/src/base/objs/CameraFactory/CameraFactory.cpp
```diff
@@ -30,16 +30,23 @@
 #include "CameraFactory.h"
 
 #include "Camera.h"
+#include "CSMCamera.h"
 #include "FileName.h"
 #include "IException.h"
 #include "Plugin.h"
 #include "Preference.h"
+#include "StringBlob.h"
```
acpaquette: I don't think this import is necessary here(?)

isis/src/base/objs/CameraFactory/CameraFactory.h
```diff
@@ -81,6 +80,7 @@ namespace Isis {
       ~CameraFactory() {};
 
       static Plugin m_cameraPlugin;   //!< The plugin file for the camera
+      static bool m_initialized; //!<Has the plugin list been initialized
```
acpaquette, line 70: Needs a history comment

isis/src/base/objs/CameraPointInfo/CameraPointInfo.cpp
```diff
@@ -405,17 +403,27 @@ namespace Isis {
...
-        gp->findKeyword("RightAscension").setValue(toString(
-                        m_camera->RightAscension()), "DEGREE");
-        gp->findKeyword("Declination").setValue(toString(
-                        m_camera->Declination()), "DEGREE");
+        try {
+          gp->findKeyword("RightAscension").setValue(toString(
+                          m_camera->RightAscension()), "DEGREE");
+        }
+        catch (IException &e) {
+          gp->findKeyword("RightAscension").setValue("Null");
```
acpaquette, line 402: I don't know if it's possible but it may be better to have a keyword by default be populated with "Null". I can understand why it wouldn't be possible but something to consider rather than having to set it ourselves all over the place.

isis/src/base/objs/Cube/Cube.h
```diff
@@ -315,6 +315,7 @@ namespace Isis {
       PvlGroup &group(const QString &group) const;
       bool hasGroup(const QString &group) const;
       bool hasTable(const QString &name);
+      bool hasBlob(const QString &type, const QString &name);
```
acpaquette, line 303: Needs history comment

isis/src/base/objs/LinearAlgebra/LinearAlgebra.h
```diff
@@ -163,6 +163,7 @@ namespace Isis {
       static Matrix identity(int size);
       static Matrix transpose(const Matrix &matrix);
       static Matrix inverse(const Matrix &matrix);
+      static Matrix pseudoinverse(const Matrix &matrix);
```
acpaquette, line 151: Needs history comment

isis/src/base/objs/LinearAlgebra/LinearAlgebra.cpp
```diff
@@ -342,11 +345,41 @@ namespace Isis {
...
 
 
+  /**
+   * Returns the pseudoinverse of a matrix.
+   *
+   * @param matrix The matrix to compute the pseudoinverse of.
+   *
+   * @return @b LinearAlgebra::Matrix The pseudoinverse matrix.
+   */
+  LinearAlgebra::Matrix LinearAlgebra::pseudoinverse(const Matrix &matrix) {
```
acpaquette, line 339: Sucks that boost doesn't have this :(

isis/src/base/objs/Sensor/Sensor.cpp
```diff
@@ -574,7 +575,9 @@ namespace Isis {
    * @return @b double The angle of right ascension, in degrees.
    */
   double Sensor::RightAscension() {
-    if (m_newLookB) computeRaDec();
+    if (m_newLookB) {
+      computeRaDec();
+    }
     return m_ra;
   }
```
acpaquette, line 566: Is computeRaDec supposed to be called first whenever the RightAscension is asked for or is the newLookB needed for something else?

isis/src/base/objs/Spice/Spice.cpp
```diff
@@ -78,21 +78,30 @@ namespace Isis {
   // TODO: DOCUMENT EVERYTHING
   Spice::Spice(Cube &cube) {
     Pvl &lab = *cube.label();
-    PvlGroup kernels = lab.findGroup("Kernels", Pvl::Traverse);
-    bool hasTables = (kernels["TargetPosition"][0] == "Table");
-    init(lab, !hasTables);
+    // TODO: update to check to see if it has a state string
```
acpaquette: Remove comment(?)

isis/src/base/objs/Spice/Spice.cpp
```diff
@@ -78,21 +78,30 @@ namespace Isis {
...
    */
-  Spice::Spice(Cube &cube, bool noTables) {
-    init(*cube.label(), noTables);
-  }
-
+// NO LONGER USED
+//  Spice::Spice(Cube &cube, bool noTables) {
+//    init(*cube.label(), noTables);
+//  }
+//
```
acpaquette: Remove(?)

isis/src/base/objs/Sensor/Sensor.h
```diff
@@ -207,22 +207,22 @@ namespace Isis {
       Distance LocalRadius(Latitude lat, Longitude lon);
       Distance LocalRadius(double lat, double lon);
 
-      double PhaseAngle() const;
-      double EmissionAngle() const;
-      double IncidenceAngle() const;
+      virtual double PhaseAngle() const;
+      virtual double EmissionAngle() const;
+      virtual double IncidenceAngle() const;
```
acpaquette, line 199: Needs history comment

isis/src/base/objs/Spice/Spice.h
```diff
@@ -275,72 +275,72 @@ namespace Isis {
...
-      Spice(Cube &cube, bool noTables);
+//      Spice(Cube &cube, bool noTables); NO LONGER USED
       Spice(Pvl &lab, nlohmann::json);
 
       // destructor
       virtual ~Spice();
 
       // Methods
-      void setTime(const iTime &time);
+      virtual void setTime(const iTime &time);
```
acpaquette, line 293: Needs history comment

isis/src/base/objs/StringBlob/StringBlob.cpp
```diff
@@ -46,16 +46,26 @@ namespace Isis {
...
+//    char *temp = p_buffer;
+//    p_buffer = new char[p_nbytes+bytes];
+
+//    if (temp != NULL) memcpy(p_buffer, temp, p_nbytes);
+//    const char *ptr = m_string.c_str();
+//    memcpy(&p_buffer[p_nbytes], (void *)ptr, bytes);
+//    p_nbytes += bytes;
 
-    if (temp != NULL) delete [] temp;
+//    if (temp != NULL) delete [] temp;
```
acpaquette: Remove(?)

isis/src/base/objs/Target/Target.h
```diff
@@ -78,6 +78,7 @@ namespace Isis {
     public:
       // constructors
       Target(Spice *spice, Pvl &label);
+      Target(Pvl &label);
```
acpaquette, line 68: Needs history comment

isis/src/base/objs/iTime/iTime.cpp
```diff
@@ -443,6 +445,29 @@ namespace Isis {
   }
 
   void iTime::setUtc(QString utcString) {
+    // If the time string is in ISO basic format add separators for utc2et
```
acpaquette, line 433: Need history comment

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -0,0 +1,960 @@
...
+      return false;
+    }
+
+    // If we are here then everything went well so save the pixel and return true
+    m_lookB[0] = locusVec[0];
+    m_lookB[1] = locusVec[1];
+    m_lookB[2] = locusVec[2];
+    m_newLookB = true;
+    p_pointComputed = true;
+    // TODO do we need to apply the alpha cube?
```
acpaquette: Bump

isis/src/base/objs/Camera/Camera.h
```diff
@@ -262,8 +262,8 @@ namespace Isis {
       virtual bool SetUniversalGround(const double latitude, const double longitude);
       virtual bool SetUniversalGround(const double latitude, const double longitude,
                                       const double radius);
-      bool SetGround(Latitude latitude, Longitude longitude);
-      bool SetGround(const SurfacePoint & surfacePt);
+      virtual bool SetGround(Latitude latitude, Longitude longitude);
```
acpaquette, line 249: Need history comment

isis/src/base/objs/Spice/Spice.h
```diff
@@ -275,72 +275,72 @@ namespace Isis {
...
    *                           string. As such, when used sct2e_c is used to convert to an ET rather
-   *                           than scs2e_c. 
+   *                           than scs2e_c.
    */
   class Spice {
     public:
       // constructors
       Spice(Cube &cube);
-      Spice(Cube &cube, bool noTables);
+//      Spice(Cube &cube, bool noTables); NO LONGER USED
```
AustinSanders: Is there any reason to leave this here?

isis/tests/CamptFunctionalTests.cpp
```diff
@@ -260,3 +264,56 @@ TEST_F(DefaultCube, FunctionalTestCamptAllowOutside) {
...
+
+  double pointRadius = csm::Ellipsoid().getSemiMajorRadius();
+  // We don't care exactly how the Mock is called so just setup return values
+  EXPECT_CALL(mockModel, getSensorIdentifier)
+      .WillRepeatedly(::testing::Return("MockSensorID"));
+  EXPECT_CALL(mockModel, getPlatformIdentifier)
+      .WillRepeatedly(::testing::Return("MockPlatformID"));
+  EXPECT_CALL(mockModel, getReferenceDateAndTime)
+      .WillRepeatedly(::testing::Return("20000101T115855.816"));
+EXPECT_CALL(mockModel, imageToRemoteImagingLocus)
```
AustinSanders: Indent is off

isis/src/base/apps/csminit/csminit.cpp
```diff
@@ -46,101 +48,188 @@ namespace Isis {
...
+
+    else if (ui.WasEntered("STATE")) {
+      FileName stateFilePath = ui.GetFileName("STATE");
+
+      std::ifstream file(stateFilePath.expanded().toStdString());
+      std::stringstream buffer;
+      buffer << file.rdbuf();
+      QString stateString = QString::fromStdString(buffer.str());
+
+      if (!ui.WasEntered("PLUGINNAME") && !ui.WasEntered("MODELNAME")) {
```
AustinSanders, line 171: If both PLUGINNAME and MODELNAME are required, shouldn't this be either 

```
if (!(ui.WasEntered("PLUGINNAME" && ui.WasEntered("MODELNAME")))
```
or
```
if (!ui.WasEntered("PLUGINNAME") || !ui.WasEntered("MODELNAME"))
```

I think that, as it's written, this will only throw an exception if both parts are missing, but not if one is missing.

isis/src/base/apps/csminit/csminit.cpp
```diff
@@ -46,101 +48,188 @@ namespace Isis {
...
+      originalKernels = cube->group("Kernels");
     }
 
-    string modelState = model->getModelState();
+    if (cube->hasGroup("CsmInfo")) {
+      originalCsmInfo = cube->group("CsmInfo");
+    }
 
-    // Add the TargetName to the instrument group, if specified:
+    // If the user doesn't specify a target name, then we will still need
```
AustinSanders: Can we move this closer to the block of code it's describing?  ~line 222

isis/tests/iTimeTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+    std::make_pair("2003-05-02T12:15:01.1234", "2003122T121501.1234"),
+    std::make_pair("2003-05-02T12:15:01.1234", "2003-122T12:15:01.1234"),
+    std::make_pair("2003-01-02T12:15:01", "20030102T121501"),
+    std::make_pair("2003-01-02T12:15:01", "2003-01-02T12:15:01"),
+    std::make_pair("2003-01-02T12:15:00", "20030102T1215"),
+    std::make_pair("2003-01-02T12:15:00", "2003-01-02T12:15"),
+    std::make_pair("2003-01-02T12:00:00", "20030102T12"),
+    std::make_pair("2003-01-02T12:00:00", "2003-01-02T12"),
+    std::make_pair("2003-01-02T00:00:00", "20030102T"),
+    std::make_pair("2003-01-02T00:00:00", "2003-01-02T")));
```
AustinSanders: Github is complaining about the "no newline at end of file."

isis/src/base/objs/StringBlob/StringBlob.cpp
```diff
@@ -46,16 +46,26 @@ namespace Isis {
...
+      delete p_buffer;
+    }
+
+    p_buffer = new char[p_nbytes];
     const char *ptr = m_string.c_str();
-    memcpy(&p_buffer[p_nbytes], (void *)ptr, bytes);
-    p_nbytes += bytes;
+    memcpy(p_buffer, (void *)ptr, p_nbytes);
+
+//    int bytes = m_string.size();
```
AustinSanders: Can we clean up the commented out / original code, or are we intentionally keeping it around?

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -0,0 +1,960 @@
...
+          validBackProject = false;
+        }
+      }
+    }
+
+    // Check for occlusion
+    double line, sample;
+    csmToIsisPixel(imagePt, line, sample);
+    csm::EcefLocus imageLocus = m_model->imageToRemoteImagingLocus(imagePt);
+    // TODO should this be in meters or kilometers?
```
acpaquette: Bump

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -0,0 +1,960 @@
...
+    csm::Ellipsoid targetEllipsoid = csm::SettableEllipsoid::getEllipsoid(m_model);
+    std::vector<Distance> radii  = {Distance(targetEllipsoid.getSemiMajorRadius(), Distance::Meters),
+                                    Distance(targetEllipsoid.getSemiMajorRadius(), Distance::Meters),
+                                    Distance(targetEllipsoid.getSemiMinorRadius(), Distance::Meters)};
+    target->setRadii(radii);
+
+    // Target needs to be able to access the camera to do things like
+    // compute resolution
+    // TODO find a better way to do this. It would better if we could set this up
+    //      inside of the Target constructor so that it always has a Spice object.
```
acpaquette: Turn this into an issue

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -0,0 +1,960 @@
...
+   *
+   * This is not supported for CSM cameras because the time is a function of the
+   * image coordinate and the two cannot be changed independently.
+   *
+   * @param time The time to set
+   */
+  void CSMCamera::setTime(const iTime &time) {
+    QString msg = "Setting the image time is not supported for CSM camera models";
+    throw IException(IException::Programmer, msg, _FILEINFO_);
+  }
```
acpaquette, line 787: I don't know how desperately we need this functionality but couldn't this be done by some kind of binary search through the image?

isis/src/base/objs/CSMCamera/CSMCamera.h
```diff
@@ -0,0 +1,144 @@
+#ifndef CSMCamera_h
```
acpaquette, line 1: Could use a history section. Most of the .h files in this PR are missing history entries. I don't think we need to continue doing them with git as our version control but it's still something we need to talk about

isis/tests/CubeTests.cpp
```diff
@@ -231,3 +232,13 @@ TEST(CubeTest, TestCubeAttachSpiceFromIsd) {
...
+
+TEST_F(SmallCube, TestCubeHasBlob) {
+  // Make sure there isn't a TestBlob on here already
+  testCube->deleteBlob("String", "TestBlob");
+  StringBlob testBlob("Test String", "TestBlob");
+  testCube->write(testBlob);
+  EXPECT_TRUE(testCube->hasBlob("String", "TestBlob"));
+  testCube->deleteBlob("String", "TestBlob");
+  EXPECT_FALSE(testCube->hasBlob("String", "TestBlob"));
+}
```
acpaquette: I think this test could be simplified to check for a different blob? It seems to be doing a lot by deleting, creating, and writing a new string blob.

isis/src/base/objs/Spice/Spice.cpp
```diff
@@ -78,21 +78,30 @@ namespace Isis {
...
    * Constructs a Spice object.
    *
    * @param lab  Pvl labels.
    * @param noTables Indicates the use of tables.
    */
-  Spice::Spice(Cube &cube, bool noTables) {
-    init(*cube.label(), noTables);
-  }
-
+// NO LONGER USED
```
AustinSanders: Can this be removed if it's no longer used?

isis/tests/CSMCameraTests.cpp
```diff
@@ -0,0 +1,466 @@
...
+    testCam->setTime(iTime("2000-01-01T11:58:55.816"));
+  }
+  catch(Isis::IException &e)
+  {
+    EXPECT_TRUE(e.toString().toLatin1().contains("Setting the image time is not supported "
+        "for CSM camera models")) << e.toString().toStdString();
+  }
+  catch(...)
+  {
+      FAIL() << "Expected an IExcpetion with message \""
```
AustinSanders: IExcpetion -> IException

isis/tests/CSMCameraTests.cpp
```diff
@@ -0,0 +1,466 @@
...
+    testCam->subSolarPoint(lat ,lon);
+  }
+  catch(Isis::IException &e)
+  {
+    EXPECT_TRUE(e.toString().toLatin1().contains("Sub solar point is not supported "
+        "for CSM camera models")) << e.toString().toStdString();
+  }
+  catch(...)
+  {
+      FAIL() << "Expected an IExcpetion with message \""
```
AustinSanders: It looks like IExcpetion -> IException needs to happen on all the FAIL messages.

isis/tests/iTimeTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+TEST(iTimeTests, Comparison) {
+    QString beforeString("2003-01-02T12:15:01.1234");
+    QString afterString("2010-04-03T16:32:56.2487");
+
+    iTime beforeTime(beforeString);
+    iTime afterTime(afterString);
+
+    EXPECT_TRUE(beforeTime < afterTime);
+    EXPECT_FALSE(afterTime < beforeTime);
```
AustinSanders, line 81: Unless I'm missing something, it seems unnecessary to both test that (before < after == true) and tha (tafter < before == false).

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -0,0 +1,960 @@
...
+  bool CSMCamera::SetUniversalGround(const double latitude, const double longitude) {
+    return SetGround(
+        Latitude(latitude, Angle::Degrees),
+        Longitude(longitude, Angle::Degrees));
+  }
+
+
+/**
+   * Set the latitude, longitude, and radius for the Camera Model and then compute the
+   * cooresponding image time, look vector, and image coordinate.
```
AustinSanders: cooresponding -> corresponding throughout documentation

isis/tests/FunctionalTestsCsminit.cpp
```diff
@@ -322,9 +327,51 @@ TEST_F(DefaultCube, CSMinitSpiceCleanup) {
 }
 
 
+// This test uses the DefaultCube fixture because it has already has attached
```
acpaquette: Unfinished comment

isis/tests/TestCsmModel.cpp
```diff
@@ -484,4 +484,147 @@ std::vector<double> TestCsmModel::getCrossCovarianceMatrix(
   return covariance;
 }
 
+// csm::RasterGM methods
+csm::ImageCoord TestCsmModel::groundToImage(const csm::EcefCoord& groundPt,
```
acpaquette, line 488: I'm not really sure where the testcsmmodel is being used

acpaquette (commented):
A few smaller things to address and we also need to address whether we want to add history comments or not. Overall a solid PR

AustinSanders (changes requested):

isis/src/base/objs/CameraPointInfo/CameraPointInfo.cpp
jessemapel: This is tricky because PvlKeywords default to no value at all. I.E.

```
Group = SomeGroup
  SomeKeyword
EndGroup
```

There is also logic where values are iteratively appended to the keyword and that would be tricky if the default value is null, not empty.

I agree this is clunky. Maybe a simple function to set it to null? so these would be like

```
gp->findKeyword("RightAscension").nullify()
```

isis/src/base/objs/LinearAlgebra/LinearAlgebra.cpp
jessemapel: This is too complex of an operation for boost, it only supports BLAS operations. Computing the pseudoinverse requires SVD which is far beyond BLAS.

isis/src/base/objs/Sensor/Sensor.cpp
jessemapel: computeRaDec should be called if the look vector has changed, indicated by m_newLookB being true. It's basically a work around to avoid computing the RA and Dec everytime the camera model is set to a new thing.

isis/src/base/objs/CSMCamera/CSMCamera.cpp
jessemapel: This was resolved but I forgot to take out the TODO

isis/src/base/objs/CSMCamera/CSMCamera.cpp
jessemapel: Unfortunately, time is not unique within an image. For example, what would you do for a framing image? Every line and sample would have the same time.

isis/tests/TestCsmModel.cpp
jessemapel: This gets used in the csminit tests. They use TestCsmPlugin which creates TestCsmModel and AlternativeTestCsmModel

isis/src/base/objs/Spice/Spice.h
jessemapel: Nope it should go

isis/src/base/apps/csminit/csminit.cpp
jessemapel: The else on this if / else block handles the neither case. I'll move the else up to here and do an explicit check for neither so that the two errors are next to each other.

isis/src/base/objs/StringBlob/StringBlob.cpp
jessemapel: No reason to keep it around, it should go away

isis/tests/iTimeTests.cpp
jessemapel: Mathematically, a < b does imply  b !< a, but the properties of comparisons are not enforced by C++. It's up to the implementation to make sure that happens.

jessemapel:
My stance on API history comments is:
* If there are already history comments either remove them or add to them
* If there aren't already history comments, don't create them
@acpaquette Does that sound reasonable?

AustinSanders (dismissed):
Overall I think this was a super clean PR.  I had to get a little nitpick-y with some of the comments just so that I could find something to write.

All of my concerns have been addressed either through code changes or responses to comments.

isis/src/base/objs/CameraPointInfo/CameraPointInfo.cpp
acpaquette: Talked about this over teams. This is fine for the time being and we often don't deal with null keyword values in ISIS rather, we check if the keyword is present. Seeing as we are trying to fit CSM into ISIS, this current system of setting these keywords to null manually is okay

acpaquette (dismissed):

AustinSanders (approved):

acpaquette (approved):

PR #4301, opened 2021-02-20
History Blob Update

acpaquette:
Updates the history class to no longer be a subclass of blob.

## Description
Changed the history classes dependency on the blob class to a "uses" relationship rather than an "is a" relationship. So history objects can be built from blobs or constructed on their own. Reading the history blob is now done by the cube which feeds the blob to the history object which generates the PVL, then the history object can produce a blob from itself.

## Related Issue
N/A

## Motivation and Context
This is to facility the separation of responsibilities between IO elements of ISIS. Removing the IO elements from History allowing blob to take care of the bulk of it.

## How Has This Been Tested?
This has been tested with the current ISIS tests and will receive it's own tests

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
+++ b/isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
@@ -170,13 +170,13 @@ namespace Isis {
 
         if (cube->label()->hasObject("History")) {
             // Record apollofindrx history to the cube
-            PvlObject histObj = cube->label()->findObject("History");
+            QString histName = (QString)cube->label()->findObject("History")["Name"];
             // create a History Blob with value found in the History PvlObject's Name keyword
             // read cube's History PvlObject data into the History Blob
-            Isis::History *hist = cube->readHistory();
+            Isis::History *hist = cube->readHistory(histName);
             // add apollofindrx History PvlObject into the History Blob and write to cube
             hist->AddEntry();
-            cube->write(*(hist->toBlob((QString)histObj["Name"])));
+            cube->write(*(hist->toBlob(histName)));
             cube->close();
         }
     }
--- a/isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
+++ b/isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
@@ -169,15 +169,14 @@ namespace Isis {
         }
 
         if (cube->label()->hasObject("History")) {
-            PvlObject histObj = cube->label()->findObject("History");
             // Record apollofindrx history to the cube
+            PvlObject histObj = cube->label()->findObject("History");
             // create a History Blob with value found in the History PvlObject's Name keyword
-            Isis::History histBlob( (QString)histObj["Name"] );
             // read cube's History PvlObject data into the History Blob
-            cube->read(histBlob);
+            Isis::History *hist = cube->readHistory();
             // add apollofindrx History PvlObject into the History Blob and write to cube
-            histBlob.AddEntry();
-            cube->write(histBlob);
+            hist->AddEntry();
+            cube->write(*(hist->toBlob((QString)histObj["Name"])));
             cube->close();
         }
     }
--- a/isis/src/base/apps/cathist/main.cpp
+++ b/isis/src/base/apps/cathist/main.cpp
@@ -5,7 +5,7 @@
 #include "FileName.h"
 #include "IString.h"
 #include "History.h"
-#include "Pvl.h"   
+#include "Pvl.h"
 #include "TextFile.h"
 
 using namespace Isis;
@@ -26,7 +26,8 @@ void IsisMain() {
   }
 
   // Extract history from file
-  History hist("IsisCube", fromfile.expanded());
+  Blob historyBlob("IsisCube", "History", fromfile.expanded());
+  History hist(historyBlob);
   Pvl pvl = hist.ReturnHist();
 
   // Print full history
--- a/isis/src/base/objs/Blob/Blob.cpp
+++ b/isis/src/base/objs/Blob/Blob.cpp
@@ -46,7 +46,7 @@ namespace Isis {
    * @param type The blob type
    * @param file The filename to read from.
    */
-  Blob::Blob(const QString &name, const QString &type, 
+  Blob::Blob(const QString &name, const QString &type,
              const QString &file) {
     p_blobName = name;
     p_buffer = NULL;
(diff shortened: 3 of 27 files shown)


isis/src/base/objs/History/unitTest.cpp
```diff
@@ -17,14 +17,14 @@ using namespace std;
...
+  // Isis::History h("Haha");;
+  // h.AddEntry();
+  // QString file = "unitTest.tttt";
+  // h.Write(file);
+  //
+  // Isis::History h2("Haha", file);
+  // Isis::PvlObject o = h2.ReturnHist();
+  // std::cout << o << std::endl;
+  //
+  // remove(file.toLatin1().data());
```
acpaquette, line 29: I'm working on writing new tests for the history object along with the functions I've added.

isis/src/base/objs/History/History.h
```diff
@@ -29,27 +29,24 @@ namespace Isis {
...
       ~History();
 
       void AddEntry();
       void AddEntry(Isis::PvlObject &obj);
       Pvl ReturnHist();
-      void Read(const Isis::Pvl &pvl, std::istream &is);
 
-    protected:
-      void WriteInit();
+      Blob *toBlob(const QString &name = "IsisCube");
```
acpaquette, line 45: We may want to change this to just return a char *buffer and let the blob class construct itself from it

isis/src/base/objs/Blob/Blob.cpp
```diff
@@ -368,6 +368,19 @@ namespace Isis {
...
+
+    if (p_buffer != NULL) {
+      delete [] p_buffer;
+    }
+    p_buffer = new char[p_nbytes];
+
+    for (int i = 0; i < p_nbytes; i++) {
+      p_buffer[i] = buffer[i];
+    }
+  }
```
jessemapel, line 382: It's possible this function should just pass ownership of the buffer to the Blob to avoid copying potentially quite large chunks of data.

isis/src/base/objs/History/History.cpp
```diff
@@ -56,24 +59,19 @@ namespace Isis {
...
-    if (p_nbytes > 0) ostr << std::endl;
     ostr << p_history;
     string histStr = ostr.str();
-    int bytes = histStr.size();
+    int nbytes = histStr.size();
 
-    char *temp = p_buffer;
-    p_buffer = new char[p_nbytes+bytes];
-    if (temp != NULL) memcpy(p_buffer, temp, p_nbytes);
+    char *buffer = new char[nbytes];
```
jessemapel, line 68: This buffer is a memory leak right now. If we instead pass ownership in Blob::setData this problem goes away.

isis/src/base/objs/History/History.cpp
```diff
@@ -56,24 +59,19 @@ namespace Isis {
...
     const char *ptr = histStr.c_str();
-    memcpy(&p_buffer[p_nbytes], (void *)ptr, bytes);
-    p_nbytes += bytes;
+    memcpy(&buffer[0], (void *)ptr, nbytes);
```
jessemapel, line 70: Is there a reason to define ptr here? Can this line just be:

```
memcpy(buffer, histStr.c_str(), nbytes);
```

jessemapel: Also, I'm fairly certain that `&buffer[0]` is just the same as `buffer` because it's already a pointer.

isis/src/base/apps/cathist/main.cpp
```diff
@@ -26,7 +26,8 @@ void IsisMain() {
   }
 
   // Extract history from file
-  History hist("IsisCube", fromfile.expanded());
+  Blob historyBlob("IsisCube", "History", fromfile.expanded());
```
jessemapel, line 29: I am not a fan of having "IsisCube" and "History" hard coded in places outside of the History class. Not sure the best way around this, maybe some way to get an empty Blob for reading/writing History objects?

isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
```diff
@@ -169,15 +169,14 @@ namespace Isis {
...
 
         if (cube->label()->hasObject("History")) {
-            PvlObject histObj = cube->label()->findObject("History");
             // Record apollofindrx history to the cube
+            QString histName = (QString)cube->label()->findObject("History")["Name"];
             // create a History Blob with value found in the History PvlObject's Name keyword
-            Isis::History histBlob( (QString)histObj["Name"] );
             // read cube's History PvlObject data into the History Blob
-            cube->read(histBlob);
+            Isis::History *hist = cube->readHistory(histName);
```
jessemapel, line 176: I find it really weird that apollofindrx has different names for the history BLOB compared to everything else. Maybe we could take this out and use the normal logic?

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -829,6 +830,18 @@ namespace Isis {
...
+  History *Cube::readHistory(const QString &name) const {
+    Blob historyBlob(name, "History");
+    try {
+      // read history from cube, if it exists.
+      historyBlob.Read(fileName());
+    }
+    catch (IException &) {
+    // if the history does not exist in the cube, this function creates it.
+    }
+    History *history = new History(historyBlob);
```
jessemapel, line 842: Is there a reason why this is returning a pointer? I'm concerned about memory leaks here.

jessemapel (commented):
I think this is good enough to merge into the feature branch already so that we can all use the Blob changes.

isis/src/base/objs/Cube/Cube.cpp
acpaquette: I will update this to not return a pointer

isis/src/apollo/apps/apollofindrx/apollofindrx.cpp
acpaquette: Maybe, I think leaving things as they are in the apps isn't the worst thing but I agree. Seems like there was some idea that history blobs could be stored in different locations on the label

jessemapel (approved):

isis/src/base/apps/cathist/main.cpp
acpaquette: Maybe? The only reason these are hard coded here is because we don't have a cube open. The "IsisCube" is also hard coded in the `readHistory` function on the cube and the `toBlob` method on the history object. They are both set in the associated `.h` files

isis/src/base/objs/History/History.cpp
acpaquette: Again just a copy of previous work. I will update this with the suggested changes

Issue #4303, opened 2021-02-22
ISIS Calibration Programs could be updated to run without local spice


krlberry:
**Description**  
Update ISIS calibration programs to work without local spice, so it is possible to run calibration programs without downloading mission-specific kernels. In this scenario, it would be possible to run spiceinit using the Spice Server and then be able to run calibration programs, all without downloading kernels. At this time, this is not possible for many different calibration programs. 

Many calibration programs directly furnish kernels at this time, by calling furnsh_c in the code, and this obviously does not work without local kernels.  

@Kelvinrr  already converted a calibration program, lronaccal, in this way:  
https://github.com/USGS-Astrogeology/ISIS3/commit/8cda49e652045ba497c7732ee35585250a2e8605 

* Updated lronaccal to instantiate a camera to get the moon-sun distance from the camera, rather than furnishing kernels and using a naif routine directly do to this calculation. 

* Fall back to furnishing kernels if this fails. 

This feature has been requested in the following tickets: 
#1790
#1473
and by the [MAPSIT SAT](https://github.com/USGS-Astrogeology/softwaremanagement/blob/master/SpecialActionTeam_Review/2018_SAT_Report.md)

This ticket covers updating: 
* ctxcal
* gllssical
* vikcal 
* lrowaccal 
* amicacal
* mdiscal 
* hical
* hicalbeta
* moccal 

And also the following apps were updated to remove furnsh calls: 
* hijitreg
* shadow

**Example**

PR #4115, opened 2020-11-09
Added error checking around spicelib furnsh call. Closes #4038

scsides:
Added error checking around calls to spicelib furnsh_c, so missing files would be reported instead of continuing without an error.

## Motivation and Context

The program was continuing on and creating invalid output when it was unable to find SPICE kernel files required for proper calibration.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I have read and agree to abide by the [Code of Conduct](https://usgs-astrogeology.github.io/code/)
- [X] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.
- [X] I have added any user impacting changes to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [X] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

Issue #4038, opened 2020-09-28
lrowaccal is nonfunctional on MacOS (Edit: On any system without kernels directory downloaded.)


rvwagner:
**ISIS version(s) affected**: 3.10.1 and 4.1.0 on MacOS 10.14.6, and 3.12.0 on a different MacOS 10.14.6 system.
EDIT: Just updated the 4.1.0 install to 4.2.0.  The issue persists.
EDIT: The "3.12.0" was probably 4.something, with the version edited to say "3.12.0" as a work-around for a web SPICE error, per https://github.com/USGS-Astrogeology/ISIS3/issues/3725#issuecomment-619091103. Ben Bussey may have more detail- it's his system.

**Description**  
When running a LROC WAC image through lrowaccal on MacOS, the. output values all end up as 0.0.  The exact same processing steps on ISIS 3.9.1 and 4.1.0 on Linux produce a properly-calibrated image.  All the calibration files listed in the Radiometry groups are identical (same MD5 checksum) between the two systems, so I suspect it's the executable file that's the problem.

**How to reproduce**  
    wget http://pds.lroc.asu.edu/data/LRO-L-LROC-2-EDR-V1.0/LROLRC_0027/DATA/ESM2/2016163/WAC/M1220245217CE.IMG
    lrowac2isis from=M1220245217CE.IMG to=M1220245217CE
    lrowaccal from=M1220245217CE.vis.even.cub to=M1220245217CE.vis.even.cal.cub

UPDATE:
Skipping the radiometric calibration step causes the output to stop being 0, so that's probably where the problem is:
lrowaccal from=M1220245217CE.vis.even.cub+1 to=M1220245217CE.vis.even.cal.cub Radiometric=false

rvwagner:
Update: It's not MacOS, it's not having the $Lro/kernels directory.  Both of the above computers were relying on web SPICE, while the Linux system had local kernels.  When I set up an Lro data directory without a kernels directory on that Linux system, I got the same 0 result.  EDIT: If I add the contents of kernels/tspk/ and the de421 files from kernels/pck/, the radiometry correction works correctly.

Not sure how this should be solved long-term, but it means there's a workaround of downloading ~20MB of kernels.

jessemapel:
[...]

PR #4304, opened 2021-02-22
ctxcal now uses cam

Kelvinrr:
## Description

Part of making calibration apps work without local spice, ctxcal attempts to use the camera before using kernels. 

## Related Issue

#4303

## How Has This Been Tested?

Tested locally, both camera and local spice return the same result. Should we consider adding tests as original tests do not use spice inited cubes? 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/mro/apps/ctxcal/main.cpp
+++ b/isis/src/mro/apps/ctxcal/main.cpp
@@ -15,6 +15,7 @@ find files of those names at the top level of this repository. **/
 #include "LineManager.h"
 #include "Brick.h"
 #include "Table.h"
+#include "Camera.h"
 
 using namespace std;
 using namespace Isis;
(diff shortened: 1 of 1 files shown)


isis/src/mro/apps/ctxcal/main.cpp
```diff
@@ -139,21 +140,30 @@ void IsisMain() {
...
+      QString satKernel = p.MissionData("base", "/kernels/spk/mar???.bsp", true);
+      furnsh_c(satKernel.toLatin1().data());
+      QString pckKernel = p.MissionData("base", "/kernels/pck/pck?????.tpc", true);
+      furnsh_c(pckKernel.toLatin1().data());
+      double sunpos[6], lt;
+      spkezr_c("sun", etStart, "iau_mars", "LT+S", "mars", sunpos, &lt);
+      dist1 = vnorm_c(sunpos);
+      unload_c(bspKernel.toLatin1().data());
+      unload_c(satKernel.toLatin1().data());
+      unload_c(pckKernel.toLatin1().data());
```
jessemapel, line 174: It would be good to add a NaifStatus::CheckErrors() call here to check for failures in any of these SPICE calls. See #4038  #4115 for why this matters.

krlberry:
This looks good to me! I think your idea to add a test (or tests) that use spiceinited cubes, so that this updated code is tested, is a good idea.

Kelvinrr:
@kberryUSGS So the question is: Should I just update the tests to use spiceinit-ed cubes (easy, done today), or, should I make a GTest for the case instead (requires converting the app, would add a lot of time to development). I am kinda fine with either, it's more if you think we can still get though all of them if we also covert the apps as I imagine we will have to test all of these.

krlberry:
@Kelvinrr I'm honestly fine with either, too. It's tricky because that's technically beyond the scope of this project, but also it feels like adding technical debt to add new tests in the old testing framework. 

Could you go ahead and convert just `ctxcal` and its tests and see how long it takes? Then, we can use that as a rough estimate to see if it's even feasible to convert the rest of the calibration apps while we're in here or not. 

My preference is to have the conversion come in as a separate PR anyway, since it's hard to see the diffs if the code changes and the conversion are in the same PR, so I'd lean toward merging this PR if CI passes (and you're confident in the local tests you ran,) and then you could put up a new PR for just the conversion and new test(s) when that's ready.

krlberry (dismissed):

isis/src/mro/apps/ctxcal/main.cpp
scsides: Only one NaifStatus check is necessary. Just surround all the spicelib call with one. It will throw on the first one.

scsides (changes requested):

isis/src/mro/apps/ctxcal/main.cpp
scsides: Per our conversation, this looks good.

scsides (dismissed):

krlberry (approved):
:+1:

PR #4309, opened 2021-02-23
Refactor OriginalLabel class to remove inheritance from Blob

AustinSanders:
Refactor OriginalLabel class to remove inheritance from Blob

## Description

Refactor OriginalLabel to remove inheritance from Blob.  

## Motivation and Context
More clearly separates functionality so that we can move toward building pvl / blob / functionality independently.

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -843,6 +844,21 @@ namespace Isis {
...
+   * This method will read an OriginalLabel from a cube.
+   */
+  OriginalLabel Cube::readOriginalLabel() const {
+    Blob origLabelBlob("IsisCube", "OriginalLabel");
+    try {
+      origLabelBlob.Read(fileName());
+    }
+    catch(IException &) {
+     //Intentionally left blank
+    }
```
jessemapel, line 874: Unlike History, we want to actually throw an error here if we can't read the original label from the file.

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
```diff
@@ -16,19 +16,40 @@ namespace Isis {
...
    */
-  OriginalLabel::OriginalLabel(const QString &file) :
-    Isis::Blob("IsisCube", "OriginalLabel") {
-    Blob::Read(file);
+  OriginalLabel::OriginalLabel(const QString &file){
+    Pvl pvl;
+    stringstream os;
+    Blob blob = Blob("IsisCube", "OriginalLabel");
+    blob.Read(file);
+    char *buff = blob.getBuffer();
```
jessemapel: From here down is the same logic as the Blob constructor. We should move this logic into a shared function, it could be public or private.

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
```diff
@@ -37,35 +58,46 @@ namespace Isis {
...
-    int bytes = orglblStr.size();
 
-    // Copy label data to bytes variable
-    char *temp = p_buffer;
-    p_buffer = new char[p_nbytes+bytes];
-    if(temp != NULL) memcpy(p_buffer, temp, p_nbytes);
-    const char *ptr = orglblStr.c_str();
-    memcpy(&p_buffer[p_nbytes], (void *)ptr, bytes);
-    p_nbytes += bytes;
+  Isis::Blob *OriginalLabel::toBlob() {
```
jessemapel, line 68: Why return a pointer here?

isis/src/galileo/apps/gllnims2isis/main.cpp
```diff
@@ -146,6 +146,7 @@ void importQubs(QString coreParamName, QString suffixParamName) {
   }
 
   // Create holder for original label
+  // TODO are we using origLabel anywhere?  It's not being written
   OriginalLabel origLabel(*pdsLabel);
```
jessemapel: It looks like this can just be removed, it never gets passed to anything or used anywhere

jessemapel (commented):
Looks good. Just two small comments really.

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
jessemapel: Nevermind, this is consistent with History

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
acpaquette: Yes but this may not always be the case and I don't know where a function like that would live. If anything this may be reason to have history and original label exist as the same object

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
```diff
@@ -76,7 +108,7 @@ namespace Isis {
   Pvl OriginalLabel::ReturnLabels() {
     Pvl pvl;
     stringstream os;
-    for(int i = 0; i < p_nbytes; i++) os << p_buffer[i];
+    os << m_originalLabel;
     os >> pvl;
     return pvl;
```
acpaquette: This should just need to return m_originalLabel

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -913,6 +929,15 @@ namespace Isis {
...
 
+  /**
+   * This method will write an OriginalLabel object.
+   * to the cube as specified by the contents of the Blob object.
+   *
+   * @param Original label data to be written
+   */
+  void Cube::write(OriginalLabel lab) {
+    write(*(lab.toBlob()));
+  }
```
acpaquette, line 957: I know this write method makes things easier but I don't know how necessary it is and that's something I have been going back and forth on. It'd be good for us to figure out exactly how we want to be writing these blobs back out. While the `*(lab.toBlob())` is a little messy I don't think it's bad to have `write(*(lab.toBlob()))` all over the place rather than `write(lab)`

acpaquette (commented):
A few things to discuss but overall looks good!

isis/src/base/objs/OriginalLabel/OriginalLabel.cpp
AustinSanders: Yeahhhhhh this was the result of trying to fix a bug by changing the wrong part of the code, then I got it working and didn't change everything back.

Good catch

jessemapel:
For reading/writing these things, I think the purest thing to do is enforce File <-> Blob <-> Class. It requires a bit more code but it's straight forward.

Then, I start writing code and I just wish I had a direct read/write method that wrapped the File <-> Blob part.

I could see either way, but we should be consistent.

isis/src/base/objs/Cube/Cube.cpp
AustinSanders: Is there a specific error that should be thrown, or should I just remove the try/catch so that the original error is raised?

jessemapel: I would append something like "Could not read the original label from [cube.cub]." and then rethrow.

jessemapel (approved):

acpaquette (approved):

PR #4312, opened 2021-02-23
Modified Table to not inherit from Blob anymore

jessemapel:
I am still running tests on this to ensure that nothing random broke but I want to get it up here for input.


isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -914,6 +916,12 @@ namespace Isis {
   }
 
 
+  void Cube::write(const Table &table) {
+    Blob tableBlob = table.toBlob();
+    write(tableBlob);
+  }
```
acpaquette, line 964: Similarly with Austin's PR I don't know if we need a write function for the table object specifically. Just adding `write(table.toBlob())` seems okay to me

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -1838,6 +1846,13 @@ namespace Isis {
   }
 
 
+  Table Cube::readTable(const QString &name) {
+    Blob tableBlob(name, "Table");
+    read(tableBlob);
```
acpaquette: Should there be a try catch around this read?

acpaquette (commented):
Overall looks good! :+1: Few things to talk about

isis/src/base/objs/Cube/Cube.cpp
jessemapel: Yeah I should add a try catch to rethrow the Blob error with a better message.

isis/src/base/objs/Cube/Cube.cpp
acpaquette: We (Jesse, Austin, and I) will be discussing this offline and come to a conclusion on how the writes should be handled

acpaquette (approved):

PR #4321, opened 2021-02-25
Removed StringBlob Class

jessemapel:
Moved to no longer be a subclass of Blob. I made some methods virtual so that we can inherit but I'm not sure the best way to handle that right now.


isis/src/base/objs/StringBlob/StringBlob.cpp
```diff
@@ -7,82 +7,55 @@ find files of those names at the top level of this repository. **/
...
 
 using namespace std;
 namespace Isis {
   /**
    * Constructor for creating a string blob with no arguments
    */
-  StringBlob::StringBlob() : Isis::Blob("IsisCube", "String") {
+  StringBlob::StringBlob() {
     m_string = "";
+    m_name = "";
```
acpaquette: I don't know if we want to carry around the name of the PVL container name for the Blob PVL. Similarly to the history object, this should probably only live in the IsisCube

isis/src/base/objs/StringBlob/StringBlob.h
```diff
@@ -25,24 +28,31 @@ namespace Isis {
...
+      std::string string() const {
         return m_string;
       }
 
-    protected:
-      // prepare data for writing
-      void WriteInit();
-      void ReadData(std::istream &stream);
+      QString name() const {
+        return m_name;
```
acpaquette: These are generally all over the place in ISIS but I would rather these be getString and getName

acpaquette (commented):
Overall looks good! A few things to address

isis/tests/FunctionalTestsSpiceinit.cpp
```diff
@@ -528,21 +528,21 @@ TEST(Spiceinit, TestSpiceinitPadding) {
 TEST_F(DefaultCube, TestSpiceinitCsmCleanup) {
   // Add stuff from csminit
   testCube->putGroup(PvlGroup("CsmInfo"));
-  StringBlob testBlob("test string", "CSMState");
+  Blob testBlob("CSMState", "String");
```
acpaquette, line 531: Something I mentioned to Austin was that the blob type should probably be an enum to properly define a selection of blob types. Right now it just exists as a string so the type could be anything.

isis/src/base/apps/csminit/csminit.cpp
```diff
@@ -411,7 +410,10 @@ namespace Isis {
     cube->deleteBlob("Polygon", "Footprint");
 
     // Create our CSM State blob as a string and add the CSM string to the Blob.
-    StringBlob csmStateBlob(modelState, "CSMState");
+    Blob csmStateBlob("CSMState", "String");
+    char *buf = new char[modelState.size()];
+    memcpy(buf, modelState.c_str(), modelState.size());
+    csmStateBlob.setData(buf, modelState.size()); // This passes ownership so don't delete buf
```
acpaquette: We could do this in the set data function, but then we would need to delete the locally scoped char blob. Maybe we could also do a blob constructor from a string?

acpaquette (commented):
:+1: Dig the overall reduction in code! Always good to see

isis/src/base/apps/csminit/csminit.cpp
jessemapel: Yeah, I had a hard time figuring out a good way to do this. We could have Blob::setData do this and change it to take a const char* I'll look and see if that works better.

isis/tests/FunctionalTestsSpiceinit.cpp
jessemapel: We could do something like:

* Binary <- Blob default
* String
* Table
* OriginalLabel
* OriginalXmlLabel
* History
* Stretch
* Polygon

acpaquette: I think this looks good! I would rather see Geometry over Polygon but either works

isis/src/base/apps/csminit/csminit.cpp
AustinSanders: This is something that I started to comment on in a review as well, but I wasn't really sure where a function like this would live.  I think changing setData makes the most sense

isis/tests/FunctionalTestsSpiceinit.cpp
jessemapel: These are the currently used types except for Binary. We need to maintain backwards compatibility.

jessemapel: This is going to potentially cause some weird merge conflicts with other stuff, so I'm going to put this is in as a separate PR after this.

isis/src/base/apps/csminit/csminit.cpp
jessemapel: I ended up having 2 ways to set that data. You can call setData to copy the data and takeData to pass the data. This is because of how Table stores it data. With the copy inside of setData you end up with 3 copies of the Table data at one point.

acpaquette (approved):

PR #4323, opened 2021-02-26
Removed blob inheritance from stretchblob

AustinSanders:
Removed blob inheritance from stretchblob.  One test failure.  I think it's something to do with the toBlob function.

## Description
Removed blob inheritance from stretchblob
## Related Issue
None

## Motivation and Context
Factors out blob inheritance.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/StretchBlob/StretchBlob.cpp
+++ b/isis/src/base/objs/StretchBlob/StretchBlob.cpp
@@ -7,6 +7,7 @@ find files of those names at the top level of this repository. **/
 
 #include <iostream>
 #include <fstream>
+#include <sstream>
 
 #include "StretchBlob.h"
 #include "IException.h"
@@ -17,17 +18,16 @@ namespace Isis {
   /**
    * Default constructor
    */
-  StretchBlob::StretchBlob() : Blob("CubeStretch", "Stretch"), m_stretch() {
+  StretchBlob::StretchBlob() {
+    m_stretch = CubeStretch("CubeStretch", "Stretch");
   }
 
 
   /**
    * Construct a StretchBlob from a CubeStretch.
    */
-  StretchBlob::StretchBlob(CubeStretch stretch) : Blob("CubeStretch", "Stretch"), m_stretch(stretch){
-    p_blobPvl["Name"] = m_stretch.getName();
-    p_blobPvl += PvlKeyword("StretchType", m_stretch.getType());
-    p_blobPvl += PvlKeyword("BandNumber", QString::number(m_stretch.getBandNumber()));
+  StretchBlob::StretchBlob(CubeStretch stretch) {
+    m_stretch = CubeStretch(stretch);
   }
 
 
@@ -36,7 +36,16 @@ namespace Isis {
    *
    * @param name Name to use for Stretch
    */
-  StretchBlob::StretchBlob(QString name) : Blob(name, "Stretch"), m_stretch(name) {
+  StretchBlob::StretchBlob(QString name) {
+    m_stretch = CubeStretch(name, "Stretch");
+  }
+
+  StretchBlob::StretchBlob(Blob blob) {
+    stringstream os;
+    char *buff = blob.getBuffer();
+    std::string stringFromBuffer(buff);
+    m_stretch = CubeStretch(blob.Name(), blob.Type());
+    m_stretch.Parse(QString::fromStdString(stringFromBuffer));
   }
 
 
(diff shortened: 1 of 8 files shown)


isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -17,17 +18,16 @@ namespace Isis {
   /**
    * Default constructor
    */
-  StretchBlob::StretchBlob() : Blob("CubeStretch", "Stretch"), m_stretch() {
+  StretchBlob::StretchBlob() {
+    m_stretch = CubeStretch("CubeStretch", "Stretch");
```
krlberry: You could just leave this without arguments if you like. CubeStretch's constructor has default parameters for every value.

krlberry: If nothing else, the arguments should probably match the existing default values in ```CubeStretch(QString name="DefaultStretch", QString stretchType="Default", int bandNumber = 1);```

isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -36,7 +36,16 @@ namespace Isis {
    *
    * @param name Name to use for Stretch
    */
-  StretchBlob::StretchBlob(QString name) : Blob(name, "Stretch"), m_stretch(name) {
+  StretchBlob::StretchBlob(QString name) {
+    m_stretch = CubeStretch(name, "Stretch");
```
krlberry: Similar to comment above, I'd say either just use `CubeStretch(name)` or at least use the same default argument of "Default" since the second argument to the `CubeStretch` constructor is the type of the stretch, and that's used in the GUI StretchTool to determine what to display.

isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -36,7 +36,16 @@ namespace Isis {
...
+  }
+
+  StretchBlob::StretchBlob(Blob blob) {
+    stringstream os;
+    char *buff = blob.getBuffer();
+    std::string stringFromBuffer(buff);
+    m_stretch = CubeStretch(blob.Name(), blob.Type());
```
krlberry: I think this might be the wrong name and type for the CubeStretch? The blob's name is "CubeStretch" and its type is "Stretch" but the CubeStretch's name and type should be different. In the old version of the code, that information is stored in the Blob PVL as the name, stretch type, and band number. I'm confused about where this information is being stored in the updated code, but I'm not that familiar with the details of this project.

isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -51,67 +60,52 @@ namespace Isis {
...
-   * Read saved Stretch data from a Cube into this object. 
-   *  
-   * This is called by Blob::Read() and is the actual data reading function 
-   * ultimately called when running something like cube->read(stretch);  
-   * 
-   * @param is input stream containing the saved Stretch information
-   */
-  void StretchBlob::ReadData(std::istream &is) {
-    // Set the Stretch Type
-     m_stretch.setType(p_blobPvl["StretchType"][0]);
```
krlberry, line 64: Where does this happen in updated code?

isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -51,67 +60,52 @@ namespace Isis {
...
-     m_stretch.setBandNumber(p_blobPvl["BandNumber"][0].toInt());
```
krlberry, line 65: Where does this happen in updated code?

isis/src/base/objs/StretchBlob/StretchBlob.cpp
```diff
@@ -51,67 +60,52 @@ namespace Isis {
...
-   }
 
-
-  /**
-   *  Initializes for writing stretch to cube blob
-   */
-  void StretchBlob::WriteInit() {
-    p_nbytes = m_stretch.Text().toStdString().size(); 
+  Isis::Blob *StretchBlob::toBlob() {
+    Isis::Blob *blob = new Blob(m_stretch.getName(), m_stretch.getType());
```
krlberry: The Stretch Blobs are named "CubeStretch" and have type "Stretch", which is different from the name of the specific CubeStretch name, specified by the user and could be anything, like "MyBestStretch2021!" and the type is the specific type of stretch specified in the AdvancedStretchTool (i.e. Linear, Sawtooth, etc)

krlberry (changes requested):
I left some comments on here. I hope they're helpful and I'm still very happy to talk about any of this on Teams. The major issue I see is that the StretchBlob should have* name=CubeStretch and type=Stretch, but the CubeStretch should have name="whatever the user specified in the StretchTool, which could be anything" and type="a type of stretch, like Linear" and it seems like those got collapsed into a single name and type for both in these updates. 

\* = This is the way it used to work -- maybe this update is part of the project?

isis/src/base/objs/StretchBlob/StretchBlob.cpp
krlberry: So, the stretch name should be different than the blob name, for example.

isis/src/base/objs/StretchBlob/StretchBlob.cpp
AustinSanders: I was thinking that this happened during the parsing step, i.e.:

```m_stretch.Parse(QString::fromStdString(stringFromBuffer));```

edit:
I guess it's only copying the stretch pairs?  That's not what I expected.

AustinSanders:
This branch refuses to accept changes.  I'm opening a new PR with more fixes.

PR #4349, opened 2021-03-09
Fixed pvl core test config

jessemapel:
This sets up the PVL core tests so that they both properly build and link gtest if it's not already available and they can be enabled/disabled.

### DIFF
--- a/isis/src/core/CMakeLists.txt
+++ b/isis/src/core/CMakeLists.txt
@@ -4,7 +4,7 @@ project (core VERSION 0.0.1)
 option(buildStaticCore "Build libisis static as well as dynamic"        OFF )
 option (BUILD_CORE_TESTS "Build core module tests" ON)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
 # Short and long name of this package
 set(PACKAGE            "CORE")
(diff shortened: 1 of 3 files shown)


isis/src/core/CMakeLists.txt
```diff
@@ -13,8 +14,45 @@ set(PACKAGE_NAME       "USGS ISIS PVL CORE")
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
-FILE(GLOB_RECURSE Pvl_source_files ${PROJECT_SOURCE_DIR}/src *.cpp)
```
jessemapel, line 16: The CMake docs explicitly discourage using GLOB to get source files because it makes the rebuild stuff not work well.

isis/src/core/CMakeLists.txt
```diff
@@ -2,8 +2,9 @@ cmake_minimum_required(VERSION 3.10)
 
 project (core VERSION 0.0.1)
 option(buildStaticCore "Build libisis static as well as dynamic"        OFF )
+option (BUILD_CORE_TESTS "Build core module tests" ON)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```
jessemapel, line 7: PROJECT_SOURCE_DIR should be used in sub-modules because it is relative to the last project call, the PVL Core in this case, instead of whichever CMakeLists.txt file was called first.

AustinSanders (approved):

acpaquette (approved):
:+1: Looks solid and fixed my build problems!

PR #4351, opened 2021-03-10
Moved old ISIS unitTests into core

jessemapel:
Moved all of the old style unit tests into the core and got them passing except for IException because we have commented out the program name part of the print out. I also moved a bunch of things out of the core because they are not require for the PVL classes.

This is a huge PR, but just because it's adding back a bunch of files that were removed. Look specifically at the CMakeLists.txt files and the new RunUnitTest.cmake file (this is copied from isis/cmake).


AustinSanders (approved):

isis/src/core/tests/unitTests/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+  set(executableName "core_unit_test_${class}")
+
+  # Create the test executable
+  add_executable(${executableName} ${testFile})
+  target_link_libraries(${executableName} core)
+
+  # Set up a cmake script which will run the executable
+  # and then check the results against the truth file.
+  set(testScript "${PROJECT_SOURCE_DIR}/cmake/RunUnitTest.cmake")
+  set(fullTestPath $<TARGET_FILE:${executableName}>)
```
jessemapel, line 27: The location of the test executable changes based on if just core is being built or all of ISIS. This resolves to the proper location regardless of where it ends up.

jessemapel:
@AustinSanders CI looks good I'm going to go ahead and merge this

PR #4354, opened 2021-03-12
Minor fix for pvl core preferences

jessemapel:
The preference files didn't get updated properly and are causing test failures. This should fix that.


acpaquette (approved):
:+1:

Issue #4006, opened 2020-09-08
hideal2pds outputs a label where SAMPLE_BITS value is always 16


tgiroux:
**ISIS version(s) affected**: 4.2.0

**Description**  

When a user selects bits=8, the SAMPLE_BITS value in the output label should be 8. For all other cases (bits=9-16), SAMPLE_BITS should be 16. At present it is 16 for all cases including bits=8.

**How to reproduce**  

In isis/src/mro/apps/hideal2pds/tsts/differentBitCompressions run `$make output` and check output/output8.pvl to see that SAMPLE_BITS=16

**Possible Solution**  

Include the following code within IsisMain() of main.cpp for hideal2pds:
`if( nbits == 8 ) {`
`  image.addKeyword(PvlKeyword("SAMPLE_BITS", toString(nbits)), Pvl::Replace);`
`}`

**Additional context**  

Related to #3978 by @audiefen

Issue #3978, opened 2020-08-20
hideal2pds outputs incorrect value of SAMPLE_BIT_MASK in label for all but 16 bit compression


audiefen:
**ISIS version(s) affected**: 3.9.0

**Description**  
With hideal2pds, when selecting any bit compression other than 16, the SAMPLE_BIT_MASK in the output label is incorrect. Below is an excerpt from an output label where bit=10 was selected.
[...
OBJECT = IMAGE
  DESCRIPTION                = "HiRISE mosaicked product, not map projected"
  LINES                      = 2500
  LINE_SAMPLES               = 5000
  SOURCE_LINE_SAMPLES        = 5000.0
  SOURCE_LINES               = 2500.0
  FIRST_LINE_SAMPLE          = 0.5
  FIRST_LINE                 = 0.5
  BANDS                      = 1
  OFFSET                     = 0.0037222964252952
  SCALING_FACTOR             = 3.14850405127014e-06
  SAMPLE_BITS                = 16
  SAMPLE_BIT_MASK            = 2#1111111111111111#
  SAMPLE_TYPE                = MSB_UNSIGNED_INTEGER
  FILTER_NAME                = "Red"
  CENTER_FILTER_WAVELENGTH   = 700 <NM>
  BAND_WIDTH                 = 300 <NM>
  BAND_STORAGE_TYPE          = BAND_SEQUENTIAL
  CORE_NULL                  = 0
  CORE_LOW_REPR_SATURATION   = 1
  CORE_LOW_INSTR_SATURATION  = 2
  CORE_HIGH_REPR_SATURATION  = 1023
  CORE_HIGH_INSTR_SATURATION = 1022
END_OBJECT = IMAGE
...]

Should be:
`SAMPLE_BIT_MASK 2#0000001111111111#`

**How to reproduce**  
hideal2pds max=0.006936919061642 to=ESP_057858_1410_RED.NOPROJ.img type=MANUAL version=1.0 min=0.003731741937449 from=PSP_009492_1280_RED.NOPROJ.cub bits=10

**Possible Solution**  

**Additional context**  
[PSP_009492_1280_RED.NOPROJ.lbl.gz](https://github.com/USGS-Astrogeology/ISIS3/files/5104979/PSP_009492_1280_RED.NOPROJ.lbl.gz)
[...]

PR #4404, opened 2021-04-15
hideal2pds SAMPLE_BITS = 8 for ubyte output

tgiroux:
## Description

Change SAMPLE_BITS to 8 in the PDS image label if the user entered BITS=8. 

From the app documentation for the BITS parameter:
> The output data type will be automaticity choosen using this value. A value of 8 will create unsigned byte output files. Values from 9 to 16 will create unsigned word output files. Unused bits in the unsigned word output file will be set to zero.

The way I understand this is that there are 16 "sample bits" for BITS=[9, 16] and 8 for BITS=8. I seem to remember talking to @scsides about this back when I made the issue.

Before merging:
I think this should be verified by someone more familiar with this app / hirise / pds images.
Would this require a changeLog entry? IIRC small output differences like this can be time-consuming for products / missions to validate.

## Related Issue

#4006 

## Motivation and Context

Closes an issue for ISIS support milestone. Fixes unintended output for hideal2pds.

## How Has This Been Tested?

Output from makefile tests verifies that these changes work as intended

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/mro/apps/hideal2pds/main.cpp
+++ b/isis/src/mro/apps/hideal2pds/main.cpp
@@ -197,6 +197,11 @@ void IsisMain() {
   image.addKeyword(PvlKeyword("SAMPLE_BIT_MASK", toString((int)pow(2.0, (double)nbits) - 1)),
                    Pvl::Replace);
 
+  // SAMPLE_BITS defaults to 16 but should be 8 when BITS=8
+  if( nbits == 8 ) {
+    image.addKeyword(PvlKeyword("SAMPLE_BITS", toString(nbits)), Pvl::Replace);
+  }
+
   Camera *cam = inputCube->camera();
   updatePdsLabelRootObject(isisCubeLab, pdsLabel, ui, cam);
 


isis/src/mro/apps/hideal2pds/main.cpp
```diff
@@ -197,6 +197,11 @@ void IsisMain() {
   image.addKeyword(PvlKeyword("SAMPLE_BIT_MASK", toString((int)pow(2.0, (double)nbits) - 1)),
                    Pvl::Replace);
 
+  // SAMPLE_BITS defaults to 16 but should be 8 when BITS=8
```
scsides, line 200: Changing the mask without changing the bit type of the output doesn't make sense to me. See #4006. These two keywords have to be in sync with each other. See line 131 in this file. It appears to be setting unsignedword for an 8bit output request. I wonder if the mask would work correctly without the fix below if line 131 was set to unsignedbyte????

scsides (changes requested):

isis/src/mro/apps/hideal2pds/main.cpp
tgiroux: Good eye. Changing that output type to unsignedbyte produces correct mask and samples_bits values in the output pvl for BITS=8.

Line 197 is still needed for BITS values 9-15, because the mask always defaults to the full 16 bits. ( #3978 )

Also, I was under the impression that (output pvl) SAMPLE_BITS should be 16 for (user-entered) BITS=[9, 16] and 8 for BITS=8. Is that correct?

Issue #4410, opened 2021-04-19
Bundle Adjusting CSM images


jessemapel:
## Description

With #4291 we now have CSM camera models working in ISIS. The next step is to add the ability to bundle adjust CSM camera models with jigsaw. We've done some planning and have a good idea of how to handle the differences between CSM and ISIS sensors, but there are a few places we would like more input from users.

### Background

First of all, a brief overview of bundle adjustment in CSM vs ISIS. ISIS bundle adjustments works by adjusting constant offsets or polynomial offsets to the position and pointing where as CSM bundle adjustment just adjusts generic parameters. The way that CSM handles bundle adjustment is just a more general way than the way that ISIS does. So, we're planning to leverage this in the actual implementation.

When you adjust a CSM sensor model, you specify a set of parameters that you want to adjust. The parameters available are specific to that model and are completely opaque to the user outside of a name and a [type](https://github.com/sminster/csm/blob/master/csm.h#L90). For example with our USGSCSM line scan model the options are:

* IT Pos. Bias
* CT Pos. Bias
* Rad Pos. Bias
* IT Vel. Bias
* CT Vel. Bias
* Rad Vel. Bias
* Omega Bias
* Phi Bias
* Kappa Bias
* Omega Rate
* Phi Rate
* Kappa Rate
* Omega Accl
* Phi Accl
* Kappa Accl
* Focal Bias

Each parameter can be turned on or off independently and the sensor model will handle the a priori values and covariances between them. You can also specify a [set](https://github.com/sminster/csm/blob/master/csm.h#L111) of parameters you want to work with. There are also more complex options such as parameter sharing that we will likely not get to.

The key difference between ISIS and CSM parameters is that with ISIS we have full knowledge of each parameter and how it translates to the orientation of the model, but with CSM we only know a small amount. For this reason, setting parameters for CSM sensor models is going to require users to be knowledgeable about the parameters the model they are using provides.

### Changes to arguments

This feature should be purely additive and not interfere with how jigsaw currently works. For this reason we will begin by adding a new parameter that specifies a type or set of CSM parameters to adjust. All parameters of the specified type or set will be enabled for all CSM images when used.

[...]

PR #4419, opened 2021-04-26
Csmobservation

jessemapel:
First pass at CSM observation class and settings class. This only allows for setting the parameters using a parameter set, not a type or explicit list.

References #4410

### DIFF
--- a/isis/src/base/objs/CSMCamera/CSMCamera.h
+++ b/isis/src/base/objs/CSMCamera/CSMCamera.h
@@ -116,6 +116,10 @@ namespace Isis {
       virtual double RightAscension();
       virtual double Declination();
 
+      std::vector<int> getParameterIndices(csm::param::Set paramSet) const;
+      void applyParameterCorrection(int index, double correction);
+      double getParameterCovariance(int index1, int index2);
+
     protected:
       void setTarget(Pvl label);
 
--- a/isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
+++ b/isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
@@ -18,7 +18,6 @@ find files of those names at the top level of this repository. **/
 #include "LinearAlgebra.h"
 
 namespace Isis {
-  class BundleObservationSolveSettings;
 
   /**
    * @brief Class for bundle observations
@@ -43,7 +42,7 @@ namespace Isis {
       virtual AbstractBundleObservation &operator=(const AbstractBundleObservation &src);
 
       // copy method
-      // not implementedn in BundleObservation either??? 
+      // not implementedn in BundleObservation either???
 //      virtual void copy(const AbstractBundleObservation &src);
 
       virtual void append(const BundleImageQsp &value);
@@ -61,7 +60,7 @@ namespace Isis {
       virtual LinearAlgebra::Vector &adjustedSigmas();
 
 
-      virtual const BundleObservationSolveSettingsQsp solveSettings() = 0;              
+      virtual const BundleObservationSolveSettingsQsp solveSettings() = 0;
       virtual int numberParameters();
       virtual bool applyParameterCorrections(LinearAlgebra::Vector corrections);
 
@@ -73,7 +72,7 @@ namespace Isis {
       virtual QStringList parameterList();
       virtual QStringList imageNames();
 
-    private:
+    protected:
       QString m_observationNumber; /**< This is typically equivalent to serial number
                                         except in the case of "observation mode" (e.g.
                                         Lunar Orbiter) where for each image in the
--- a/isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.cpp
+++ b/isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.cpp
@@ -1,0 +1,876 @@
+/** This is free and unencumbered software released into the public domain.
+
+The authors of ISIS do not claim copyright on the contents of this file.
+For more details about the LICENSE terms and the AUTHORS, you will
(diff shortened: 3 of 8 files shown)


isis/src/base/objs/CSMCamera/CSMCamera.h
```diff
@@ -116,6 +116,10 @@ namespace Isis {
       virtual double RightAscension();
       virtual double Declination();
 
+      std::vector<int> getParameterIndices(csm::param::Set paramSet) const;
+      void applyParameterCorrection(int index, double correction);
+      double getParameterCovariance(int index1, int index2);
```
jessemapel, line 121: Note that these are not exposed via the Camera API so we have to dynamic cast to get to them.

isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
```diff
@@ -73,7 +72,7 @@ namespace Isis {
       virtual QStringList parameterList();
       virtual QStringList imageNames();
 
-    private:
+    protected:
```
jessemapel, line 75: The members below here are duplicated in the BundleObservation class. They probably need to go away.

isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.cpp
```diff
@@ -0,0 +1,876 @@
...
+
+
+  // TODO change this to CSM solve settings once compute partials is updated
+  /**
+   * Accesses the solve settings
+   *
+   * @return @b const BundleObservationSolveSettingsQsp Returns a pointer to the solve
+   *                                                    settings for this CsmBundleObservation
+   */
+  const BundleObservationSolveSettingsQsp CsmBundleObservation::solveSettings() {
```
jessemapel, line 136: This returns the wrong type of solve settings because of how the abstract interface works. It expects to return an ISIS settings. The only place it's used is in computePartials, so once we get that updated we should remove this from the abstract class so that we can have unique settings.

isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.cpp
```diff
@@ -0,0 +1,876 @@
...
+ *     called from BundleSolutionInfo::outputImagesCSV().  It is set to false by default
+ *     for backwards compatibility.
+ *
+ * @return @b QString Returns a formatted QString representing the CsmBundleObservation
+ *
+ * @internal
+ *   @history 2016-10-26 Ian Humphrey - Default values are now provided for parameters that are
+ *                           not being solved. Fixes #4464.
+ */
+QString CsmBundleObservation::formatBundleOutputString(bool errorPropagation, bool imageCSV) {
```
jessemapel, line 209: I left these two functions unimplemented until we figure out how we want to report stuff.

isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.h
```diff
@@ -0,0 +1,79 @@
...
+      CsmBundleObservation &operator=(const CsmBundleObservation &src);
+
+      // copy method
+      void copy(const CsmBundleObservation &src);
+
+      bool setSolveSettings(CsmBundleObservationSolveSettingsQsp solveSettings);
+
+      int numberParameters();
+
+      const BundleObservationSolveSettingsQsp solveSettings();
```
jessemapel, line 61: Again, returning the wrong type to satisfy the parent API.

isis/src/control/objs/CsmBundleObservation/CsmBundleObservationSolveSettings.h
```diff
@@ -0,0 +1,42 @@
...
+    private:
+      csm::param::Set m_solveSet; //! The set of parameters to solve for
+
+  };
+
+  typedef QSharedPointer<CsmBundleObservationSolveSettings> CsmBundleObservationSolveSettingsQsp;
+
+}
+
+#endif // CsmBundleObservationSolveSettings_h
```
jessemapel, line 42: This class is a dumb holder for right now, but I plan for it to encapsulate the different ways you can specify things:

1. parameter set
2. parameter type
3. explicit parameter names

isis/tests/CSMCameraTests.cpp
```diff
@@ -250,6 +251,49 @@ TEST_F(CSMCameraSetFixture, EmissionAngle) {
...
+      .WillOnce(::testing::Return(3));
+  EXPECT_CALL(mockModel, getParameterType(0))
+      .Times(1)
+      .WillOnce(::testing::Return(csm::param::REAL));
+  EXPECT_CALL(mockModel, getParameterType(1))
+      .Times(1)
+      .WillOnce(::testing::Return(csm::param::REAL));
+  EXPECT_CALL(mockModel, getParameterType(2))
+      .Times(1)
+      .WillOnce(::testing::Return(csm::param::REAL));
```
jessemapel, line 267: The mock is a little weird here because GeometricModel::getParameterSetIndices is not virtual so we can't mock it out. We have to mock out the functions it calls. See

https://github.com/sminster/csm/blob/master/GeometricModel.cpp#L58

krlberry (approved):

isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
krlberry: Agreed. Making the values protected in the parent and removing from the child classes works as a solution and I merged this, but am still potentially interested in investigating if it's possible to handle weights/corrections/apriori/adjustedsigmas at the parent level if we were to modify how these are applied to be be more generalized.

PR #4430, opened 2021-04-29
Moved CSM settings into BOSS

jessemapel:
Changed the CSM bundle settings to be included with the regular bundle settings. This allows us to easily store the settings prior to creating the bundle observations and we don't have to modify the bundle observation API as much.

There is still some work to be done on this that I'm going to continue working on, but I want to get this up for review at this point.


isis/src/control/objs/BundleUtilities/BundleObservationSolveSettings.cpp
```diff
@@ -240,25 +240,6 @@ namespace Isis {
   }
 
 
-//  BundleObservationSolveSettings::BundleObservationSolveSettings(const BundleObservationSolveSettings &other)
```
krlberry, line 243: Was this commented-out code that existed prior to this project or is this newly removed?

krlberry (approved):
This looks good to me. I like the options of solving by CSM paramter type, set, or list. Can I merge this or should I wait for more updates?

isis/src/control/objs/BundleUtilities/BundleObservationSolveSettings.cpp
jessemapel: That already existed prior to this. There's so much random commented out code that we should clean out what we can.

PR #4436, opened 2021-05-04
Moves partial calculation into CSM and ISIS observation classes.

krlberry:
## Description
Moves partials calculation into CSM and ISIS observation classes to support Bundle Adjustment with both ISIS and CSM sensor models. 

Works in ISIS with the following tests definitely passing: ctest -R Jigsaw && ctest -R jigsaw && ctest -R Bundle all pass
(otherwise untested)

Ready for review. 

Also includes a minor change to the jigsaw.xml file to make the new `CSMSOLVESET` parameter optional. This is required to get old-style (`ctest -R jigsaw`) tests to pass. 

## Related Issue
#4410 

## Motivation and Context
See #4410 

## How Has This Been Tested?
ctest -R Jigsaw && ctest -R jigsaw && ctest -R Bundle all pass

## Types of changes

## Checklist:

## Licensing


isis/src/control/objs/CsmBundleObservation/CsmBundleObservation.cpp
```diff
@@ -873,4 +873,24 @@ QString CsmBundleObservation::formatBundleOutputString(bool errorPropagation, bo
 
     return "";
   }
+
```
krlberry, line 885: whoops, I just thought I'd left this file out of the PR completely, but this is an old version. I'll update it.

isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
```diff
@@ -72,6 +75,12 @@ namespace Isis {
       virtual QStringList parameterList();
       virtual QStringList imageNames();
 
+      virtual bool computeTargetPartials(LinearAlgebra::Matrix &coeffTarget, BundleMeasure &measure,
```
krlberry, line 78: @jessemapel Do these seem reasonable to you: I switched from passing the serial number to the measure itself because I needed access to the camera, and parent point in a lot of these. Is there a good way to get these from the serial number that I missed?

jessemapel: You can get the camera from the BundelImage in the observation. You can use m_cubeSerialNumberToBundleImageMap to get it by serial number.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1896,234 +1880,68 @@ namespace Isis {
...
-        index++;
-      }
-
-      // Add the partial for the z coordinate of the position
-      for (int cameraCoef = 0; cameraCoef < numCamPositionCoefficients; cameraCoef++) {
-        measureCamera->GroundMap()->GetdXYdPosition(SpicePosition::WRT_Z, cameraCoef,
-                                                    &coeffImage(0, index),
-                                                    &coeffImage(1, index));
-        index++;
+    if (measureCamera->GetCameraType() != 6) {
```
jessemapel: Why is this checking the camera type? It should just make the same calls for both CSM and ISIS, that's why we wanted to move the logic into the observation.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1896,234 +1880,68 @@ namespace Isis {
...
 
-    obsValue = deltaY / measureCamera->PixelPitch();
-    m_bundleResults.addResidualsProbabilityDistributionObservation(obsValue);
+      obsValue = deltaY / measureCamera->PixelPitch();
+      m_bundleResults.addResidualsProbabilityDistributionObservation(obsValue);
 
-    observationSigma = 1.4 * measureCamera->PixelPitch();
-    observationWeight = 1.0 / observationSigma;
+      double observationSigma = 1.4 * measureCamera->PixelPitch();
+      observationWeight = 1.0 / observationSigma;
```
jessemapel: Observation weighting should be done in the observation. Then, the RHS should be passed to the probability distributions.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -1411,4 +1363,231 @@ QString BundleObservation::formatBundleOutputString(bool errorPropagation, bool
 
     return finalqStr;
   }
+
+
+  bool BundleObservation::computeTargetPartials(matrix<double> &coeffTarget, BundleMeasure &measure,
+                                                BundleSettingsQsp &bundleSettings, BundleTargetBodyQsp &bundleTargetBody) {
+    if (bundleSettings->solveTargetBody()) {
```
jessemapel: It doesn't make sense to check if we're solving for target body in the function specifically for solving for target body. Assume this is true already by here.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -1411,4 +1363,231 @@ QString BundleObservation::formatBundleOutputString(bool errorPropagation, bool
...
+
+    return true;
+  }
+
+
+  bool BundleObservation::computeImagePartials(matrix<double> &coeffImage, BundleMeasure &measure) {
+    coeffImage.clear(); 
+
+    const BundleObservationSolveSettingsQsp observationSolveSettings =
+        measure.observationSolveSettings();
```
jessemapel: The observation stores its own solve settings, you should use them, not the ones from the measure. We should probably remove them from the measure.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -1411,4 +1363,231 @@ QString BundleObservation::formatBundleOutputString(bool errorPropagation, bool
...
+    }
+
+    return true;
+  }
+
+
+  bool BundleObservation::computePoint3DPartials(matrix<double> &coeffPoint3D, BundleMeasure &measure, SurfacePoint::CoordinateType coordType) {
+    coeffPoint3D.clear();
+    Camera *measureCamera = measure.camera();
+    BundleControlPoint* point = measure.parentControlPoint();
```
jessemapel, line 1591: For this function I think passing the measure makes sense.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -1411,4 +1363,231 @@ QString BundleObservation::formatBundleOutputString(bool errorPropagation, bool
...
+    double computedX, computedY;
+    if (!(measureCamera->GroundMap()->GetXY(point->adjustedSurfacePoint(),
+                                            &computedX, &computedY, false))) {
+      QString msg = "Unable to map apriori surface point for measure ";
+      msg += measure.cubeSerialNumber() + " on point " + point->id() + " into focal plane";
+      throw IException(IException::User, msg, _FILEINFO_);
+    }
+
+    double measuredX = measure.focalPlaneMeasuredX();
+    double measuredY = measure.focalPlaneMeasuredY();
```
jessemapel, line 1647: This may also require the measure for these two calls?

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
krlberry: Yeah, I agree. I added that back in when I was trying to figure out what broke this morning and forgot to take it out. Thanks for catching it!

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
krlberry: Agreed. I think separate PR to remove the observations from the measure?

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
krlberry: Agreed.

krlberry:
@jessemapel I ended up needing to leave `BundleMeasure measure` in the function signature for `BundleObservation::computeImagePartials()` because I needed the measure to access the parent control point in `CsmObservation::computeImagePartials()`, since CSM's sensor partials function requires the surface point.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1896,234 +1880,69 @@ namespace Isis {
...
-        measureCamera->GroundMap()->GetdXYdPosition(SpicePosition::WRT_Y, cameraCoef,
-                                                    &coeffImage(0, index),
-                                                    &coeffImage(1, index));
-        index++;
+    if (measureCamera->GetCameraType() != 6) {
+      // Compute the look vector in instrument coordinates based on time of observation and apriori
+      // lat/lon/radius.  As of 05/15/2019, this call no longer does the back-of-planet test. An optional
+      // bool argument was added CameraGroundMap::GetXY to turn off the test.
+      double computedX, computedY;
+      if (!(measureCamera->GroundMap()->GetXY(point.adjustedSurfacePoint(),
```
jessemapel, line 1898: These are never used here. This entire if block should go.

krlberry: Could we actually chat about this post-standup? It does change the state of the camera ground map, and breaks everything if I pull it out. I'm concerned about the extra time it might take to run this in each individual "compute partials" function, but haven't tested this yet. I'm considering pulling it into a partialsSetup function, but not sure if that's a good idea or not.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
krlberry: Weighting has been moved to the observation classes, unless the maximum likelihood option is used in the bundle, in which case some additional weighting is done in `BundleAdjust::computePartials()`

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
krlberry: Discussed in person -- the `CameraGroundMap()` doesn't exist for csm cameras, and we don't want to (a) do an image-to-ground call in each newly added partials function which would get called every time or (b) have a `observation::setupPartials()` function that gets called once for each observation, so we decided to leave this in. There's already a check to see if the camera is a framer, so it's not completely out of place to check if it's an Isis or Csm sensor.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
krlberry: Discussed in person, this is actually needed for Isis observations, so it's going to stay in.

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -848,6 +894,17 @@ namespace Isis {
     return m_model->getParameterCovariance(index1, index2);
   }
 
+  
+  vector<double> CSMCamera::getSensorPartials(int index, SurfacePoint groundPoint) {
+    // csm::SensorPartials holds (line, sample) in order for each parameter
+    //   typedef std::pair<double,double> SensorPartials;
```
krlberry: This comment shouldn't be in here. I'll take it out.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1861,29 +1861,13 @@ namespace Isis {
    * @throws IException::User "Unable to map apriori surface point for measure"
    */
   bool BundleAdjust::computePartials(matrix<double> &coeffTarget,
-                                     matrix<double> &coeffImage,
-                                     matrix<double> &coeffPoint3D,
-                                     vector<double> &coeffRHS,
-                                     BundleMeasure &measure,
-                                     BundleControlPoint &point) {
+                                       matrix<double> &coeffImage,
```
krlberry: This extra spacing also shouldn't be in here.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1896,237 +1880,65 @@ namespace Isis {
...
+
       int currentModelIndex = m_bundleResults.maximumLikelihoodModelIndex();
       observationWeight *= m_bundleResults.maximumLikelihoodModelWFunc(currentModelIndex)
                             .sqrtWeightScaler(residualR2ZScore);
     }
 
-    // multiply coefficients by observation weight
     coeffImage *= observationWeight;
     coeffPoint3D *= observationWeight;
     coeffRHS *= observationWeight;
```
krlberry: These needed to stay in in case the maximum likelihood is used and needs to update the weighting. The default weighting has now been moved into the observation classes.

isis/src/control/objs/CsmBundleObservation/CsmBundleObservationSolveSettings.h
```diff
@@ -0,0 +1,42 @@
+#ifndef CsmBundleObservationSolveSettings_h
```
krlberry: This file got deleted, right? I'll pull it back out.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
jessemapel: Could this just be moved into the if using maximum likelihood block? That would better isolate it.

jessemapel (commented):
Looks good, just one more clean-up thing.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
krlberry: Good point! Not sure why I didn't do that in the first place!

jessemapel (approved):

PR #4444, opened 2021-05-10
Implemented bundleOutputString for CSM

amystamile-usgs:
## Description

Implemented bundleOutputString for CSM. 

## Related Issue

#4410 

## Motivation and Context

This allows us to output CSM observation values properly to the bundleout.txt file.

## How Has This Been Tested?

Added unit test for bundleOutputString and added unit tests for added functions to CSMCamera

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have read and agree to abide by the [Code of Conduct](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/Code-Of-Conduct.md)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.
- [ ] I have added any user impacting changes to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #4446, opened 2021-05-11
Csmcsv

jessemapel:
## Description

Updates BundleAdjust to output separate Images CSV files for each sensor. This solves the issue of mismatched columns when different images are solving for different parameters. Currently when different parameters sets are being solved for only the very first observation's parameters are represented in the header and the images CSV is malformed with variable numbers of columns on each line.

Related to #4410 

## Examples

This impacts CSM bundle, multi-sensor bundle with and without scconfig, and held images. Here's a few examples and how this impacts them:

### CSM bundle

If you are bundle adjusting with multiple CSM sensors, then each CSM sensor will have a separate images CSV containing the updates for that sensor's image parameters.

### Multi-Sensor bundle with scconfig

If you are bundle adjusting with multiple ISIS sensors using scconfig, then each separate entry in your scconfig will result in a separate images CSV. For example, the following scconfig:

```
Object = SensorParameters
  Group = MARS_ODYSSEY/THEMIS_IR
    CamSolve=accelerations
    OverExisting=yes
    Twist=yes
    camera_angles_sigma=.25
    camera_angular_velocity_sigma=.1
    camera_angular_acceleration_sigma=.01
    SPSolve=none
  EndGroup
  Group = VIKING_ORBITER_1/VISUAL_IMAGING_SUBSYSTEM_CAMERA_B
    camsolve=angles
    Twist=yes
    camera_angles_sigma=1.0
    SPSolve=none
  EndGroup
  Group = VIKING_ORBITER_2/VISUAL_IMAGING_SUBSYSTEM_CAMERA_A
    camsolve=angles
    Twist=yes
    camera_angles_sigma=1.0
    SPSolve=none
  EndGroup
EndObject
End
```

You will have 3 images CSV files:

 - bundleout_images_MARS_ODYSSEY_THEMIS_IR.csv
 - bundleout_images_VIKING_ORBITER_1_VISUAL_IMAGING_SUBSYSTEM_CAMERA_B.csv
 - bundleout_images_VIKING_ORBITER_2_VISUAL_IMAGING_SUBSYSTEM_CAMERA_A.csv

### Multi-Sensor bundle without scconfig

For example if you are bundle adjusting a set of LRO NAC images that contains both NAC left and NAC right images you will still get separate images CSV files:

 - bundleout_images_LUNARRECONNAISSANCEORBITER_NACL.csv
 - bundleout_images_LUNARRECONNAISSANCEORBITER_NACR.csv

Even if you are solving for the same parameters for both sensors.

### Held images bundle adjustment

If you are holding some images fixed using jigsaw's `HELDLIST` parameter then you will have one images CSV file your your held images called bundleout_images_held.csv and another images CSV file for the rest of your images. The name of your non-helf images CSV will again contain the spacecraft and sensor name.

## Testing

The following tests are expected to fail due to updated truth data:

 - jigsaw_app_test_radar (Failed)
 - jigsaw_app_test_scconfig (Failed)
 - jigsaw_app_test_scconfigHeld (Failed)


isis/src/control/apps/jigsaw/tsts/scconfigHeld/Makefile
```diff
@@ -13,9 +13,18 @@ commands:
...
+	   > $(OUTPUT)/bundleout_scconfig_images_themis.csv;
+	$(CAT) bundleout_images_VIKING_ORBITER_1_VISUAL_IMAGING_SUBSYSTEM_CAMERA_B.csv \
+	   | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+	   > $(OUTPUT)/bundleout_scconfig_images_viking_1.csv;
+	$(CAT) bundleout_images_VIKING_ORBITER_2_VISUAL_IMAGING_SUBSYSTEM_CAMERA_A.csv \
+          | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+          > $(OUTPUT)/bundleout_scconfig_images_viking_2.csv;
+	$(CAT) bundleout_images_held.csv \
+          | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+          > $(OUTPUT)/bundleout_scconfig_images_held.csv;
```
jessemapel, line 27: This is a good example of what the new output can look like. This test has THEMIS images, Viking 1 images, VIking 2 images, and a held image.

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -905,6 +905,30 @@ namespace Isis {
...
+   *
+   * @param index The index of the parameter
+   *
+   * @return @b double The parameter value
+   */
+  double CSMCamera::getParameterValue(int index) const {
+    return m_model->getParameterValue(index);
+  }
+
+
```
jessemapel: These need tests and will likely conflict with #4444

isis/src/control/objs/BundleResults/unitTest.cpp
```diff
@@ -81,8 +80,13 @@ namespace Isis {
        * Destroys the tester object
        */
       ~BundleResultsXmlHandlerTester() {
+        if (m_file.exists()) {
+          m_file.remove();
+        }
       }
 
+      QFile m_file;
+
```
jessemapel, line 89: This is a small fix to an unrelated issue where this file never got cleaned-up.

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.cpp
```diff
@@ -1161,82 +1085,113 @@ namespace Isis {
...
-        fpOut << toString(rmsImageSampleResiduals[imgIndex].Rms()).toLatin1().data();
-        sprintf(buf,",");
-        fpOut << buf;
+          // TODO this is a bad linear search, can we change the rms vectors to maps?
+          int observationIndex = m_statisticsResults->observations().indexOf(observation);
+          int imgIndex = k; // Start at k to account for the images in this observation
+          // Accumulate all of the images from previous observations
+          for (int obsIndex = 0; obsIndex < observationIndex; obsIndex++) {
+            imgIndex += m_statisticsResults->observations().at(obsIndex)->size();
+          }
```
jessemapel, line 1157: I want to make this part better, but it'll be a lot of work to modify the RMS vectors so I'm starting with this change.

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.cpp
```diff
@@ -1161,82 +1085,113 @@ namespace Isis {
...
 
-      for (int j = 0; j < numImages; j++) {
+        if(!observation) {
+          continue;
+        }
 
-        BundleImageQsp image = observation->at(j);
+        int numImages = observation->size();
 
+        for (int k = 0; k < numImages; k++) {
```
jessemapel, line 1147: This triple nested for loop isn't as bad as it looks because collectively it still just iterates over all the images. We just split the images up into 3 different tiers of sets. First we separate by instrument ID, then by observation, then by image.

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.h
```diff
@@ -13,6 +13,7 @@ find files of those names at the top level of this repository. **/
 #include <QObject>
 #include <QString>
 
+#include "AbstractBundleObservation.h"
```
jessemapel, line 16: This could be forward declared as a typedef but that's rather ugly, can be confusing, and even cause errors; so I just included it.

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.truth
```diff
@@ -424,12 +424,12 @@ Serializing test XML object to file...
...
-                    <parameter>  Z  (km)  </parameter>
-                    <parameter> RA  (dd)  </parameter>
-                    <parameter>DEC  (dd)  </parameter>
-                    <parameter>TWI  (dd)  </parameter>
+                    <parameter>X</parameter>
+                    <parameter>Y</parameter>
+                    <parameter>Z</parameter>
+                    <parameter>RA</parameter>
+                    <parameter>DEC</parameter>
+                    <parameter>TWIST</parameter>
```
jessemapel, line 432: This is the one big functional difference with the new parameter list implementation, it doesn't have units anymore.

isis/src/control/objs/BundleUtilities/AbstractBundleObservation.h
```diff
@@ -92,7 +92,6 @@ namespace Isis {
       //! Map between cube serial number and BundleImage pointers.
       QMap<QString, BundleImageQsp> m_cubeSerialNumberToBundleImageMap;
       QStringList m_serialNumbers;      //!< List of all cube serial numbers in observation.
-      QStringList m_parameterNamesList; //!< List of all cube parameters.
```
jessemapel, line 95: I don't want to have to store a whole bunch of strings for each observation, so I decided to just dynamically generate this for memory efficiency. It requires the same amount of code as generating and then saving them off anyway.

isis/src/control/objs/BundleUtilities/BundleObservation.cpp
```diff
@@ -936,6 +931,88 @@ QString BundleObservation::formatBundleOutputString(bool errorPropagation, bool
...
+ * @return @b QStringList List of observation parameter names
+ */
+QStringList BundleObservation::parameterList() {
+  QStringList paramList;
+
+  // We still want to output the center postion even if not solving for it
+  // so always do at least 1
+  int numberCamPosCoefSolved = std::max(
+      solveSettings()->numberCameraPositionCoefficientsSolved(),
+      1);
```
jessemapel, line 950: I do not like this but it's needed to preserve functionality. Ideally, this wouldn't be here and the twist would not be there if we weren't solving for it, but that's not the current functionality for the output files.

isis/src/control/objs/BundleUtilities/BundleObservationVector.cpp
```diff
@@ -159,52 +163,18 @@ namespace Isis {
...
-   * contained BundleObservations.
-   *
-   * @return @b int Returns the total number of pointing parameters for the BundleObservations
-   */
-  int BundleObservationVector::numberPointingParameters() {
-    int pointingParameters = 0;
+      // update instrument ID to observation ptr map
+      // separate held observations out into their own entry
+      if (bundleObservation->numberParameters() == 0) {
+        m_instIdToObservationMap.insertMulti("held", bundleObservation);
```
jessemapel, line 170: This is how we generate a separate images.csv for held images. This works for the scconfigHeld test, but not for FunctionalTestJigsawHeldList I haven't figured out why yet.

isis/src/control/objs/BundleUtilities/BundleObservationVector.cpp
```diff
@@ -215,8 +185,12 @@ namespace Isis {
...
    */
   int BundleObservationVector::numberParameters() {
-    //TODO: change this to include CSM parameters
-    return numberPositionParameters() + numberPointingParameters();
+    int numParameters = 0;
+
+    for (int i = 0; i < size(); i++) {
+      numParameters += at(i)->numberParameters();
+    }
+    return numParameters;
```
jessemapel, line 193: @kberryUSGS There's a chance you already made this change.

isis/tests/FunctionalTestsJigsaw.cpp
```diff
@@ -526,98 +526,107 @@ TEST_F(ObservationPair, FunctionalTestJigsawCamSolveAll) {
...
-  EXPECT_NEAR(elems.at(56).toDouble(), 0.365064224,       0.00001); 
-  
-  
-  line = file.readLine();
-  elems = line.split(","); 
-  
+  EXPECT_NEAR(elems.at(56).toDouble(), 0.365064224,       0.00001);
+
+
+  QFile rightFile(prefix.path() + "/bundleout_images_LUNARRECONNAISSANCEORBITER_NACR.csv");
```
jessemapel, line 580: This test is an example of how the new images.csv output could be annoying. The NACL and NACR are two copies of the same system and we would solve for their images exactly the same way, but because they are actually different instrument IDs, we will output two images.csv files when we don't need to.

I want to talk to some users to see how they feel about this change.

isis/src/base/objs/CSMCamera/CSMCamera.cpp
amystamile-usgs: Tests for these functions have been added in my PR.

isis/src/control/apps/jigsaw/tsts/scconfigHeld/Makefile
```diff
@@ -13,9 +13,18 @@ commands:
...
+	$(CAT) bundleout_images_MARS_ODYSSEY_THEMIS_IR.csv \
+	   | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+	   > $(OUTPUT)/bundleout_scconfig_images_themis.csv;
+	$(CAT) bundleout_images_VIKING_ORBITER_1_VISUAL_IMAGING_SUBSYSTEM_CAMERA_B.csv \
+	   | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+	   > $(OUTPUT)/bundleout_scconfig_images_viking_1.csv;
+	$(CAT) bundleout_images_VIKING_ORBITER_2_VISUAL_IMAGING_SUBSYSTEM_CAMERA_A.csv \
+          | perl -pe 's/(^|,|: )([^,:]+\/)([^,\/:]*\.)(net|cub)/\1\3\4/g' 2> /dev/null \
+          > $(OUTPUT)/bundleout_scconfig_images_viking_2.csv;
+	$(CAT) bundleout_images_held.csv \
```
krlberry, line 25: Is this just an improvement to the existing test that doesn't have anything to do with the output format change?

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.truth
krlberry: Are the units always the same? So in the docs, we can say X,Y, Z are always in km, RA/DEC/TWIST are all in deicmal degrees.

isis/src/control/objs/BundleUtilities/BundleUtilities.truth
```diff
@@ -501,9 +501,7 @@ TWI  (dd)          0.00000000            0.00000000               0.00000000
 index =  "1"
 instrument id =  "InstrumentId2"
 number parameters =      "3"
-number position param =  "0"
-number pointing param =  "3"
-parameter list:  ()
+parameter list:  ("X", "Y", "Z", "RA", "DEC", "TWIST")
```
krlberry, line 504: If this is replicating the old test, shouldn't the parameter list just be RA/DEC/TWIST for 3 pointing parameters and 0 position parameters?

isis/src/control/objs/BundleUtilities/BundleUtilities.truth
```diff
@@ -501,9 +501,7 @@ TWI  (dd)          0.00000000            0.00000000               0.00000000
 index =  "1"
 instrument id =  "InstrumentId2"
 number parameters =      "3"
```
krlberry, line 503: And how is number parameters still 3 if there are 6 things in the parameter list?

isis/tests/FunctionalTestsJigsaw.cpp
krlberry: That sounds like a great idea. If it does turn out that the users don't like this, is there a straightforward way to produce just one output CSV in the case where the solve parameters are identical?

isis/src/control/objs/BundleUtilities/BundleObservationVector.cpp
krlberry: I did this a bit differently locally, but your version is better, taken along with your other changes.

krlberry (commented):
This looks good to me. Some comments / questions.

isis/src/control/objs/BundleSolutionInfo/BundleSolutionInfo.truth
jessemapel: For ISIS the units are always the same. In the output files, though we have separate ways to output the units.

isis/src/control/apps/jigsaw/tsts/scconfigHeld/Makefile
jessemapel: This just changed the sed to run on each images CSV now

isis/src/control/objs/BundleUtilities/BundleUtilities.truth
jessemapel: There's a weird thing with the ISIS observations where you always have x,y,z,ra,dec,twist because we always output the center value even if we don't solve for them. This is keeping consistent with how it used to work, it's just set earlier now. Before the parameter list was set after the output files were written.

krlberry: Thanks for the explanation! That really confused me.

isis/src/control/objs/BundleUtilities/BundleUtilities.truth
krlberry: Answered above.

isis/src/control/apps/jigsaw/tsts/scconfigHeld/Makefile
krlberry: Thanks just wanted to double check.

krlberry (approved):

krlberry (approved):

PR #4464, opened 2021-05-13
Gets csm bundle converging

jessemapel:
This PR gets the CSM bundle converging for our test case. Note, you cannot solve for all parameters as there is not enough data and you have to do a rectangular solve because we haven't implemented latitudinal ground partials for CSM yet.

References #4410

### DIFF
--- a/isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
+++ b/isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
@@ -497,12 +497,12 @@ namespace Isis {
 
     if (m_bundleSettings->solveTargetBody()) {
       m_rank += m_bundleSettings->numberTargetBodyParameters();
-      
+
       if (m_bundleTargetBody->solveMeanRadius() || m_bundleTargetBody->solveTriaxialRadii()) {
         outputBundleStatus("Warning: Solving for the target body radii (triaxial or mean) "
                            "is NOT possible and likely increases error in the solve.\n");
       }
-      
+
       if (m_bundleTargetBody->solveMeanRadius()){
         // Check if MeanRadiusValue is abnormal compared to observation
         bool isMeanRadiusValid = true;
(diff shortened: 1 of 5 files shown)


isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1857,6 +1857,8 @@ namespace Isis {
                                      BundleMeasure &measure,
                                      BundleControlPoint &point) {
 
+    std::cout << "Computing partials for [" << measure.cubeSerialNumber() << "] in [" << point.id() << "]." << std::endl;
```
krlberry, line 1860: I think it probably makes sense to leave this debugging print statement in here, just want to note that it is in here.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -1967,8 +1970,8 @@ namespace Isis {
 
       if (m_bundleSettings->solveTargetBody()) {
         // TODO: needs to be updated for ISIS vs. CSM CSM has no updateBodyRotation]
-        // TODO: this is no good. 
-        QSharedPointer<BundleObservation> isisObservation = qSharedPointerDynamicCast<BundleObservation>(observation);  
+        // TODO: this is no good.
```
krlberry, line 1973: In a future PR, these TODOs should be removed. The second one was mine and is not specific enough to be useful.

krlberry: (to remove the first TODO, I guess this should probably also throw an error (or perform some other expected behavior) if this is attempted with a CSM sensor.)

krlberry (approved):
Made some notes about future work to be done, but merging this so we have the running-to-completion bundle code in! :tada:

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
jessemapel: Oops I thought i took all of these out.

PR #4479, opened 2021-05-19
PDS4 - topds4 TEMPLATE default option, working TGOCaSSIS.tpl

tgiroux:
## Description

The TEMPLATE option can still be user-entered. Otherwise it defaults to $ISISROOT/appdata/export/<SpacecraftID><InstrumentId>.tpl 
An exception is thrown if the template does not exist. I also added a $ISISROOT/appdata/export/TGOCaSSIS.tpl that mimics the functionality of ProcessExportPDS4 for tgo cassis.

I didn't update the info on topds4.xml, but docs should be updated before any of this gets to dev.

## Motivation and Context

Use inja templating engine for exporting isis cubes.

## How Has This Been Tested?

Taking the cube from `$ISISTESTDATA/isis/src/tgo/apps/tgocassisrdrgen/tsts/default/input/CAS-MCO-2016-11-26T22.32.39.582-BLU-03025-00.cub`
Running: `topds4 from= CAS-MCO-2016-11-26T22.32.39.582-BLU-03025-00.cub to= output.xml`
gives expected output.

gtests for topds4 all passed.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/appdata/export/TGOCaSSIS.tpl
+++ b/isis/appdata/export/TGOCaSSIS.tpl
@@ -1,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?xml-model href="http://pds.nasa.gov/pds4/disp/v1/PDS4_DISP_1700.sch" schemetypens="http://purl.oclc.org/dsdl/schematron"?>
+<?xml-model href="http://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1800.sch" schemetypens="http://purl.oclc.org/dsdl/schematron"?>
+<Product_Observational>
+ <CaSSIS_Header>
+  <IDENTIFICATION_DATA/>
+  <GEOMETRIC_DATA>
+   <PREDICTED_MAXIMUM_EXPOSURE_TIME unit="{{MainLabel.IsisCube.Archive.PredictMaximumExposureTime.Units}}">{{MainLabel.IsisCube.Archive.PredictMaximumExposureTime.Value}}</PREDICTED_MAXIMUM_EXPOSURE_TIME>
+   <CASSIS_OFF_NADIR_ANGLE unit="{{MainLabel.IsisCube.Archive.CassisOffNadirAngle.Units}}">{{MainLabel.IsisCube.Archive.CassisOffNadirAngle.Value}}</CASSIS_OFF_NADIR_ANGLE>
+   <PREDICTED_REQUIRED_REPETITION_FREQUENCY unit="{{MainLabel.IsisCube.Archive.PredictedRepetitionFrequency.Units}}">{{MainLabel.IsisCube.Archive.PredictedRepetitionFrequency.Value}}</PREDICTED_REQUIRED_REPETITION_FREQUENCY>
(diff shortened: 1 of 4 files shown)


isis/src/base/apps/topds4/topds4.cpp
```diff
@@ -197,3 +229,34 @@ namespace Isis {
     }
   }
 }
+
+
+QString PDS4PixelType(Isis::PixelType ipixelType, Isis::ByteOrder ibyteOrder) {
```
scsides, line 234: This is xml output dependent, so it will have to change when we adapt to other output types.
Why not put this in the template? Does the template logic not have the ability to do this?

isis/appdata/export/TGOCaSSIS.tpl
```diff
@@ -0,0 +1,116 @@
...
+  </Modification_History>
+ </Identification_Area>
+ <Observation_Area>
+  <Time_Coordinates>
+   <start_date_time>{{MainLabel.IsisCube.Instrument.StartTime.Value}}</start_date_time>
+   {% if exists("MainLabel.IsisCube.Instrument.StopTime") %}
+   <stop_date_time>{{MainLabel.IsisCube.Instrument.StopTime.Value}}Z</stop_date_time>
+   {% else %}
+   <stop_date_time>{{MainLabel.IsisCube.Instrument.StartTime.Value}}Z</stop_date_time>
+   {% endif %}
```
amystamile-usgs, line 44: I tried setting up my if/else statements in my template in the same way and it caused blank lines in between the variables. Not sure if this was the case for you but it may be worth noting in case it causes issues with reading the label.

scsides: Let's check with Tim. He placed a couple of inja environment lines in his most recent PR for topds4. When I looked them up they sounded like they may be removing new lines and something else. Checkout  #4479 topds4 line 175, 176

scsides: Also, blank lines shouldn't cause problems with the parsers, but a rendered template shouldn't have new-line characters in it that were not already there or did not come from the replacement string (i.e., json). Can we verify the new-lines are not in the json?

amystamile-usgs: So I confirmed that these new lines are being handled when processed to an isis label but I am not sure how they are handled when converting to an xml label.

isis/src/base/apps/topds4/topds4.cpp
tgiroux: This could be included in the template, but I think it's better suited as a callback because any template for exporting to xml will need to use this same chunk of logic.

isis/appdata/export/TGOCaSSIS.tpl
tgiroux: From the inja docs 

> While enabling set_trim_blocks removes the first newline after a statement, set_lstrip_blocks strips tabs and spaces from the beginning of a line to the start of a block.

This could also be handled by including a `-` in the syntax e.g. `{% else -%}` will clear whitespace around the tag according to the inja docs. Could be a better way to deal with this stuff rather than getting rid of leading/trailing whitespace wholesale.

isis/src/base/apps/topds4/topds4.cpp
scsides: let's go ahead and do it this way for now. We can change later if needed.

scsides (approved):

Issue #4213, opened 2020-12-18
Thm2isis not properly propagating cube attributes


acpaquette:
**ISIS version(s) affected**: 4.3.0 (and likely much older)

**Description**  
Thm2isis is not propagating cube attributes applied to the output cube. 

**How to reproduce**  
Take any Themis image, VIS or IR.

Run the following:
`Thm2isis from=/Path/to/input/Themis.img to=/Path/to/output/cube.cub+16bit`

The resulting output cube will have a Real bit type rather than a Signed Word bit type.

**Possible Solution**  
It's likely that the cube needs to have it's output attributes parsed from the direct string in the TO parameter of the UI.

jessemapel:
Closed by #4504

PR #4504, opened 2021-06-02
thm2isis now uses out attrs and converted to func

Kelvinrr:
## Description

thm2isis now applies output attributes and converted to a func. Still needs gtests.  

## Related Issue

#4213 

## Motivation and Context

Before this bug fix, thm2isis would no honor output cube attributes

## How Has This Been Tested?

original tests passing. 

## Types of changes

## Checklist:

## Licensing


isis/src/odyssey/apps/thm2isis/thm2isis.cpp
```diff
@@ -0,0 +1,268 @@
...
+      outputCubes.push_back(odd);
+      outputCubes.push_back(even);
+    }
+    else {
+      Cube *outCube = new Cube();
+      outCube->setDimensions(p.Samples(), p.Lines(), p.Bands());
+      
+      // outCube->setPixelType(Isis::Real);
+
+      outCube->create(outFile.expanded(), outAttr);
```
Kelvinrr: This is basically the bug fix (hard to find when you also do an app convert). Output attributes passed into `Cube::create`.

acpaquette (commented):
Overall this looks good! Did you try to replicate the test that I mentioned in the issue?

Kelvinrr:
@acpaquette yeah, I ran the images in the test folders with different pixel types locally and they all worked as expected.

Kelvinrr:
From writing the GTests for the output attributes. Something about making the pixel type to anything other than `Real` causes the pixel values to go to garbage. 


`hist` output from a Themis image ingested with +16bit using the original test image

```
Cube:              output/thm2isis-Truth3.even.cub
Band:              1
Average:           N/A
Std Deviation:     N/A
Variance:          N/A
Median:            N/A
Mode:              N/A
Skew:              N/A
Minimum:           N/A
Maximum:           N/A

Total Pixels:      589824
Valid Pixels:      0
Pixels Below Min:  0
Pixels Above Max:  0
Null Pixels:       401724
Lis Pixels:        0
Lrs Pixels:        0
His Pixels:        0
Hrs Pixels:        188100
MinInclusive,MaxExclusive,Pixels,CumulativePixels,Percent,CumulativePercent
```

Is this a bug in using output attributes with PDS ingestion apps? 


The crux of the problem is how the offsets and base are being initiated 

```
    Group = Pixels
      Type       = SignedWord
      ByteOrder  = Lsb
      Base       = -1.79769313486231e+308
      Multiplier = 0.0
    End_Group
```

I'll keep looking into it but throwing it out in case anyone has an idea of what might be happening

acpaquette:
@Kelvinrr Have you tried with a +8 or +32 and seeing what the output looks like? I agree that this may be an ingestion app problem but I figure the cube attributes "should just propogate" but I don't know

Kelvinrr:
+8 is also garbage, +32 is fine as +32 = Real. +8 or +16 without setting min and max as well through the output attributes basically gives 0 valid pixels.

acpaquette:
Maybe this wasn't a bug and is actually how the images need to be ingested? The only other thing I can think to try is to ingest some other PDS image, setting the output bit type and seeing what happens

isis/tests/FunctionalTestsThm2isis.cpp
```diff
@@ -0,0 +1,166 @@
...
+  EXPECT_DOUBLE_EQ(evenCube.multiplier(), 0.0039525691699605001);
+
+  std::unique_ptr<Histogram> hist (evenCube.histogram());
+   
+  EXPECT_NEAR(hist->Minimum(), 0.0, 0.0001);
+  EXPECT_NEAR(hist->Maximum(), 0.0, 0.0001); 
+  EXPECT_NEAR(hist->Average(), 0.0, 0.0001);
+  EXPECT_NEAR(hist->Sum(), 0.0, .00001);
+  EXPECT_EQ(hist->ValidPixels(), 188100);
+  EXPECT_NEAR(hist->StandardDeviation(), 0, .00001);
```
Kelvinrr: You can see it here, I set a min and max to avoid all invalid pixels and ended up with 0 pixels only.

isis/tests/FunctionalTestsThm2isis.cpp
```diff
@@ -0,0 +1,166 @@
...
+  EXPECT_NEAR(hist->Average(), 0.0012095900723426705, 0.0001);
+  EXPECT_NEAR(hist->Sum(), 457.42227419186383, .00001);
+  EXPECT_EQ(hist->ValidPixels(), 376200);
+  EXPECT_NEAR(hist->StandardDeviation(), 2.241887e-05, .00001);
+}
+
+
+TEST_F(TempTestingFiles, FunctionalTestThm2isisOutAttributes) {
+  // tempDir exists if the fixture subclasses TempTestingFiles, which most do
+  QString outCubeFileName = tempDir.path() + "/test.cub+msb+8bit+0.0012:0.0013";
```
Kelvinrr, line 203: So by picking a min max that are within the bounds of the original image seems to have fixed this issue? So maybe user error?

acpaquette: @Kelvinrr Maybe? Did you see my previous message about trying this with a different data set?

acpaquette: Also is there some way to determine what acceptable bounds are programmatically?

acpaquette:
Looks like the test image wasn't cropped potentially?

Kelvinrr:
@acpaquette Sorry, missed that comment on different data sets. I tried 3 different images so far (mix of vis and infrared) and they all output the same. 

I don't think there is a way to know min/max programmatically, I can look into the logic of how the output attributes are actually being used because I got a feeling it *should* be doing that and it's not. 

I didn't crop the one in there as it's already fairly small at ~5mb, but if that seems to big I can go back to seeing how to make it smaller.

isis/src/odyssey/apps/thm2isis/thm2isis.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "LineManager.h"
+#include "UserInterface.h"
+#include "CubeAttribute.h"
+#include "FileName.h"
+#include "IException.h"
+#include "iTime.h"
+#include "ProcessByBrick.h"
+#include "Brick.h"
+#include "OriginalLabel.h"
+
```
acpaquette, line 14: Doesn't this need an `#include "thm2isis.h"`?

isis/tests/FunctionalTestsThm2isis.cpp
```diff
@@ -0,0 +1,232 @@
...
+  EXPECT_NEAR(hist->Sum(), 0.023804268421372399, .00001);
+  EXPECT_EQ(hist->ValidPixels(), 50);
+  EXPECT_NEAR(hist->StandardDeviation(), 0.00011232993701816659, .00001);
+}
+
+
+TEST_F(TempTestingFiles, FunctionalTestThm2isisOutAttributes) {
+  // tempDir exists if the fixture subclasses TempTestingFiles, which most do
+  QString outCubeFileName = tempDir.path() + "/test.cub+msb+8bit+0.0012:0.0013";
+  QVector<QString> args = {"from=data/thm2isis/V00821003RDR.QUB",  "to="+outCubeFileName};
```
acpaquette: Could we crop this cube for consistance?

acpaquette (commented):
Overall looks good! A few things that could be addressed

acpaquette (approved):
Looks good!

PR #4509, opened 2021-06-04
Add tiling functionality to findfeatures via fastgeom

### LINKED ISSUE #4396: Findfeatures fails for LROC NAC images
**ISIS version(s) affected**: isis3.9.1, isis4.3.0, isis5.0.0-RC1

**Description**  
findfeatures fails for two lroc nac images with the following error:

```
./runFF.cmd 
terminate called recursively
terminate called recursively
terminate called recursively
./runFF.cmd: line 1: 68596 Aborted                 (core dumped) findfeatures algorithm=fast/brief/parameters@saverenderedimages:true@savepath:Rendered maxpoints=5000 maxthreads=7 match=M102128467LE.cub from=M102142784LE.cub fastgeom=true geomtype=camera epitolerance=3.0 ratio=0.9 hmgtolerance=3.0 filter=SOBEL onet=FF.net tolist=FF.lis pointid='M102128467LE_fast_?????' networkid=M102128467LE description='Create image-image control network' debug=true debuglog=FF.log
```
The above is for images without footprint information.  When images have footprintinit run on them, then findfeatures includes this additional line in the error message as shown above:
`terminate called after throwing an instance of 'cv::Exception'`

There is very little recorded in the output FF.log file and it never manages to render the images.

**How to reproduce**  
See /work/users/lweller/Isis3Tests/FindFeatures/LROCNAC/runFF.cmd for command execution.  Contains the same information contained in the error message above.

See proc.scr for the pre-processing which includes ingestion, spiceinit and footrpintinit (which I don't think is necessary but I included it because I was just trying everything to get findfeatures to work).  It fails on calibrated images as well, but  I don't think it is a necessity for findfeatures anyway, but I had that data available so I tried it.

I am currently running findfeatures successfully on Kaguya TC images as well as Titan and have not experienced this sort of failure.  I believe folks outside of Astro have successfully run findfeatures on LROC NAC images, which is partly why I went back to older versions. 

**Possible Solution**  

**Additional context**  
I have limited the number of keypoints because I have found when I don't and there is a lot of overlap and very interesting terrain the program fails/aborts.  It fails regardless.  It seems it is just not getting past step 1 to use any of this information.

Let me know if there is something else worth testing.

AustinSanders:
Adds tiling functionality to findfeatures via addition of GenericTiledTransform and edits to fastgeom
## Description
Replaces instances of GenericTransform with GenericTiledTransform within fastgeom.

## Related Issue
Closes #4396 

## Motivation and Context
Findfeatures fails with large images and fastgeom=true due to limits in cv::warp

## How Has This Been Tested?
Hand tested + added a test.  Disclaimer that the programmatic test is not ideal, and we should find a better way to test this.  The programmatically generated tests work ok with other findfeatures tests because they do not use any adjustments to the geometry, so they can be aligned fairly well to form "features."  Once adjustments take place, the images are essentially attempting to match noise to noise.

Using Lynn's images and parameterizations (which can be found in the original issue), the following network was generated with fastgeom=true:
Screen Shot 2021-06-04 at 4 34 12 PM
Screen Shot 2021-06-04 at 4 35 46 PM

## Types of changes

## Checklist:

## Licensing


isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -98,8 +98,8 @@ ImageTransform *FastGeom::compute(MatchImage &query, MatchImage &train)  {
     // same size as the output image below.
     if ( tSizeFull.area()  < (m_maxarea * qSize.area()) ) {
       // std::cout << "Use direct full mapping of Train-to-Query...\n";
-      fastg.reset(new GenericTransform("FastGeomMap", t_to_q,
-                                       tSizeFull));
+      fastg.reset(new GenericTiledTransform("FastGeomMap", t_to_q,
+                                       tSizeFull, 30000));
```
AustinSanders, line 102: Currently hard coded to tilesize 30000.  Should this be adjusted?  Configurable?

isis/tests/FunctionalTestsFindfeatures.cpp
```diff
@@ -248,3 +248,34 @@ TEST_F(ThreeImageNetwork, FunctionalTestFindfeaturesErrorNoMatch) {
     EXPECT_THAT(e.what(), HasSubstr("**USER ERROR** No control points found!"));
   }
 }
+
+TEST_F(ThreeImageNetwork, FunctionalTestFindfeaturesTiled) {
```
AustinSanders, line 252: This test is not ideal.  We need to find a way to construct an image other than "noise" so that features still match after geometric correction.

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
```diff
@@ -0,0 +1,175 @@
+/** This is free and unencumbered software released into the public domain.
```
AustinSanders, line 1: A lot of this code was ripped directly from https://gist.github.com/kinchungwong/141bfa2d996cb5ae8f42.  Should there be a "credit" comment?

AustinSanders (commented):
In general, I think this should be commented / tested better.  I ran short on time and wasn't able to write as cleanly as I'd like, but this should serve as a fairly solid first implementation.

isis/src/control/apps/findfeatures/FastGeom.cpp
jessemapel: The limit is based on OpenCV's usage of a 16 bit int, so I think hard coded is just fine.

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
jessemapel: Add a link to the gist on the class docs for GenericTilesTransform.

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
```diff
@@ -0,0 +1,180 @@
...
+  cv::Mat tmatrix = ImageTransform::translation(-subarea.x, -subarea.y);
+  setMatrix( tmatrix * matrix );
+  setSize( subarea.size() );
+}
+
+int GenericTiledTransform::getTileSize() const{
+  return m_tileSize;
+}
+
+GenericTiledTransform::RectArea GenericTiledTransform::transformedSize(const cv::Mat &tmat,
```
jessemapel, line 71: All of these functions need doc strings.

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
```diff
@@ -0,0 +1,180 @@
...
+std::tuple<int, int> GenericTiledTransform::getNumTiles(const cv::Size size) const{
+  int width = size.width;
+  int height = size.height;
+  int tileSize = getTileSize();
+  int nxTiles = (width + tileSize - 1) / tileSize;
+  int nyTiles = (height + tileSize - 1) / tileSize;
+  return {nxTiles, nyTiles};
+}
+
+std::pair<cv::Range, cv::Range> GenericTiledTransform::getTile(int tileID, cv::Size size) const{
```
jessemapel: Particularly here it would be helpful to document that tileID increases in the X direction first. So, a 2 by 4 would have these IDs:

```
0 1 2 3
4 5 6 7
```

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
```diff
@@ -0,0 +1,180 @@
...
+  int nyTiles = std::get<1>(xyTiles);
+  int n_tiles = nxTiles*nyTiles;
+  for (int i = 0; i < n_tiles; i++){
+    const std::pair<cv::Range, cv::Range> dstRangeXY = getTile(i, tFormSize);
+    const std::pair<cv::Range, cv::Range> srcRangeXY = addSourceInterpMargin(computeSourceRange(dstRangeXY), image, 3);
+    const cv::Range& dstRangeX = dstRangeXY.first;
+		const cv::Range& dstRangeY = dstRangeXY.second;
+		const cv::Range& srcRangeX = srcRangeXY.first;
+		const cv::Range& srcRangeY = srcRangeXY.second;
+		const cv::Mat_<cv::Vec2f> mapXY = computeMapping(srcRangeXY, dstRangeXY);
```
jessemapel: What's with the weird indents here?

isis/tests/FunctionalTestsFindfeatures.cpp
jessemapel: I agree, but this is a general problem with the findfeatures testing not being very helpful in the first place. All of the tests are effectively testing the OpenCV feature matching which has caused issues when we make small tweaks to things.

I think this is acceptable for now.

jessemapel:
An option for the test on this would be:

* set the number of points really high and the thresholds really low to get a lot of stuff
* check that there is at least one measure past the 32k line

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
AndrewAnnex: I think there is a copyright in the headers for those files, likely a problem for use in ISIS?

jessemapel: Yes, the copyright on the code is a problem. We can't just use the code and will have to write this from scratch :(

rbeyer: The Gist indicates that this was to get around a bug 8 years ago.  Has the bug not been fixed that we could benefit without having to rewrite?

Is it that an approach or technique in that code was mimicked?  That's different from cutting and pasting (which, 
sadly the "ripped directly" implies), so maybe you do need to clean-room reimplement this.  Bummer.

jessemapel: Unfortunately the bug was not fixed as it has to do with some fundamental choices that were made in OpenCV's remap function.

This PR completely copies a handful of functions from the gist, which I think would violate the copyright. It doesn't take all of the code, but a good chunk of it.

isis/src/control/apps/findfeatures/GenericTiledTransform.cpp
AustinSanders: The indent is only showing up for me on git. it looks perfectly fine in my editor.  I'm not sure how to fix it since I can't see it locally

jessemapel: I downloaded the file locally and these lines are indented with tab characters. Check your tab -> space conversion. ISIS only uses spaces in the source code.

AustinSanders:
I spent some time attempting to reimplement the algorithm, and the results appear consistent with the original implementation.  I'm not sure if this is "different enough," but it should be a step in the right direction.

Screen Shot 2021-06-07 at 9 09 39 PM

AndrewAnnex:
so, if you want to be truly sure that there is no copyright infringement, you may want to see if 1) more recent releases of opencv have the bug fixed/or comparable feature, and/or 2) have someone else who hasn't looked at the original copyrighted code re-implement it https://en.wikipedia.org/wiki/Clean_room_design. or 3) ask for permission from the author to re-license.

jessemapel:
The Issue is still open on OpenCV so it has not been fixed unforrtunately: https://github.com/opencv/opencv/issues/7544

I'll see if I can reach out to the author about copyright.

jessemapel:
Closing this as we'll need to take a different approach due to copyright.

PR #4532, opened 2021-06-16
soft pin on ffmpeg

Kelvinrr:
## Description

This soft pins ffmpeg to >= 3.4.1 as 3.4.0 has linking errors. 

Once this is merged i'll cherry-pick into 5.0.1 to keep the history clean. 

## Motivation and Context

Part of getting release going. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -121,7 +121,7 @@ requirements:
   - curl
   - eigen
   - embree
-  - ffmpeg>=3.4.1
+  - {{ pin_compatible('ffmpeg', min_pin='x.x') }}
   - {{ pin_compatible('geos', min_pin='x.x', max_pin='x.x') }}
   - {{ pin_compatible('geotiff', max_pin='x.x') }}
   - geotiff
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -68,6 +68,7 @@ requirements:
   - curl
   - eigen
   - embree
+  - ffmpeg>=3.4.1
   - geos>=3.7,<3.8
   - geotiff
   - gmm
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -68,6 +68,7 @@ requirements:
   - curl
   - eigen
   - embree
+  - ffmpeg>=3.4.1
```
jessemapel, line 71: Do we need to pin in the env file too?

Kelvinrr: only seems to be on issue on build on linux? I can make the add anyways.

recipe/meta.yaml
```diff
@@ -120,6 +121,7 @@ requirements:
...
```
acpaquette: It's possible that the change only needs to be made in the run section with a `pin_compatable` function. Did you upload the tarball for this already, and if not can you send me that linux tar ball?

Kelvinrr: not to the USGS channel because of this error, I wasn't 100% sure whether to add to one or both so I just did both to be safe. Not super knowledgable of the pin_compatible stuff, what would be the benefit here?

recipe/meta.yaml
jessemapel: Okay, seems fine to just have it here then.

recipe/meta.yaml
acpaquette: The main benefit is that it wouldn't pin ffmpeg in two places, we would only need to deal with a single pin. It's also possible that we are getting some version for the build and a different version in run, I don't really understand why conda does this but it happened in the past which is why there are `pin_compatable` definitions. There is a file you can look at in the build that defines what packages were obtained at the different steps from the build which will inform you whether we should use the `pin_compatable` vs a hard pin in both `build` and `run`

acpaquette: The file to check the versions obtained is under `info/recipe/meta.yaml`

Kelvinrr: made the change, I'm doing another linux build to test it.

Kelvinrr: ubuntu build cam back fine so I think this is good to go

acpaquette (approved):
👍

PR #4540, opened 2021-06-18
General clean-up for csmbundle

jessemapel:
General docs and todo clean-up. The biggest actual change is I tweaked how BundleSolutionInfo finds the index of each image. Before we had to do a linear indexOf call, but I realized that the observaitons store their own indices, so we can do that instead. We still have to do a linear pass through all previous observations to count the number of images, but this saves a little bit and is cleaner. We could do something like a look-up table for the sum of images at each observation, but the output code is not a bottle-neck so I think it's better to just do the simpler but slower approach.

related to #4410


isis/src/control/objs/BundleUtilities/IsisBundleObservation.cpp
```diff
@@ -313,15 +301,6 @@ namespace Isis {
...
    * @return @b bool Returns true upon successful intialization
-   *
-   * @internal
-   *   @todo Don't like this, don't like this, don't like this, don't like this, don't like this.
-   *         By the way, this seems klunky to me, would like to come up with a better way.
-   *         Also, apriori sigmas are in two places, the BundleObservationSolveSettings AND in the
-   *         the IsisBundleObservation class too - this is unnecessary should only be in the
-   *         BundleObservationSolveSettings. But, they are split into position and pointing.
-   *
-   *   @todo always returns true?
```
jessemapel, line 324: We actually need this now that we push all of the parameter logic into the observation sub-classes. So, this function now has a reason to exist!

AustinSanders (approved):

PR #4578, opened 2021-08-02
Pb Focal length per filter

amystamile-usgs:
## Description

implementation of focal length per filter.

## Motivation and Context

EIS Sprint

## How Has This Been Tested?

Existing tests passing

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
+++ b/isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
@@ -46,16 +46,10 @@ namespace Isis {
 
      Pvl &lab = *cube.label();
 
-     // try/catch is temporary until we have an iak with ck based frames with filter information
-     try {
-       PvlGroup &bandBin = lab.findGroup("BandBin", Pvl::Traverse);
-       QString key = "INS" + toString(naifIkCode()) + "_" + bandBin["FilterName"][0] + "_FOCAL_LENGTH";
-       SetFocalLength(Spice::getDouble(key));
-     }
-     // temporarily grabbing non filter focal length until we have iak with ck based frames with filter information
-     catch(...) {
-       SetFocalLength();
-     }
+     PvlGroup &bandBin = lab.findGroup("BandBin", Pvl::Traverse);
+     QString key = "INS" + toString(naifIkCode()) + "_" + bandBin["FilterName"][0] + "_FOCAL_LENGTH";
+     SetFocalLength(Spice::getDouble(key));
+
      SetPixelPitch();
 
      PvlGroup &inst = lab.findGroup("Instrument", Pvl::Traverse);
--- a/isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
+++ b/isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
@@ -46,12 +46,13 @@ namespace Isis {
 
      Pvl &lab = *cube.label();
 
+     // try/catch is temporary until we have an iak with ck based frames with filter information
      try {
        PvlGroup &bandBin = lab.findGroup("BandBin", Pvl::Traverse);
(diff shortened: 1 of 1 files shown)


isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
```diff
@@ -44,10 +44,19 @@ namespace Isis {
 
      NaifStatus::CheckErrors();
 
-     SetFocalLength();
+     Pvl &lab = *cube.label();
+
+     try {
```
AustinSanders: Not sure how much it matters, but I think try/except is more consistent with Python practice, and if/then would be more consistent with c++.   I think we could check for .hasGroup()

rbeyer: The usage of if/then vs try/except has *nothing* to do with language choice (unless your language doesn't have try/except, in which case you may want to choose a different language).  In fact, C++ was one of the first languages to implement it, before Python existed.  Sure, the if/then structure can be used in many cases to deal with exceptional or error circumstances, but generally in [those languages](https://en.wikipedia.org/wiki/Exception_handling_syntax) which support try/except the use of try/except is more efficient and cleaner to read in order to deal with those kinds of situations.

There has been lots of digital ink spilled and strong feelings exist about when and how to use if/then vs try/except, but at a very high level, if/then should be used for normal flow control management, and try/except should be used to handle anticipated error conditions or other exceptional circumstances.

I'm not deeply familiar with this code, but this seems like a very appropriate usage of try/except.

AustinSanders: Certainly c++ has try/except functionality, but in terms of language conventions it has been engrained in me that c++ favors a "look before you leap" approach (logical flow via conditional statements) whereas Python traditionally favors "easier to ask forgiveness than permission" (logical flow via exception handling). It isn't a matter of what _can_ be done, but the conventions of the language and its community.

I will concede that it may be easier to read, but I'm not sure how it could be more performant to instantiate an exception object than to _not_ instantiate an exception object.

"The idiom is that it should only be used for exceptional cases, not the normal flow of a program (unlike in Python, in which it is even used for iteration)."
https://softwareengineering.stackexchange.com/questions/182232/c-exceptions-vs-preliminary-error-condition-checks

AustinSanders (dismissed):

isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
AustinSanders: My comment mainly stems from the fact that an unparameterized SetFocalLength() is _currently_ the default functionality (though this is a data issue, not a programming issue), and default functionality feels weird in an except block since it constitutes normal flow.  Using exceptions to handle normal flow is more common in Python, and that's all I was getting at.

isis/src/clipper/objs/ClipperPushBroomCamera/ClipperPushBroomCamera.cpp
```diff
@@ -44,10 +44,14 @@ namespace Isis {
 
      NaifStatus::CheckErrors();
 
-     SetFocalLength();
+     Pvl &lab = *cube.label();
+
+     PvlGroup &bandBin = lab.findGroup("BandBin", Pvl::Traverse);
+     QString key = "INS" + toString(naifIkCode()) + "_" + bandBin["FilterName"][0] + "_FOCAL_LENGTH";
+     SetFocalLength(Spice::getDouble(key));
```
amystamile-usgs, line 51: Ended up not placing this SetFocalLength call in a try/catch because getDouble throws adequate and detailed error statements. For example if the key was not in the BandBin, getDouble outputs `The camera is requesting spice data [INS-159103_CLEAR_FOCAL_LENGTH] that was not attached, please re-run spiceinit." thrown in the test body.`

scsides (approved):

Issue #672, opened 2018-12-19
🐛 CISS2ISIS: Incorrect pixel value mapping


michaelaye:
Copied from https://isis.astrogeology.usgs.gov/fixit/issues/5485:

Possibly due to the use of an un-initialized variable in the ciss2isis application, all pixel values 255 in the 16 bit ISS images are mapped to be of "High Saturation" (and therefore end up set to -32764 in the int16 data cube).

This is most likely caused by the line number 87 in the ciss2isis.cpp where it seems that the variable dataConversionType is queried uninitialized, and validMax is then set to 255:

```C++
if(dataConversionType != "Table") {   //Conversion Type is 12Bit or 8LSB, only save off overclocked pixels
    validMax = 255;
```

I found this by checking the pixel values of the original PDS raw data at the locations where the cube array had the pixels marked as HighSat. Those values were all = 255.
This results in seemingly random pixels being marked as HighSat, in the middle of rather dark image regions, as can be seen in the attached screenshot. (attaching screenshots was broken in RedMine at the time).

Justin Hulet commented on this bug with:

> I can also attest to this bug. I followed a workaround advised in an older thread from 2014 that suggested deleltion of the "sample_type = sun_integer" line, with very scary results (see for yourself in the attachment). Another user posted a "safer" alternative to the above suggestion where one simply replaced "SUN_INTEGER" with "LSB_INTEGER"; the results are exactly the same as if you deleted the line (in the attached cub file, I used the line edit method rather than the deletion method.

lwellerastro:
watching this issue

jessemapel:
It doesn't look like the issue is the global variable. It's set in the helper function `TranslateCassIssLabels` that gets called on line 74.

@michaelaye Can you provide a file that reproduces the bug?

michaelaye:
A good example that shows this is the image [N1875232984](https://tools.pds-rings.seti.org/opus/#/primaryfilespec=N1875232984&view=browse&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=primaryfilespec,planet,target&widgets2=&detail=)

You can find the files here:

LBL: https://www.dropbox.com/s/r7vu209jfju4hdu/N1875232984_1.LBL?dl=1
IMG: https://www.dropbox.com/s/lqscmbr753ul24m/N1875232984_1.IMG?dl=1

[...]

PR #4589, opened 2021-08-09
ciss2isis - Set validmax global to input label value

tgiroux:
## Description

Set the value for validMax along with the other globals by reading from the input label. Previously was set according to faulty logic, leading some images to be ingested with an incorrect valid maximum DN. The helper function FixDns() replaces all DN's == validMax with Hrs. This was spoiling pixels when validMax was set incorrectly.

## Related Issue

#672 

## Motivation and Context

Isis Support

## Types of changes

## Checklist:

## Licensing


jessemapel:
How bad does the test conversion look on this? We should have a simple test for this case.

tgiroux:
> How bad does the test conversion look on this? We should have a simple test for this case.

Simple to convert, difficult to crop data. There's no existing script for cropping cassis edr and it's nontrivial for me to write one during this sprint.

Also - a test for this may not be valuable. This bug was the result of some faulty logic that has been removed, the same faulty logic would have to be rewritten for this to become an issue again in the future.

AustinSanders (dismissed):

acpaquette:
@jessemapel are we still waiting on a test here or can this be merged?

isis/src/cassini/apps/ciss2isis/ciss2isis.cpp
```diff
@@ -0,0 +1,449 @@
...
+    PvlKeyword opticsTemp = inst.findKeyword("OpticsTemperature");
+    inst.findKeyword("OpticsTemperature").setValue(opticsTemp[0]);
+    inst.findKeyword("OpticsTemperature").addValue(opticsTemp[1], "DegreesCelcius");
+
+    double instDataRate = inst.findKeyword("InstrumentDataRate");
+    inst.findKeyword("InstrumentDataRate").setValue(toString(instDataRate), "KilobitsPerSecond");
+
+    //  initialize global variables
+    dataConversionType = (QString) inst.findKeyword("DataConversionType");
+    validMax = inputLabel.findKeyword("ValidMaximum")[1].toInt();
```
acpaquette, line 297: Updated line, see the first commit on this PR

isis/tests/FunctionalTestsCiss2isis.cpp
```diff
@@ -0,0 +1,273 @@
...
+  EXPECT_EQ(outLabel->findObject("Table", Pvl::Traverse)["Name"][0], "ISS Prefix Pixels");
+
+  std::unique_ptr<Histogram> hist (outCube.histogram());
+  EXPECT_NEAR(hist->Average(), 70.914941406249994, .00001);
+  EXPECT_EQ(hist->Sum(), 726169);
+  EXPECT_EQ(hist->ValidPixels(), 10240);
+  EXPECT_NEAR(hist->StandardDeviation(), 0.84419124016427105, .0001);
+}
+
+TEST(Ciss2Isis, Ciss2isisCustomMax) {
```
acpaquette, line 232: Conversion to gtest necessary to accommodate this test

isis/tests/FunctionalTestsCiss2isis.cpp
```diff
@@ -0,0 +1,273 @@
...
+  ASSERT_EQ(pixels["Type"][0].toStdString(), "SignedWord");
+  ASSERT_EQ(pixels["ByteOrder"][0].toStdString(), "Lsb");
+  ASSERT_EQ((double)pixels["Base"], 0.0);
+  ASSERT_EQ((double)pixels["Multiplier"], 1.0);
+
+  std::unique_ptr<Histogram> hist (outCube.histogram());
+  EXPECT_NEAR(hist->Average(), 68.998626373626379, .00001);
+  EXPECT_EQ(hist->Sum(), 50231);
+  EXPECT_EQ(hist->ValidPixels(), 728);
+  EXPECT_NEAR(hist->StandardDeviation(), 0.037062465833055058, .0001);
```
jessemapel: I don't think you need to check the average, sum, and standard deviation. I'd instead just check that the maximum is <= 70. I'd also check the number of HRS pixels as that's what things over the valid max are set to.

jessemapel (commented):
Conversion and new tests look good, just one small change to more directly check what the test is actually doing.

isis/tests/FunctionalTestsCiss2isis.cpp
acpaquette: So something interesting about the import app is that the max is also considered to be a fully saturated pixel. So all valid pixels if max == 70 will be < 70. This was something I discussed with @scsides a week or two ago and we decided that it is better to retain current functionality at this point

isis/src/cassini/apps/ciss2isis/ciss2isis.cpp
```diff
@@ -0,0 +1,449 @@
...
+      }
+      else if(dataConversionType == "Table") {
+        buf[i] = stretch.Map((int)buf[i]);
+      }
+      // save max values (4095 for table-converted images and 255 for others) as HRS
+      if(buf[i] >= validMax) {
+        buf[i] = Hrs;
+      }
+    }
+  }
```
acpaquette, line 246: @jessemapel This function was modified on line 242 changing from `buf[i] == validMax` to `buf[i] >= validMax`. As I mentioned, @scsides and I discussed how valid it would be to actually have `buf[i] > validMax` which is what the variable validMax seems to imply. However, reading the comment on line 241 seems to imply differently

jessemapel: From the SIS, it seems like this value indicates the maximum possible DN. So, it would be ambiguous if the true amount of exposure for that pixel was right at that number or above it.

https://github.com/USGS-Astrogeology/ISIS3/issues/672#issuecomment-448812918

I think >= is correct here.

jessemapel (dismissed):

isis/src/cassini/apps/ciss2isis/ciss2isis.cpp
scsides: That's that way we saw it.

scsides (dismissed):

jessemapel (approved):

scsides (approved):

PR #4593, opened 2021-08-10
Revert "Initial data snooping test"

jessemapel:
Reverts USGS-Astrogeology/ISIS3#4560

This PR was unintentionally merged into dev instead of the snooping branch. Reverting for now as it will get squashed into the snooping branch PR.


ladoramkershner:
This reversion makes sense to me. How do reviews work for reversions?

ladoramkershner (approved):

jessemapel:
We just make sure that it is only reverting what it is supposed to and that the reason for reversion is correct.

PR #4597, opened 2021-08-16
Added USECAMSTATSTBL to caminfo

### LINKED ISSUE #3605: Extract Pre-Existing statistics from the CameraStatistics Table Object in caminfo
**Description**  

The caminfo application computes a rather large amount of useful geometric and observational data for a given image when CAMSTATS=TRUE. Much of this data is redundant with data computed by the camstats application which adds the data to the cube in a Table object called CameraStatistics. We added an option to read the data from the CameraStatistics table rather than recompute each time in caminfo. This can save considerable time as loading shape models to required to compute this data can take a lot of time.

A new parameter called USECAMSTATSTBL=TRUE was added to select this option in caminfo. caminfo changed some of the names of the columns so they are preserved, however all the camstats data is written to the output data file.

A typo in the AspectRatioMinimum was fixed and "Angle" was removed from some of the keyword labels to remain compatible with established results.

This feature will break backward compatibility if order and content of the results are required.

**Example**  

Here is an example using this option:

 ``` 
Object = Caminfo

  Object = Parameters
    Program     = caminfo
    From        = 20190328T182543S3900_pol_iofL2pan.cub
    Lines       = 1024
    Samples     = 1024
    Bands       = 1
  End_Object

  Object = Camstats
    MinimumLatitude                          = -89.252345
    MaximumLatitude                          = -1.039447
    MinimumLongitude                         = 57.422518
    MaximumLongitude                         = 339.012629
    MinimumResolution                        = 0.045611
    MaximumResolution                        = 0.050794
    MinimumPhase                             = 51.832420
    MaximumPhase                             = 52.950741
    MinimumEmission                          = 35.281018
    MaximumEmission                          = 137.845448
    MinimumIncidence                         = 77.315236
    MaximumIncidence                         = 147.598320
    LocalTimeMinimum                         = 8.165505
    LocalTimeMaximum                         = 21.661715
    LatitudeMinimum                          = -89.252345
    LatitudeMaximum                          = -1.039447
    LatitudeAverage                          = -79.697153
    LatitudeStandardDeviation                = 3.896915
    LongitudeMinimum                         = 57.422518
    LongitudeMaximum                         = 339.012629
    LongitudeAverage                         = 278.278843
    LongitudeStandardDeviation               = 24.130508
    SampleResolutionMinimum                  = 0.045611
    SampleResolutionMaximum                  = 0.050794
    SampleResolutionAverage                  = 0.045837
    SampleResolutionStandardDeviation        = 1.739509e-04
    LineResolutionMinimum                    = 0.045611
    LineResolutionMaximum                    = 0.050794
    LineResolutionAverage                    = 0.045837
    LineResolutionStandardDeviation          = 1.739509e-04
    ResolutionMinimum                        = 0.045611
    ResolutionMaximum                        = 0.050794
    ResolutionAverage                        = 0.045837
    ResolutionStandardDeviation              = 1.739509e-04
    ObliqueSampleResolutionMinimum           = 0.055875
    ObliqueSampleResolutionMaximum           = 0.072828
    ObliqueSampleResolutionAverage           = 0.062054
    ObliqueSampleResolutionStandardDeviation = 0.004073
    ObliqueLineResolutionMinimum             = 0.055875
    ObliqueLineResolutionMaximum             = 0.072828
    ObliqueLineResolutionAverage             = 0.062054
    ObliqueLineResolutionStandardDeviation   = 0.004073
    ObliqueResolutionMinimum                 = 0.055875
    ObliqueResolutionMaximum                 = 0.072828
    ObliqueResolutionAverage                 = 0.062054
    ObliqueResolutionStandardDeviation       = 0.004073
    AspectRatioMinimum                       = 1.0
    AspectRatioMaximum                       = 1.0
    AspectRatioAverage                       = 1.0
    AspectRatioStandardDeviation             = 0.0
    PhaseMinimum                             = 51.832420
    PhaseMaximum                             = 52.950741
    PhaseAverage                             = 52.395677
    PhaseStandardDeviation                   = 0.232526
    EmissionMinimum                          = 35.281018
    EmissionMaximum                          = 137.845448
    EmissionAverage                          = 42.023091
    EmissionStandardDeviation                = 4.273678
    IncidenceMinimum                         = 77.315236
    IncidenceMaximum                         = 147.598320
    IncidenceAverage                         = 84.876335
    IncidenceStandardDeviation               = 3.651786
    LocalSolarTimeMinimum                    = 8.165505
    LocalSolarTimeMaximum                    = 21.661715
    LocalSolarTimeAverage                    = 12.333387
    LocalSolarTimeStandardDeviation          = 1.593742
    LocalRadiusMinimum                       = 237.025506
    LocalRadiusMaximum                       = 251.915269
    LocalRadiusAverage                       = 241.741461
    LocalRadiusStandardDeviation             = 1.692599
    NorthAzimuthMinimum                      = 0.093394
    NorthAzimuthMaximum                      = 359.762841
    NorthAzimuthAverage                      = 132.052515
    NorthAzimuthStandardDeviation            = 81.114880
  End_Object
End_Object
End
```

### PR DESCRIPTION
**Pull request by amystamile-usgs**
## Description

Added the USECAMSTATSTBL option to caminfo. This allows caminfo to extract existing
camera statistics from the CameraStatistics Table of the input cube instead
of recalculating CameraStatistics.

~~Fixed AspectRationMaximum typo in CameraStatistics.~~
UPDATE: This change was revert since it is API breaking and needs to be implemented on the next major version tick of ISIS. 

## Related Issue

closes #3605 

## Motivation and Context

Support Sprint

## How Has This Been Tested?

Added test for new parameter. Updated tests that had the CameraStatistics typo.

Expecting osirisrex_module_test_mapcam to fail in Jenkins since it uses camstats. Will update truth data once merged. Need to check that there are not any other tests missed that current has the typo. 

## Types of changes

## Checklist:

## Licensing

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -35,13 +35,21 @@ release.
 
 ## [Unreleased]
 
+### Added
+- Added the ability to search filenames in measure's drop down boxes in Qnet Point Editor. [#4581](https://github.com/USGS-Astrogeology/ISIS3/issues/4581)
```
jessemapel: This should be in a separate PR

isis/src/base/apps/caminfo/caminfo.cpp
```diff
@@ -286,6 +286,67 @@ namespace Isis{
           camstats->append(MakePair("ObliqueResolutionMaximum", cg["ObliqueResolutionMaximum"][0]));
         }
 
+        // Extracts camstat data from existing CameraStatistics Table in cube label
+        if(ui.GetBoolean("USECAMSTATSTBL")) {
+
+          if (!incube->hasTable("CameraStatistics")) {
```
jessemapel: Add a test case for this

isis/src/base/apps/caminfo/caminfo.cpp
```diff
@@ -286,6 +286,67 @@ namespace Isis{
...
+            camstats->append(MakePair("LocalTimeMinimum", time.split(",").at(1)));
+            camstats->append(MakePair("LocalTimeMaximum", time.split(",").at(2)));
+          }
+
+          QString obl = TableRecord::toString(csTable[7]);
+          if(obl.split(",").at(0) == "ObliqueResolution") {
+            camstats->append(MakePair("ObliqueResolutionMinimum", obl.split(",").at(1)));
+            camstats->append(MakePair("ObliqueResolutionMaximum", obl.split(",").at(2)));
+          }
+        }
```
jessemapel, line 345: Should we handle a malformed table? What happens if one of these statistics is not present? What happens if the table has fewer than 13 records?

isis/src/base/apps/caminfo/caminfo.cpp
```diff
@@ -286,6 +286,67 @@ namespace Isis{
           camstats->append(MakePair("ObliqueResolutionMaximum", cg["ObliqueResolutionMaximum"][0]));
         }
 
+        // Extracts camstat data from existing CameraStatistics Table in cube label
+        if(ui.GetBoolean("USECAMSTATSTBL")) {
```
jessemapel: We generally don't want to both run camstats and use the camstats label in this. You should add a check before the "CAMSTATS" block checks the user didn't enable both. The exclude in the XML only works for GUI, not the command line.

jessemapel (commented):
I still need to compare against what @KrisBecker implemented, but this looks like it's doing the correct logic. Just need some more safe-guards and edge-case tests.

isis/src/base/apps/caminfo/caminfo.cpp
amystamile-usgs: This is a good point. I will start looking into a way to handle these situations.

amystamile-usgs: Looking into camstat and CameraStatistics::toPvl() I cannot find a way that they could create a malformed table or add fewer than 13 records. Is there somewhere else in ISIS that could create a CameraStatistics table that may cause these particular cases?

isis/src/base/apps/caminfo/caminfo.cpp
amystamile-usgs: I assumed it was handled in the command line as well since when trying to set both true, I get the error `**USER ERROR** Parameter [USECAMSTATSTBL] must NOT be used if parameter [CAMSTATS] equates to true.`

isis/src/base/apps/caminfo/caminfo.cpp
jessemapel: This would be in case CameraStatistics changes in the future. Since this code is dependent upon it, we want to make sure it fails gracefully if CameraStatistics breaks or changes.

isis/src/base/apps/caminfo/caminfo.cpp
jessemapel: I must have that incorrect then. You can ignore this.

isis/src/base/apps/caminfo/caminfo.cpp
KrisBecker: In this case, I simply ran camstats even if the user specified to use the table in the label and it doesn't exist. In other words, if the table doesn't exist, create it by running camstats (they asked for it) and use the same process to extract all the keywords in the table (after correcting the typo for AspectRatioMinimum). (BTW, isn't this API breaking since the name changed even tho its a typo?)

jessemapel: This falls into the grey area for API breaking. It changes the name of an output value, but we don't specify that as part of the API.

Personally, I think this should go out in a feature release with a note in the changelog and the app history.

isis/src/base/apps/caminfo/caminfo.cpp
KrisBecker: This behavior does not address the main motivation for adding this option. 

CAMSTATS and USECAMSTATSLBL are not mutually exclusive. 

Only consider USECAMSTATSLBL when the user would prefer to use the camstats table pre-calculated with specific parameters earlier in processing for efficiency and consistency reasons. This condition is achieved when CAMSTATS=TRUE and USECAMSTATSLBL=TRUE.

camstats is typically ran as part of standard base processing since it is needed by qmos and other apps.

The other use case is when users would rather specify their own values (LINC/SINC) regardless if a camstats run already exists in the table. In this case CAMSTATS=TRUE and USECAMSTATSLBL=FALSE (consistent with current behavior) and LINC/SINC are then used in this case.

USECAMSTATSLBL specifies a preference to use an existing table if it exists but if it doesn't, go ahead and calculate it with the parameters the user provides. An error thrown because the table doesn't exist would be very unfortunate/irritating. No error should occur if the table does not exist for any condition with this approach.

This change expedites efficiency and flexibility for camera statistics.

KrisBecker:
This implementation does not fully address the motivation for adding this option.

CAMSTATS and USECAMSTATSLBL are not mutually exclusive.

camstats is typically run as part of standard basic processing since it is required/used in other apps such as qmos.

USECAMSTATSLBL was added to _indicate_ a preference to use the table if it exists. If it doesn't exist, go ahead and run camstats to create it. In this case, CAMSTATS=TRUE and USECAMSTATSLBL=TRUE and LINC/SINC are applied if it needs to be generated.

For the case where the user does not want to use the table even if it exists, this can be achieved with CAMSTATS=TRUE and USECAMSTATSLBL=FALSE, and the LINC/SINC parameters are used to create the table. 

In either case, there should be no error thrown because the table does not exist. That would be unfortunate/irritating as it would require more processing logic at the user level. 

This option provides efficiency and consistency to handling of camera statistics in caminfo.

I would also like to ask why would we limit the keywords to only some of values produced by camstats? My implementation adds all the keywords found in the table. That seems to us to provide more value rather than restrict content.

amystamile-usgs:
@KrisBecker I appreciate the input. I originally went the route of making CAMSTAT and USECAMSTATSLBL exclusive was to accommodate the option of the user specifying CAMSTAT=FALSE and USECAMSTATSLBL=TRUE. I also had the thought that it may be more simplified to go this approach to avoid having the user enter LINC/SINC each time even if the they were aware that the table existed in the cube label. 

Correct me if I am wrong, but it sounds like you want USECAMSTATSLBL only be used in addition to CAMSTAT=TRUE. Therefore there would be no reason to account for only extracting the CameraStatistics from the cube's label. 

In regards to only using a limited number of keywords, I am not sure why that is the case. I was only trying to stay consistent with the original logic of caminfo. @jessemapel Do you happen to know the reasoning for this?

KrisBecker:
@amystamile-usgs the LINC/SINC=1 defaults are the same for both camstats and caminfo. That said, running camera statistics with those defaults can be rather time consuming/costly. And when you consider other values such as LINC=SINC=10 would only utilize ~10% of the data, users should determine how best to gather and use these stats.

I think of it this way - users decide they want camera statistics in caminfo, so either use the existing table trusting the parameters/reasoning used to create the data in the cube, which you likely made anyway, or choose your conditions that better suit your need at the time. I am trying to determine if I would make any other decision than to run camstats explicitly or rerun caminfo with LINC/SINC after I get an error because the table didn't already exist. I don't think so. I would be forced to run camstats in some form and choose LINC/SINC - same result either way **except caminfo doesn't (over)write the result to the cube**.  If the table doesn't exist, I still have to decide what LINC/SINC is acceptable. You can make that decision in a single run of caminfo without consequence this way.

It seems to me that trying to document behavior with mutual exclusivity of CAMSTATS/USECAMSTATSLBL is more complicated than the decision process described above.

And, I have actually found uses for the missing camstats data. We use caminfo as the main source to use for populating databases. This database is used to determine ranking/sorting criteria for image-based feature matching (to preserve specific reference measures) when merging networks and mosaic metric/stacking order. The script we developed determines this order from camstats data using the MESSENGER MDIS ranking/stacking metric formula (ref: [MDIS_CDR_RDRSIS.PDF](https://pdsimage2.wr.usgs.gov/data/mess-h-mdis-5-rdr-rtm-v1.0/MSGRMDS_8001/DOCUMENT/MDIS_CDR_RDRSIS.PDF), page 40). Literally any field in the database can also be used/substituted for this purpose. Some of those newly added values in camstats provide some quite nice flexibility.

jessemapel (commented):
This looks good to me now, but we should consider what @rbeyer suggested over on the ISIS_TC repo:

> I suggest a two-step: **Step1**: Add an "AspectRatioMaximum" keyword that does the same as the old mis-spelled one. This addition is just a bug-fix (or a feature--minor tick--if you really want, you can even cleanse the typo keyword from the docs, and otherwise just leave it as a "silent" keyword that would still work but isn't advertised). And put in a deprecation warning for "AspectRatioMaximun" that advertises to those that use it, that it is "deprecated" and scheduled for removal at the next major patch, and that they should start using the new keyword. **Step 2**: at the next major tick, remove it.

KrisBecker:
For comparison, here is the result of my version of caminfo with the camstats table used:

```
Object = Caminfo
  Object = Parameters
    Program     = caminfo
    IsisVersion = "3.6.1  | 2019-03-14"
    RunDate     = 2021-07-29T22:13:29
    IsisId      = Near/MSI/150008986806
    From        = m0150008986f2_2p_cif_gbl.cub
    Lines       = 412
    Samples     = 537
    Bands       = 1
  End_Object

  Object = Camstats
    MinimumLatitude                          = -68.233711676334
    MaximumLatitude                          = 4.0160235738765
    MinimumLongitude                         = 0.013561149209616
    MaximumLongitude                         = 359.98890143585
    MinimumResolution                        = 17.030060616802
    MaximumResolution                        = 18.942798466828
    MinimumPhase                             = 82.298660229033
    MaximumPhase                             = 84.689662294472
    MinimumEmission                          = 2.6985972397002
    MaximumEmission                          = 107.54067746363
    MinimumIncidence                         = 3.0635914828775
    MaximumIncidence                         = 127.71838594611
    LocalTimeMinimum                         = 5.7263550434744
    LocalTimeMaximum                         = 14.499307637386
    LatitudeMinimum                          = -68.233711676334
    LatitudeMaximum                          = 4.0160235738765
    LatitudeAverage                          = -15.575396249414
    LatitudeStandardDeviation                = 13.314912575406
    LongitudeMinimum                         = 0.013561149209616
    LongitudeMaximum                         = 359.98890143585
    LongitudeAverage                         = 203.68952141336
    LongitudeStandardDeviation               = 161.13168222333
    SampleResolutionMinimum                  = 17.030060616802
    SampleResolutionMaximum                  = 18.942798466828
    SampleResolutionAverage                  = 17.551581951385
    SampleResolutionStandardDeviation        = 0.51222827715059
    LineResolutionMinimum                    = 17.030060616802
    LineResolutionMaximum                    = 18.942798466828
    LineResolutionAverage                    = 17.551581951385
    LineResolutionStandardDeviation          = 0.51222827715059
    ResolutionMinimum                        = 17.030060616802
    ResolutionMaximum                        = 18.942798466828
    ResolutionAverage                        = 17.551581951385
    ResolutionStandardDeviation              = 0.51222827715059
    ObliqueSampleResolutionMinimum           = 17.066028407844
    ObliqueSampleResolutionMaximum           = 981.15939392832
    ObliqueSampleResolutionAverage           = 40.224584109636
    ObliqueSampleResolutionStandardDeviation = 38.562468216004
    ObliqueLineResolutionMinimum             = 17.066028407844
    ObliqueLineResolutionMaximum             = 981.15939392832
    ObliqueLineResolutionAverage             = 40.224584109636
    ObliqueLineResolutionStandardDeviation   = 38.562468216004
    ObliqueResolutionMinimum                 = 17.066028407844
    ObliqueResolutionMaximum                 = 981.15939392832
    ObliqueResolutionAverage                 = 40.224584109636
    ObliqueResolutionStandardDeviation       = 38.562468216004
    AspectRatioMinimum                       = 1.0
    AspectRatioMaximum                       = 1.0
    AspectRatioAverage                       = 1.0
    AspectRatioStandardDeviation             = 0.0
    PhaseMinimum                             = 82.298660229033
    PhaseMaximum                             = 84.689662294472
    PhaseAverage                             = 83.663352915309
    PhaseStandardDeviation                   = 0.55089657675936
    EmissionMinimum                          = 2.6985972397002
    EmissionMaximum                          = 107.54067746363
    EmissionAverage                          = 67.139537257319
    EmissionStandardDeviation                = 21.630590115123
    IncidenceMinimum                         = 3.0635914828775
    IncidenceMaximum                         = 127.71838594611
    IncidenceAverage                         = 52.566238432712
    IncidenceStandardDeviation               = 28.982539559613
    LocalSolarTimeMinimum                    = 5.7263550434744
    LocalSolarTimeMaximum                    = 14.499307637386
    LocalSolarTimeAverage                    = 8.0068826066945
    LocalSolarTimeStandardDeviation          = 2.2243833430609
    LocalRadiusMinimum                       = 5963.4980026726
    LocalRadiusMaximum                       = 15355.567932005
    LocalRadiusAverage                       = 11739.55167046
    LocalRadiusStandardDeviation             = 2724.7214371325
    NorthAzimuthMinimum                      = 31.84790275889
    NorthAzimuthMaximum                      = 155.54652508396
    NorthAzimuthAverage                      = 100.44046190397
    NorthAzimuthStandardDeviation            = 16.778859234733
  End_Object

```

isis/src/base/objs/CameraStatistics/CameraStatistics.truth
```diff
@@ -93,4 +93,4 @@ NorthAzimuth:
   NorthAzimuthMaximum = 360
   NorthAzimuthAverage = 356.025
   NorthAzimuthStandardDeviation = 28.4507
-
+
```
amystamile-usgs, line 96: CameraStatistics unit test was failing due to the last line being deleted automatically by my text editor. I was unable to add a line back since my text editor automatically removes this line after adding and therefore are no changes to commit. Tried added back the line by editing the file within Github but it seems to still be failing due to it automatically adding a space. I can't seem to fix this. Any options for removing these changes altogether on this pr?

jessemapel (dismissed):

isis/src/base/apps/caminfo/caminfo.xml
```diff
@@ -39,164 +39,212 @@
...
-    MinimumPhase              = 50.107674499885
-    MaximumPhase              = 60.748154060087
-    MinimumEmission           = 0.0076793870352429
-    MaximumEmission           = 11.494912087698
-    MinimumIncidence          = 52.486663651971
-    MaximumIncidence          = 58.235714475896
-    LocalTimeMinimum          = 9.7555337547863
-    LocalTimeMaximum          = 10.575853979758
-    ObliqueResolutionMinimum = 235.23505597394
-    ObliqueResolutionMaximum  = 243.18035398447
```
KrisBecker, line 67: These values should not be removed although they are redundant. This would be API breaking and removes preexisting keyword/values that existing pipelines almost assuradly rely on.

You can see in my output I left them in primarily for this reason.

isis/src/base/apps/caminfo/caminfo.xml
```diff
@@ -603,12 +658,26 @@ End
           </UL>
         </description>
         <inclusions>
+          <item>USECAMSTATSTBL</item>
           <item>SINC</item>
           <item>LINC</item>
         </inclusions>
         <default><item>FALSE</item></default>
```
KrisBecker, line 757: Good! This preserves preexisting behavior.

isis/src/base/apps/caminfo/caminfo.cpp
```diff
@@ -253,37 +277,15 @@ namespace Isis{
...
-          camstats->append(MakePair("MinimumResolution", cg["resolutionminimum"][0]));
-          camstats->append(MakePair("MaximumResolution", cg["resolutionmaximum"][0]));
-
-          cg = camPvl.findGroup("PhaseAngle", Pvl::Traverse);
-          camstats->append(MakePair("MinimumPhase", cg["phaseminimum"][0]));
-          camstats->append(MakePair("MaximumPhase", cg["phasemaximum"][0]));
-
-          cg = camPvl.findGroup("EmissionAngle", Pvl::Traverse);
-          camstats->append(MakePair("MinimumEmission", cg["emissionminimum"][0]));
-          camstats->append(MakePair("MaximumEmission", cg["emissionmaximum"][0]));
```
KrisBecker: I think these need to remain because removing them would cause breakage. These are the only keywords produced in the previous version. 

Tag for later removal like the typo?

isis/src/base/apps/caminfo/caminfo.cpp
```diff
@@ -253,37 +277,15 @@ namespace Isis{
...
-          camstats->append(MakePair("MinimumIncidence", cg["incidenceminimum"][0]));
-          camstats->append(MakePair("MaximumIncidence", cg["incidencemaximum"][0]));
-
-          cg = camPvl.findGroup("LocalSolarTime", Pvl::Traverse);
-          camstats->append(MakePair("LocalTimeMinimum", cg["localsolartimeMinimum"][0]));
-          camstats->append(MakePair("LocalTimeMaximum", cg["localsolartimeMaximum"][0]));
-
-          cg = camPvl.findGroup("ObliqueResolution", Pvl::Traverse);
-          camstats->append(MakePair("ObliqueResolutionMinimum", cg["ObliqueResolutionMinimum"][0]));
-          camstats->append(MakePair("ObliqueResolutionMaximum", cg["ObliqueResolutionMaximum"][0]));
```
KrisBecker: These need to remain as well.

They all should be tagged for removal in a future version as suggested earlier - or not.

KrisBecker (changes requested):
I think we need to fully/completely retain previous behavior. All the preexisting keywords caminfo produced in the original version are removed in this PR.  I don't think we can/should do that based on API breaking reasons expressed by others.

isis/src/base/apps/caminfo/caminfo.xml
KrisBecker, line 67: Not all keywords made it into the last post. There may be others.

isis/tests/FunctionalTestsCaminfo.cpp
```diff
@@ -3,122 +3,311 @@
...
+                    "Column [" << columnName.toStdString() << "] value [" <<
+                    stringCsvValue.toStdString() << "] does not match Pvl value [" <<
+                    stringPvlValue.toStdString() << "] within tolerance [0.001].";
+        }
+        else {
+            EXPECT_EQ(stringCsvValue.toStdString(), stringPvlValue.toStdString()) <<
+                    "Column [" << columnName.toStdString() << "] value [" <<
+                    stringCsvValue.toStdString() << "] does not match Pvl value [" <<
+                    stringPvlValue.toStdString() << "].";
+        }
```
jessemapel, line 117: I considered writing a new predicate formatter for this comparison case as it comes up a lot but I don't know if it'll be worth it. The predicate formatters are kind of clunky to use.

isis/tests/FunctionalTestsCaminfo.cpp
```diff
@@ -16,155 +16,298 @@ static QString APP_XML = FileName("$ISISROOT/bin/xml/caminfo.xml").expanded();
 
 
 TEST_F(DefaultCube, FunctionalTestCaminfoCsv) {
-    QString outFileName = tempDir.path() + "/outTemp.csv";
-    QVector<QString> args = {"from="+ testCube->fileName(),  "to="+outFileName,
+    // QString outFlatFileName = tempDir.path() + "/outTemp.csv";
+    QString outFlatFileName = "/work/users/jmapel/ISIS3/build/outTemp.csv";
```
amystamile-usgs: These changes need to be reverted at some point.

isis/tests/FunctionalTestsCaminfo.cpp
```diff
@@ -16,155 +16,298 @@ static QString APP_XML = FileName("$ISISROOT/bin/xml/caminfo.xml").expanded();
...
-       caminfo(options);
+       caminfo(flatOptions);
     }
     catch (IException &e) {
-        FAIL() << "Unable to open image: " << e.what() << std::endl;
+        FAIL() << "Failed to run caminfo with flat output: " << e.what() << std::endl;
+    }
+
+    // QString outPvlFileName = tempDir.path() + "/outTemp.pvl";
+    QString outPvlFileName = "/work/users/jmapel/ISIS3/build/outTemp.pvl";
```
amystamile-usgs: Same here.

isis/tests/FunctionalTestsCaminfo.cpp
```diff
@@ -16,155 +16,298 @@ static QString APP_XML = FileName("$ISISROOT/bin/xml/caminfo.xml").expanded();
...
+    // CSVReader::CSVAxis csvLine;
+    // CSVReader::CSVAxis csvColumn;
+    // // Access the "Header" information first
+
+    // csvColumn = header.getColumn();
+
+    // // Validate the header information is correct
+    // csvLine = header.getRow(1);
+    // EXPECT_PRED_FORMAT2(AssertQStringsEqual, csvLine[0], "caminfo");
+    // EXPECT_PRED_FORMAT2(AssertQStringsEqual, csvLine[2], "Viking1/VISB/33322515");
```
amystamile-usgs: Any reason to keep the old test commented out instead of removing it?

jessemapel: Nope, just some stuff I was using for debugging that I missed

amystamile-usgs (commented):
@jessemapel Looks good! I am unable to approve since this is still my PR.

jessemapel:
I still have some more test clean-up to do. Maybe @KrisBecker can do a re-review or I can get a fresh set of eyes from the dev team.

isis/src/base/apps/caminfo/caminfo.xml
```diff
@@ -39,164 +39,212 @@
...
-    ObliqueResolutionMinimum = 235.23505597394
-    ObliqueResolutionMaximum  = 243.18035398447
+    LatitudeMinimum                          = 9.9286479874788
+    LatitudeMaximum                          = 10.434709753119
+    LatitudeAverage                          = 10.181983206084
+    LatitudeStandardDeviation                = 0.11084102743244
+    LongitudeMinimum                         = 255.64554871862
+    LongitudeMaximum                         = 256.14606952525
+    LongitudeAverage                         = 255.89390491018
+    LongitudeStandardDeviation               = 0.10658330458136
```
acpaquette, line 75: If we were trying to avoid breaking backwards compatibility won't some of these being `(something)Minimum` rather than `Minimum(something)` break that?

KrisBecker: I would agree here - even if its in the documentation as such. The output PVL currently has these keywords at the very top of the Camstats Group, but it appears the new logic prints these out in the order they occur in the table. Should the existing/current order be retained as well?

KrisBecker: It looks as tho the Minimum/Maximum keywords are not included at all in the XML doc file.

jessemapel: Ah I missed updating the XML example I'll get that done. The code prints out the values as both new and old.

I would not take order as part of backward compatibility, both PVL and CSV with headers account for different orders by allowing access by keyword and column name respectively.

jessemapel:
This is ready to go now. I don't know why CI isn't displaying a green check but the Jenkins run passed.

AustinSanders (approved):

Issue #3600, opened 2019-12-13
Oblique Detector Resolution Should Use Local Emission Angle


KrisBecker:
**ISIS version(s) affected**: All

Note the OSIRIS-REx team has a custom version of ISIS we have been maintaining and have modified code to address issues we have encountered working with small irregular bodies.

**Description**  

The oblique detector resolution that is being calculated in the [Camera::ObliqueDetectorResolution()](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Camera/Camera.cpp#L598) class uses the emission angle computed from the ellipsoid model. We changed from using the ellipsoid model to instead apply the DEM/shape model so that the local emission angle can be used. 

Using the local emission angle provides a much better indication of the oblique detector resolution because if takes into account surface topography.

This also affects phocube oblique detector resolution backplane as well (in a non-backward compatible way, so a new option was added to compute the old oblique detector resolution) which I could also submit for changes. Also, as a side affect, any other computation involving the ObliqueDetectorResolution, such as ObliqueSampleResolution and ObliqueLineResolution, are also affected. 

**How to reproduce**  

Applications such as phocube, camstats, campt and caminfo can provide examples of this data value.

**Possible Solution**  

We can provide a PR/code that has this modification.

**Additional context**

jessemapel:
### Why this is important

The oblique detector resolution is suppose to be a more accurate measurement of the resolution. So, we should use the most accurate surface model available. For irregular bodies, there can be large variations in the surface normal, and thus the emission angle. In these situations, the ellipsoid based emission angle is insufficient.

ascbot:
I am a bot that cleans up old issues that do not have activity.

This issue has not received feedback in the last six months. I am going to add the `inactive` label to
this issue. If this is still a pertinent issue, please add a comment or add an emoji to an existing comment.

I will post again in five months with another reminder and will close this issue on it's birthday unless it has
some activity.

jessemapel:
This is still outstanding

PR #4600, opened 2021-08-16
ObliqueDetectorResolution uses local emission

tgiroux:
## Description

I believe these are the basic changes described by the issue. If I'm not mistaken, this should cause a number of tests to fail, so truth data will need to be updated. This PR could also use a test for the "new" ObliqueDetectorResolution vs the old.
Work in Progress.

## Related Issue

#3600 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ release.
 synchronized with footprintinit default values of the same parameters.
 This corrects inconsistencies of footprint generation failing in caminfo
 but passing in footprintinit. [#4651](https://github.com/USGS-Astrogeology/ISIS3/issues/4651).
+- Changed the internal logic of ObliqueDataResolution() to use LocalEmission angle rather than Emission angle.
+This will cause differences in output; new values will be more accurate because they use DEM rather than
+ellipsoid. The cost to compute the local emissoin will increase by approximately x1.5. [#3600](https://github.com/USGS-Astrogeology/ISIS3/issues/3600)
 
 ### Added
 - Added the USECAMSTATSTBL option to caminfo. This allows caminfo to extract existing
@@ -77,7 +80,6 @@ Keywords when running CAMSTATS.  [#3605](https://github.com/USGS-Astrogeology/IS
 - isisVarInit.py no longer writes a "cat" statement by default to the activate scripts which cause the ISIS version information to be written on conda activate.  This can be included in those scripts via a command line option to isisVarInit.py.  Also, a quiet option is provided to isisVarInit.py to suppress its own writing to standard out, if needed.
 - Changed how the images.csv file is output in jigsaw when there are multiple models. Now each sensor model will have its own images.csv file so that column headers can all be correct. For example, a solution involving LRONAC pairs and Apollo Metric images would generate three images.csv files: LRONAC Left, LRONAC Right, and Apollo Metric. [#4324](https://github.com/USGS-Astrogeology/ISIS3/issues/4324)
 - Changed the API of many Bundle utility classes as part of CSM support in jigsaw. [#4324](https://github.com/USGS-Astrogeology/ISIS3/issues/4324)
-- Changed the internal logic of ObliqueDataResolution() to use LocalEmission angle rather than Emission angle. This will cause differences in output; new values will be more accurate because they use DEM rather than ellipsoid.
 
 ## [5.0.2] - 2021-07-30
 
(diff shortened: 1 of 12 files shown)


isis/src/base/objs/Camera/Camera.cpp
```diff
@@ -579,23 +579,36 @@ namespace Isis {
    *
    * @return @b double
    */
-  double Camera::ObliqueDetectorResolution(){
+  double Camera::ObliqueDetectorResolution(bool useLocal) {
```
jessemapel, line 587: Add documentation for new useLocal parameter

jessemapel (commented):
Needs a specific test to check functionality.

isis/src/base/objs/Camera/Camera.cpp
```diff
@@ -579,23 +579,36 @@ namespace Isis {
...
+    if(useLocal) {
+      Angle phase, emission, incidence;
+      bool success;
 
-          }
-          return Isis::Null;
+      LocalPhotometricAngles(phase, incidence, emission, success);
+      emissionDeg = (success) ? emission.degrees() : EmissionAngle();
+    }
+    else {
```
KrisBecker: I would argue that you return Null when the photometric angles cannot be computed. Using the ellipsoid (EmissionAngle()) when local photometric angles fail creates inconsistent geometry. For example, when using local DTMs, there may be areas within an image that does not provide full DTM coverage and in those cases, Nulls are returned for local angles in apps like phocube. If you switch up and essentially ignore the newly added bool argument, you get inconsistent backplanes/data and may need to mask the oblique detector plane or risk misinterpreting results.

In fact, I am not sure under what conditions you would want angles off the ellipsoid when a DTM is available. However, whatever the reason, I see no way for users to specify which one should be used other than explicit shape model management using spiceinit (I added both options to phocube and formally compute the ellipsoid oblique angles for posterity).

tgiroux:
I spoke with jmapel and agreed that a phocube option for this should be in a future PR.

Note that the basic test case I've written has the same truth values for old(ellipsoid) and new(DEM) obliqueDetectorResolution. I was unable to find/create a cube with a measurable difference between Emission and LocalEmission angles at a given point.

jlaura:
@jessemapel What is the status on this PR?

jessemapel:
The outstanding issue with this PR is that the tests are returning the same value when using local and non-local. Tim couldn't find a test image where we saw any differences. So, it's unclear if this is even working correctly at the moment. The Logical all seems correct, but the gtest needs to be updated with potentially a new test case.

acpaquette:
[This](https://github.com/USGS-Astrogeology/ISIS3/blob/e31f4bef71df4a9fb4702045c08485e51a6d4ecb/isis/src/base/objs/ShapeModel/ShapeModel.cpp#L188) particular line is the culprit for the issues with the tests. Since the `GetLocalEmission` call in `LocalPhotometricAngles` computes a normal, the call to `EmissionAngle` doesn't compute a new emission angle but reuses the local emission computed from `LocalPhotometricAngles` in Camera.cpp

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ Keywords when running CAMSTATS.  [#3605](https://github.com/USGS-Astrogeology/IS
 - isisVarInit.py no longer writes a "cat" statement by default to the activate scripts which cause the ISIS version information to be written on conda activate.  This can be included in those scripts via a command line option to isisVarInit.py.  Also, a quiet option is provided to isisVarInit.py to suppress its own writing to standard out, if needed.
 - Changed how the images.csv file is output in jigsaw when there are multiple models. Now each sensor model will have its own images.csv file so that column headers can all be correct. For example, a solution involving LRONAC pairs and Apollo Metric images would generate three images.csv files: LRONAC Left, LRONAC Right, and Apollo Metric. [#4324](https://github.com/USGS-Astrogeology/ISIS3/issues/4324)
 - Changed the API of many Bundle utility classes as part of CSM support in jigsaw. [#4324](https://github.com/USGS-Astrogeology/ISIS3/issues/4324)
+- Changed the internal logic of ObliqueDataResolution() to use LocalEmission angle rather than Emission angle. This will cause differences in output; new values will be more accurate because they use DEM rather than ellipsoid.
```
jessemapel: We should add a note that this will increase computation time for oblique resolution calculations, but they will be much more accurate.

jessemapel (dismissed):

jessemapel (approved):

PR #4676, opened 2021-11-08
Fixed jigsaw residual reporting in bundleout.txt

### LINKED ISSUE #4655: Jigsaw - Residual Percentiles have changed
**ISIS version(s) affected**: 6.0.0

**Description**  
The Residual Percentiles and Residual Box Plot in the bundleout.txt file has dramatically changed from the last version of isis.

**How to reproduce**  
Compare all output from isis5.0.2 jigsaw to all output from isis6.0.0.  I don't think the type of solution matters as I saw this while solving for camera accelerations via Kaguya TC, but tested on a smaller Themis IR data set solving for camera and spacecraft with errorprop turned on (re-using old test data).  The only difference in the isis5.0.2 and isis6.0.0 output (including output network files) was the Residual Percentiles/Box Plot reported under isis6.0.0 are very much larger than what is reported in isis5.0.2 and don't correlate with what is reported in the residuals.csv file.

There is a small test network under /work/users/lweller/Isis3Tests/Jigsaw/Percentiles/.  See proc.scr to see command lines for both versions of isis as well as the diff tests run.  Note that the images in the input jigsaw fromlist reside in another directory.

Since it's small enough, here are the Residual Box Plot differences in the two runs (I've omitted the lengthier Percentiles output here):
```
diff isis5.0.2_bundleout.txt isis6.0.0_bundleout.txt
<                         minimum:   -1.204
<                      Quartile 1:   -0.110
<                          Median:   -0.001
<                      Quartile 3:   +0.103
<                         maximum:   +1.490
---
>                         minimum:  -17.194
>                      Quartile 1:   -1.576
>                          Median:   -0.010
>                      Quartile 3:   +1.469
>                         maximum:  +21.281
```

The maximum residual in residuals.csv is 1.8 so the isis5.0.2 output makes better sense.

Note that this output is not useful or even relevant to me - I'm simply reporting a bug.  I just happened to notice how large the values were when I was scrolling through a bundleout file having just seen my maximum residual was less than 2 (the min/max box plot values where >900 for my much larger Kaugya TC data set then the test data I provided and reported on above).

Although the test case above touches a few different parameters, I don't know if there are other differences in the isis versions other what's reported.  It's possible something else differs for say multi- sensor solutions.  I'm just suggesting I haven't exhausted the possibilities.

jessemapel:
## Description

Fixed how residuals are re-scaled when reporting to the bundelout.txt file.

## Related Issue

Fixes #4655 

## Motivation and Context

This gets the bundleout.txt file in line with the residuals csv.

## How Has This Been Tested?

Tested on an example network I have and added an automated check in a jigsaw gtest.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/control/objs/BundleUtilities/IsisBundleObservation.cpp
+++ b/isis/src/control/objs/BundleUtilities/IsisBundleObservation.cpp
@@ -1359,7 +1359,7 @@ QStringList IsisBundleObservation::parameterList() {
    * Converts the observed value from a focal plane coordinate to
    * an image sample or line.
    *
-   * @param measure measure The measure that the partials are
+   * @param measure The measure that the partials are
    *                being computed for.
    * @param deltaVal The difference between the measured and
    *                 calculated focal plane coordinate
@@ -1368,8 +1368,7 @@ QStringList IsisBundleObservation::parameterList() {
    *                calculated (line, sample) coordinate
    */
   double IsisBundleObservation::computeObservationValue(BundleMeasure &measure, double deltaVal) {
-    Camera *measureCamera = measure.camera();
-    return deltaVal / measureCamera->PixelPitch();
+    return deltaVal * 1.4;
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ release.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed a bug where the measure residuals reported in the bundleout.txt file were incorrect. [#4655](https://github.com/USGS-Astrogeology/ISIS3/issues/4655)
+
 ### Added
 - Added the ability to search filenames in measure's drop down boxes in Qnet Point Editor. [#4581](https://github.com/USGS-Astrogeology/ISIS3/issues/4581)
 - Added slope, local normal, and ellipsoid normal calculations to phocube. [#3635](https://github.com/USGS-Astrogeology/ISIS3/issues/3645)
--- a/isis/src/control/apps/jigsaw/jigsaw.xml
+++ b/isis/src/control/apps/jigsaw/jigsaw.xml
@@ -319,6 +319,17 @@
       rand notebook that jigsaw was based on. Also added a section in the documentation
       describing the target body radii solve issue.
     </change>
+    <change name="Jesse Mapel and Kristin Berry" date="2021-06-29">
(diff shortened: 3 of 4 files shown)


isis/src/control/objs/BundleUtilities/IsisBundleObservation.cpp
```diff
@@ -1368,8 +1368,7 @@ QStringList IsisBundleObservation::parameterList() {
    *                calculated (line, sample) coordinate
    */
   double IsisBundleObservation::computeObservationValue(BundleMeasure &measure, double deltaVal) {
-    Camera *measureCamera = measure.camera();
-    return deltaVal / measureCamera->PixelPitch();
+    return deltaVal * 1.4;
```
acpaquette, line 1371: Seems like a bit of a magic number here, could we get some kind of comment as to why this is 1.4?

acpaquette (commented):
Overall looks good, only other thing is that a change log entry could be added to IsisBundleObservation.h

isis/src/control/objs/BundleUtilities/IsisBundleObservation.cpp
KrisBecker: This looks suspiciously like std::sqrt(2.0), which is the distance from the center of a square pixel to any one of its corners. The returned value is in pixel units rather than physical size (mm?) of the detector pixel.

jessemapel: Kris is correct that it's a shortcut for sqrt(2). This is porting some old logic over so if 1.4 should be used is a bit beyond this PR. See my post on the issue:

https://github.com/USGS-Astrogeology/ISIS3/issues/4655#issuecomment-944361620

Because we don't actually need to be modeling the physical reality of anything, we just need a mathematically consistent system using 1.4 is totally fine as long as it's consistent. In an ideal world, this bit would be completely obsoleted by actual measure weighting but we haven't implemented a good way to do that yet; so, we're just waiting on a sensor by sensor basis.

jessemapel:
I usually do not add @history tags to docs that don't already have them. @acpaquette Is there a reason why this should get one?

acpaquette (approved):
@jessemapel Did not check if that file had a history tag in the label so my previous comment can be ignored

acpaquette (approved):

PR #4678, opened 2021-11-08
Qview Precision Update

### LINKED ISSUE #4603: qview tracking table does not correctly display 16-bit values with large Base/small Multiplier
**ISIS version(s) affected**: 5.0.2, also tested on 3.10.2

**Description**  
In the qview Track table, DN values for pixels with values above 1,000,000 are only reported to one decimal place, even if the underlying data has more detail through the use of an integer data type with base/multiplier values set. (I recognize that for 32-bit data, the current behavior is appropriate, as 8 digits is about the limit of a 32-bit float, but it's annoying for high-detail DTMs that have been converted to radii while trying to preserve as much detail as possible.)

Example case:
```
makecube l=500 s=500 b=1 to=t1.cub
fakecube from=t1.cub to=t2.cub op=grad
fx f1=t2.cub to=t3.cub+unsignedword+3:65522 eq="0.64*f1"
# The above command turns one step in the gradient into a 1 cm vertical shift, scaled such that 1m vertical = 64 DN
# There will be some quantization in the resulting image; this is expected, but it's much less than just adding 1737400 to a 32-bit cube.
# Now to fix up the scale/multiplier appropriately to make this "elevation" a radius from the center of the Moon
editlab from=t3.cub grpn=Pixels keyw=Multiplier val=0.015625
editlab from=t3.cub grpn=Pixels keyw=Base val=1737400

# Also, an example of the unpleasant quantization from just adding the radius:
fx f1=t2.cub to=t4.cub eq="f1/100 + 1737400"

qview t2.cub t3.cub t4.cub
```

When viewing file t3.cub in qview, it correctly shows the underlying data values visually (there isn't any incorrect quantization loading the file into qview), but in the Tracking table it only shows the data to one decimal place, e.g. 1737400.1, even though the actual data is changing in the second decimal place.

I actually ran into this trying to turn the LROC NAC Apollo 16 DTM into a radius source suitable for map-projection with minimal quantization, for which I used the following commands, which did in fact produce a much higher-quality version of the DTM than just adding 1737400:
```
fx f1=NAC_DTM_APOLLO16_E090S0160.cub to=t2.cub+unsignedword+3:65522 eq="64*(f1+200)"
editlab from=t2.cub grpn=Pixels keyw=Multiplier val=0.015625
editlab from=t2.cub grpn=Pixels keyw=Base val=1737200
```

Side-note: With the existence of very high-detail DTMs these days, it would probably also be good to be able to use elevation data instead of radius data in a DTM file to avoid needing these kind of shenanigans to avoid quantization. I'm also opening a suggestion ticket on that topic.

acpaquette:
Adds the ability to adjust precision to q apps that display DNs

## Description
Precision for some high quality elevation maps were not being reported with enough precision. Now double precision values will have 12 digits displayed rather than only 8.

## Related Issue
Closes #4603 

## Motivation and Context
High quality DEMs need to be able to display more than 8 digits from a double precision value in qview and other applications. This PR allows for precision to be addressed on a number per number basis and doesn't alter the current logic that ISIS already operates with.

## How Has This Been Tested?
Tested with existing ISIS tests

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/SpecialPixel/SpecialPixel.h
+++ b/isis/src/base/objs/SpecialPixel/SpecialPixel.h
@@ -368,7 +368,7 @@ namespace Isis {
    *
    * @return string The name of the pixel type
    */
-  inline QString PixelToString(double d) {
+  inline QString PixelToString(double d, double precision=8) {
     if(Isis::IsSpecial(d)) {
       if(Isis::IsNullPixel(d)) return "Null";
       if(Isis::IsLrsPixel(d)) return "Lrs";
@@ -379,7 +379,7 @@ namespace Isis {
     }
 
     QString result;
-    return result.setNum(d, 'g', 8);
+    return result.setNum(d, 'g', precision);
   }
 
 
--- a/isis/src/qisis/objs/AdvancedTrackTool/AdvancedTrackTool.cpp
+++ b/isis/src/qisis/objs/AdvancedTrackTool/AdvancedTrackTool.cpp
@@ -337,16 +337,14 @@ namespace Isis {
     }
 
     // Otherwise write out col 4 (Pixel value)
+    QString pixel;
     if(cvp->isGray()) {
-      QString grayPixel = PixelToString(cvp->grayPixel(isample, iline));
-      QString p = grayPixel;
-      p_tableWin->table()->item(row, getIndex("Pixel"))->setText(p);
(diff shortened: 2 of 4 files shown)


jessemapel (approved):
Looks good. Good catch on pulling this updating logic out.

PR #4682, opened 2021-11-09
Fixes Maptrim failure when mode=both

### LINKED ISSUE #4646: Maptrim fails on global mosaic for mode=both
**ISIS version(s) affected**: 5.0.2

**Description**  
When running on a mosaic using mode=both (trim and crop), I get the following error messages:
```
  Group = Error
    Program = maptrim
    Code    = 1
    Message = "Unable to initialize cube projection from file
               [TEMPORARYcroppedJ76294.cub]"
    File    = ProjectionFactory.cpp
    Line    = 1119
  End_Group

  Group = Error
    Program = maptrim
    Class   = "I/O ERROR"
    Code    = 4
    Message = "Unable to initialize Projection information from group
               [Mapping]"
    File    = ProjectionFactory.cpp
    Line    = 93
  End_Group

  Group = Error
    Program = maptrim
    Code    = 1
    Message = "Projection failed.  Invalid label group [Mapping]"
    File    = TProjection.cpp
    Line    = 242
  End_Group

  Group = Error
    Program = maptrim
    Code    = 1
    Message = "Projection failed. [MinimumLongitude,MaximumLongitude] of
               [359.99806836665,0.0019316333539336] are not properly ordered"
    File    = TProjection.cpp
    Line    = 205
  End_Group
```
Note that the temporary file produced by this run (TEMPORARYcroppedJ76294.cub) does indeed have the min/max longitude's  swapped. The input mosaic mapping information is fine.  I am maintaining the longitudes for my input mosaic and trimming away latitudes =/-57 poleward.

**How to reproduce**  
See data under /work/users/lweller/Isis3Tests/MapTrim/

`maptrim from=Europa_GalileoSSI_Voyager_2020_global_mosaic_500m_PosWest_Equi_unstretched.cub to=Both.cub mode=both minlat=-57 maxlat=57 minlon=0 maxlon=360`

The input mosaic covers latitudes -90 to 90 and longitudes 0 to 360.  
The temporary file is the problem.

I can run the program setting mode=trim successfully, but then there are NULLs from +/-57 to the poles which is not desired.  I worked around this by running map2map, but the trimming and cropping via maptrim is the better way to do what I want.

**Additional context**  

There is the possibility that this related to post #2430.  Not necessarily the same problem, but the developer notes (quite a bit too!) mention TProjection.cpp and ProjectionFactory.cpp repeatedly, so it might be worth at least being aware of some things noted at the time.  I would just search on that in the post to see if anything interesting pops up.  My original problem is too different from this issue to bother reading other aspects of the post.

Perhaps useful to know and part of the problem - my mosaic has LongitudeDomain=360 and LongitudeDirection=PositiveWest.  Just wondering if the latter is not being taken into consideration when the temporary file is being created.

amystamile-usgs:
## Description

Maptrim fails when image has a PositiveWest longitude direction. I found out that the error was happening when maptrim runs Crop. At the end of the Crop application, it reevaluates the label information based on the newly cropped image using updateLabel. Since the calculation error was within the Mapping Group of the pvl, this was not caught in the crop application. It was rather caught in Maptrim because the app makes a new proj based on the updated mapping information from crop. This fails due to longitude minimum of 360 being larger than max longitude of 0.

Within the Subarea::updateLabel function that updates the mapping area, the issue lies in the fact that Universal Longitude is being called in order to convert the longitude back to PositiveWest. The problem is, Universal Longitude checks if the longitude is PositiveWest and then converts the longitude as if the longitude is already set in PositiveWest. This happens before we actually set the longitude to PositiveWest. Therefore we are converting twice.

For instance, 

- minimum longitude is found at 360 and maximum is found at 0.

- Running Universal Longitude, checks that longitude is PositiveWest, converts minimum to 0 and max to 360.
- Then runs toPositiveWest using the universal longitude range of 0 and 360. This converts the longitude range back to minimum 360 and maximum 0.

To avoid this issue, simply calling the toPositiveEast and toPositiveWest with original found Longitudes will avoid accidentally converting twice. This is easily be done on accident since the conversion between the two is the same (make negative and set domain)

## Related Issue

Closes #4646 

## Motivation and Context

ISIS Support Sprint

## How Has This Been Tested?

Confirmed that crop tests and maptrim test were not affected. These tests were only testing PositiveEast scenarios.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ release.
 ### Deprecated
 - Deprecated edrget as discussed in [#3313](https://github.com/USGS-Astrogeology/ISIS3/issues/3313).
 
+### Fixed
+- Fixed Maptrim failures when mode=both for PositiveWest longitude direction. [#4646](https://github.com/USGS-Astrogeology/ISIS3/issues/4646)
+
 ## [6.0.0] - 2021-08-27
 
 ### Added
--- a/isis/src/base/objs/SubArea/SubArea.cpp
+++ b/isis/src/base/objs/SubArea/SubArea.cpp
@@ -191,7 +191,7 @@ namespace Isis {
               }
             }
             else {
-              if (proj->Has180Domain()) {
+              if (proj->Has360Domain()) {
                 minlon = proj->ToPositiveWest(proj->Longitude(), 360);
               }
               else {
--- a/isis/src/base/objs/SubArea/SubArea.cpp
+++ b/isis/src/base/objs/SubArea/SubArea.cpp
@@ -181,37 +181,42 @@ namespace Isis {
           proj->SetWorld(p_ss-.5, p_sl-.5);
           if (proj->IsGood()) {
             maxlat = proj->UniversalLatitude();
-            if (proj->IsPlanetographic()) {
-              maxlat = proj->ToPlanetographic(maxlat);
-            }
+
             if (proj->IsPositiveEast()) {
-              minlon = proj->UniversalLongitude();
-              if (proj->Has180Domain()) {
-                minlon = proj->To180Domain(minlon);
+              if (proj->Has360Domain()) {
+                minlon = proj->ToPositiveEast(proj->Longitude(), 360);
+              }
+              else {
+                minlon = proj->ToPositiveEast(proj->Longitude(), 180);
               }
             }
             else {
-              minlon = proj->ToPositiveWest(proj->UniversalLongitude(), 360);
               if (proj->Has180Domain()) {
-                minlon = proj->To180Domain(proj->ToPositiveWest(proj->UniversalLongitude(), 360));
+                minlon = proj->ToPositiveWest(proj->Longitude(), 360);
+              }
+              else {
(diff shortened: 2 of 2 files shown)


jessemapel:
Change looks good but the subarea test appears to be failing now.

amystamile-usgs:
Looks like all unit tests for SubArea are passing except for testing sub image area with linc=1.0 and sinc=1.0 for simple cylindrical projected cube. Which is failing with the following differences:

MinimumLongitude   = 134.39705921684
MaximumLongitude   = 132.94061974267

MinimumLongitude   = 225.60294078316
MaximumLongitude   = 227.05938025733

Currently looking into why that's the case.

amystamile-usgs:
In regards to the changes for PositiveEast, it was converting the initial Longitude that was set in PositiveEast to PositiveWest. This is because ToPositiveEast immediately converts the longitude to negative and updates the domain regardless of it's initial longitude direction. A better method for PositiveEast is using the UniversalLongitude function this is due to it first checking if it is PositiveWest before converting to the inverse.

jessemapel (approved):

PR #4722, opened 2021-12-27
Updated ckwriter offset calculations and added spkwriter calculations

amystamile-usgs:
## Description
In spiceinit, Kernels are matched to the cube based on the original labels start and end times.

This comparison checks if the kernel start time is less than the label start time and the kernel end time is greater than the label end time. Therefore checking that the label's start and end time lie between the kernel's start and end time.

`kernelStartTime <= labelStartTime  < labelEndTime <= kernelEndTime`

Therefore when checking the offsets, there is only a need to apply an offset if the label's start time is less than the camera start time and if the label's end time is greater than the kernel's end time. 

Also found out that these offsets are not the same for both start time and end time. Therefore this PR adds the end time offset to the comments as well.

```
Object = SpacecraftPosition
  RunTime = 2021-12-27T11:34:33

  Group = Dependencies
    LeapsecondKernel = $Base/kernels/lsk/naif0008.tls
  End_Group

  Group = Selection
    Time        = ("2002 FEB 20 22:59:01.701369 TDB",
                   "2002 FEB 20 22:59:09.296828 TDB")
    Instrument  = THEMIS_IR
    StartOffset = 0.26313
    EndOffset   = 171.871
    File        = /Users/astamile/testData/kernelTiming/thmIR.bsp
    Type        = Smithed
  End_Group
End_Object
End
```

These offsets are in absolute values therefore if there is a StartOffset, then this must be added to the label's original start time and if there is a EndOffset, then the offset must be subtracted from the label's original end time.

## Motivation and Context

Kernel Timing sprint

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/ckwriter/CkSpiceSegment.cpp
+++ b/isis/src/base/apps/ckwriter/CkSpiceSegment.cpp
@@ -212,7 +212,9 @@ void CkSpiceSegment::import(Cube &cube, const QString &tblname) {
      _camVersion = _kernels.CameraVersion();
 
     QString labStartTime = getKeyValue(*label, "StartTime");
+    QString labEndTime = getKeyValue(*label, "StopTime");
     iTime etLabStart(labStartTime);
+    iTime etLabEnd(labEndTime);
 
     //  Get the SPICE data
     Table ckCache = camera->instrumentRotation()->LineCache(tblname);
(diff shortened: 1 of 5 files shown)


isis/src/base/apps/ckwriter/CkSpiceSegment.cpp
```diff
@@ -265,9 +267,21 @@ void CkSpiceSegment::import(Cube &cube, const QString &tblname) {
     _utcStartTime = toUTC(startTime());
     _utcEndTime   = toUTC(endTime());
 
-    _timeOffset =  fabs(etLabStart.Et() - startTime());
+    _startOffset =  etLabStart.Et() - startTime();
+    _endOffset =  etLabEnd.Et() - endTime();
     // account for padding
-    if (_timeOffset <= 0.003) { _timeOffset = 0; }
+    if (_startOffset >= 0.0031) {
```
AustinSanders: If the padding is .003, should this just be > .003 instead of >=.0031?

amystamile-usgs: @AustinSanders  I was running into a situation in which the offset was 0.00305282 I guess due to the precision of doubles? Would a better solution be to set the precision?

AustinSanders: Hmm.  I'm not really sure.  With the current solution, what happens if the offset falls between .003 and .0031?

amystamile-usgs: Fixed this check by first rounding offsets to three decimal places before doing the padding check. It is necessary that these offsets are within three decimals anyways for when we add/subtract the label time with these offsets.

AustinSanders (approved):

PR #4728, opened 2021-12-29
Kerneldb changes for spice kernel timing

AustinSanders:
Adjusts kerneldb::match to incorporate timing adjustments.

## Related Issue
Spice kernel timing

## Motivation and Context
There are differences between the start/stop time in the label and the camera model.  This leads to a failure to pick up some kernels in spiceinit.

## How Has This Been Tested?
ctest -VV -R "Spiceinit" 9 tests succeed 0 tests fail.  May need additional tests.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/system/objs/KernelDb/KernelDb.cpp
+++ b/isis/src/system/objs/KernelDb/KernelDb.cpp
@@ -582,14 +582,6 @@ namespace Isis {
     bool matchTime = !grp.hasKeyword("Time");
     bool matchKeywords = true;
 
-    // If the kernel segment has an instrument specification that doesn't match
-    // the instrument id in the label, don't perform the rest of the checks
-    if (grp.hasKeyword("Instrument")){
-        const PvlGroup &inst = lab.findGroup("Instrument", Pvl::Traverse);
-        if (inst["InstrumentId"][0].toStdString() != grp["Instrument"][0].toStdString()){
-          return false;
-        };
-    }
 
     // First, the time search. Loop through the keywords, if the name isn't
     //  Time then skip it. If it is, then get the start/end times and keep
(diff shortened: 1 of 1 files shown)


isis/src/system/objs/KernelDb/KernelDb.cpp
```diff
@@ -581,20 +582,34 @@ namespace Isis {
...
         iTime kernelEnd   = (QString) key[1];
 
+        if (grp.hasKeyword("TimeOffset")){
+          offset = (double) grp["TimeOffset"];
+        }
+
         // If the kernel times inside of the requested times we
         // set the matchTime to be true.
-        if ((kernelStart <= timeToMatch) && (kernelEnd >= timeToMatch)) {
+        if ((kernelStart - offset <= timeToMatch) && (kernelEnd + offset >= timeToMatch)) {
```
AustinSanders: I'm not sure if this is the correct logic for the offset.  My understanding is that the offset is something like abs(label_start - camera_start), so we're just expanding the search window by that amount in each direction.

amystamile-usgs: This is not completely correct. This does not have my changes in which there is an offset for start time and an offset for end time. "TimeOffset" keyword no longer exists in dbs. It's now StartOffset and EndOffset. There also needs to be checks to see if both exist. It's possible for only StartTime offset to exist or just EndTime, or both.

isis/src/system/objs/KernelDb/KernelDb.cpp
```diff
@@ -581,20 +582,34 @@ namespace Isis {
...
 
+    // If the kernel segment has an instrument specification that doesn't match
+    // the instrument id in the label, don't perform the rest of the checks
+    if (grp.hasKeyword("Instrument")){
+        const PvlGroup &inst = lab.findGroup("Instrument", Pvl::Traverse);
+        if (inst["InstrumentId"][0].toStdString() != grp["Instrument"][0].toStdString()){
+          return false;
+        };
+    }
+
```
amystamile-usgs: I'm not sure we want to fall out of the checks completely if this is the case. Say we have db with both Themis VIS and Themis IR kernel selections. If one of the selection groups has an instrument then Themis VIS would fail the matching. I believe that instead it should try to match with the other selections that do not have an instrument listed before failing the matching completely.

amystamile-usgs: Since only selection groups with offsets have this instrument keyword, it might be better to do this check after trying to apply the offset? I am not sure.

amystamile-usgs (approved):

PR #4757, opened 2022-01-21
Updated Apollo tests to be faster

jessemapel:
## Description

* Updated the apolloremrx and apollocal tests to use a test image with reseaus instead of linear data.
* Updated them to only check specific regions of the output region instead of collecting the entire cube histogram.
* Added a test for apolloremrx's patch option

## Motivation and Context

These tests were taking quite a long time as they use a VERY large cube. By changing to only working on specific regions of the cube it saves a lot of processing time. Using data that mimics actual reseaus also improves tests coverage and better matches real use cases.

## How Has This Been Tested?

Tested locally and all tests passed. apollocal test time went down by ~50% and apolloremrx test times went down by ~25%.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/Fixtures.cpp
+++ b/isis/tests/Fixtures.cpp
@@ -1083,22 +1083,24 @@ namespace Isis {
 
 
   void ApolloCube::SetUp() {
+    start_time = std::chrono::steady_clock::now();
     TempTestingFiles::SetUp();
 
     testCube = new Cube();
     testCube->setDimensions(22900, 22900, 1);
     testCube->create(tempDir.path() + "/large.cub");
 
-    LineManager line(*testCube);
-    double pixelValue = 0.0;
-    for(line.begin(); !line.end(); line++) {
-      for(int i = 0; i < line.size(); i++) {
-        line[i] = pixelValue;
-      }
+    // LineManager line(*testCube);
+    // double pixelValue = 0.0;
+    // int stepSize = 100;
+    // for(line.begin(); !line.end(); line += stepSize) {
+    //   for(int i = 0; i < line.size(); i += stepSize) {
+    //     line[i] = pixelValue;
+    //   }
 
-      pixelValue++;
-      testCube->write(line);
-    }
+    //   pixelValue++;
+    //   testCube->write(line);
+    // }
 
     PvlGroup reseaus("Reseaus");
     PvlKeyword samples = PvlKeyword("Sample", "200");
--- a/isis/tests/Fixtures.cpp
+++ b/isis/tests/Fixtures.cpp
@@ -5,6 +5,7 @@
 #include "FileName.h"
 
 #include "Blob.h"
+#include "Brick.h"
 #include "csminit.h"
 #include "Fixtures.h"
 #include "Portal.h"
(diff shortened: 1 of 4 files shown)


isis/tests/FunctionalTestsApolloremrx.cpp
```diff
@@ -32,30 +33,73 @@ TEST_F(ApolloCube, FunctionalTestApolloremrxDefault) {
...
+}
+
+TEST_F(ApolloCube, FunctionalTestApolloremrxPatch) {
+
+  testCube->group("RESEAUS")["STATUS"] = "Refined";
+
+  QTemporaryDir prefix;
+  QString outCubeFileName = prefix.path()+"/outTEMP.cub";
+  QVector<QString> args = {"to=" + outCubeFileName,
+                           "action=PATCH"};
```
acpaquette, line 78: I'm guessing this test is patching out the reseauses?

isis/tests/FunctionalTestsApolloremrx.cpp
```diff
@@ -32,30 +33,73 @@ TEST_F(ApolloCube, FunctionalTestApolloremrxDefault) {
...
+    int baseSamp = (int)(reseaus[i].first+0.5) - (reseauSize/2);
+    int baseLine = (int)(reseaus[i].second+0.5) - (reseauSize/2);
+    brick.SetBasePosition(baseSamp,baseLine,1);
+    cube.read(brick);
+    Statistics reseauStats;
+    reseauStats.AddData(&brick[0], brick.size());
+
+    EXPECT_NEAR(reseauStats.Average(), i, 0.001) << "Reseau " << i;
+    EXPECT_EQ(reseauStats.ValidPixels(), 9604) << "Reseau " << i;
+    EXPECT_EQ(reseauStats.NullPixels(), 1005) << "Reseau " << i;
```
acpaquette, line 66: In this test, are the reseauses considered null pixels? If this is the case this seems like it's a little off? It looks like the reseauses are a 103 x 103 cross that is 5 pixels wide, in total that should be (103 + 103) * 5 = 1030. I'm not sure why we seem to be 25 pixels short here

isis/tests/FunctionalTestsApollocal.cpp
```diff
@@ -29,10 +30,18 @@ TEST_F(ApolloCube, FunctionalTestApollocalDefault) {
...
+  int baseLine = (int)(reseaus[0].second+0.5) - (reseauSize/2);
+  brick.SetBasePosition(baseSamp,baseLine,1);
+  oCube.read(brick);
+  Statistics reseauStats;
+  reseauStats.AddData(&brick[0], brick.size());
+
+  EXPECT_NEAR(reseauStats.Average(), -2864.497, 0.001);
+  EXPECT_NEAR(reseauStats.Sum(), -30389453.463, 0.001);
+  EXPECT_EQ(reseauStats.ValidPixels(), 10609);
+  EXPECT_EQ(reseauStats.NullPixels(), 2160);
```
acpaquette, line 45: Similar to my other comment on valid vs null pixels. These numbers seem a little off based on my assumptions of how the reseauses are treated in ISIS. Just want to clarify why this is

isis/tests/Fixtures.cpp
```diff
@@ -1089,39 +1090,54 @@ namespace Isis {
...
 
-    PvlKeyword lines = PvlKeyword("Line", "200");
-    lines += "400";
-    lines += "600";
+          // Horizontal line
+          brick[reseauSize * (reseauSize/2 + j) + i] = reseauValue;
+        }
+      }
+      testCube->write(brick);
+    }
```
acpaquette, line 1122: New fixture looks solid!

isis/tests/FunctionalTestsApollocal.cpp
```diff
@@ -1,8 +1,9 @@
+#include "Brick.h"
 #include "Fixtures.h"
 #include "Pvl.h"
 #include "PvlGroup.h"
 #include "TestUtilities.h"
-#include "Histogram.h"
+#include "Statistics.h"
```
acpaquette, line 6: Why the switch? Was this due to the change that we made to histogram vs statistics?

acpaquette (commented):
Overall, looks pretty good! A few clarifications for me around what ISIS does with reseauses (if I am spelling that right)

isis/tests/FunctionalTestsApolloremrx.cpp
jessemapel: Yeah there's two options, either patch out the reseaus with surrounding data or set them to null. The existing test sets them to null. This test patches them out.

isis/tests/FunctionalTestsApolloremrx.cpp
jessemapel: The 25 pixel difference is from the 5x5 overlap in the middle of the cross.

isis/tests/FunctionalTestsApollocal.cpp
jessemapel: The brick for this test is 10 pixels bigger than the Reseaus so that we can get a region that has null and non-null data. So, that's why we get the 103x103 reseau data but then a bunch of nulls, (103\*10) \* 2 + 10\*10 = 2160. Part of the functionality of apollocal is that it converts any special pixels in the input to Nulls, so I wanted to check that.

isis/tests/FunctionalTestsApollocal.cpp
jessemapel: The Cube function gets a Histogram of the whole cube. We don't need that much, so I'm just using a Statistics object and manually adding the data from the bricks to it.

acpaquette (approved):

PR #4767, opened 2022-02-02
Add ability for Cube class to handle input attributes. Closes #3805

### LINKED ISSUE #3805: Opening cubes outside of Process objects is error prone
**Description**  

When opening input cubes using a command line argument outside of a Process object, like we will be doing for many of the app to callable conversions, the intuitive `Cube::open(ui.GetFilename("FROM"))` results in cube attributes being ignored. Similarly, `Cube::create(ui.GetFilename("TO"))` results in command line cube attributes being ignored. This recently popped up in #3798 and is a particularly insidious error because most app tests don't use cube attributes.

Instead, the correct way to open or create cubes from command line arguments is to make a Cube object, and then make a CubeAttributeInput or CubeAttributeOutput and set the approriate attributes on the Cube object prior to calling `Cube::open` or `Cube::create`. Logic for this can be found in `Process::SetInputCube` and `Process::SetOutputCube`

We should have a simple routine that properly opens an input or output Cube with all of its attributes from a command line argument. This could mean modifying UserInterface or even the Cube opening and creating interface to use FileName objects instead of QStrings.

**Example**  

I would really like the following to properly open a Cube from a command line argument:

```
Cube iCube(ui.GetFilename("FROM"));
```

Something like this would also work:

```
Cube iCube = openInputCube(ui, "FROM")
```

scsides:
Have the Cube open member process input cube attributes.

## Description

Input cube attributes are now automatically processed by the Cube.open member. Previously, any cube that was not processed through the ProcessX classes required the input cube attributes (i.e., band specifications) to be added to the a cube before it was opened.

## Related Issue

#3805

## Motivation and Context

All cubes should honer cube attributes. Some applications did not after they were converted to callable functions. All future conversion would have required the same code to process input cube attributes. Now the only call necessary is ui.GetCubeName instead of ui.GetFileName

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/IsisAml/IsisAml.cpp
+++ b/isis/src/base/objs/IsisAml/IsisAml.cpp
@@ -678,6 +678,65 @@ void IsisAml::GetFileName(const QString &paramName,
 }
 
 
+
+
+
+
+
+
+
+/**
+ * Retrieves of a value for a parameter of type "cubename".
+ *
+ * @param paramName The partial or full name of the parameter to be retrieved.
+ * @param extension A default extension to add if it does not already exist on
+ * the file name.  For example, "txt" will make /mydir/myfile into
+ * /mydir/myfile.txt
+ *
+ * @return The value of the parameter.
+ */
+QString IsisAml::GetCubeName(const QString &paramName, QString extension) const {
+
+  const IsisParameterData *param = ReturnParam(paramName);
+
+  if (param->type != "cube") {
+    QString message = "Parameter [" + paramName + "] is not a cubename.";
+    throw Isis::IException(Isis::IException::Programmer, message, _FILEINFO_);
+  }
+
+  QString value;
+  if (param->values.size() == 0) {
+    if (param->defaultValues.size() == 0) {
+      QString message = "Parameter [" + paramName + "] has no value.";
+      throw Isis::IException(Isis::IException::User, message, _FILEINFO_);
+    }
+    else {
+      value = param->defaultValues[0];
+    }
+  }
+  else {
+    value = param->values[0];
+  }
+
+  Isis::FileName name(value);
+  if (extension != "") name = name.addExtension(extension);
+  value = name.expanded();
+  if (name.attributes().length() > 0) {
+    value += "+" + name.attributes();
+  }
+
+  return value;
+}
+
+
+
+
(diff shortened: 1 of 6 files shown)


isis/src/base/objs/IsisAml/IsisAml.cpp
```diff
@@ -678,6 +678,65 @@ void IsisAml::GetFileName(const QString &paramName,
...
+    else {
+      value = param->defaultValues[0];
+    }
+  }
+  else {
+    value = param->values[0];
+  }
+
+  Isis::FileName name(value);
+  if (extension != "") name = name.addExtension(extension);
```
jessemapel, line 746: Do we nee to worry about the user putting something like `to=blah+32-bit` and then this converting it to `to=blah+32-bit.cub`? Does the FileName class handle that for us?

isis/src/base/objs/IsisAml/IsisAml.cpp
```diff
@@ -678,6 +678,65 @@ void IsisAml::GetFileName(const QString &paramName,
...
+
+
+
+
+
+
+
+
+
+
```
jessemapel: Seems like a bunch of extra new lines

jessemapel (commented):
New function looks good, just needs a test and some minor comments.

isis/src/base/objs/IsisAml/IsisAml.cpp
scsides: Yes it does for cubes where it is expecting attributes, however, I'll check to see what happens if it is just a file name and there is a +

scsides: FYI, plus signs are forbidden inside cube names

isis/src/base/objs/IsisAml/IsisAml.cpp
scsides: Yes, that was me setting the code apart as I edited it. I'll remove them before un-drafting this

jessemapel:
I think this is good to go with a unittest on the new function

scsides:
Unit test has been updated along with new PutCube members

jessemapel (approved):

PR #4772, opened 2022-02-03
Fix findfeatures matrix inversion issues and improve FastGeom performance

### LINKED ISSUE #4639: Findfeatures - "Can't invert empty matrix" Errors
**ISIS version(s) affected**: 5.0.2

**Description**  
When running on a list of overlapping images, if findfeatures has a catastrophic failure with one image in the list, the entire process fails and no network is produced.  After running findfeatures on 5800+ images I have 25 images that failed with the same error:

    Program = findfeatures
    Class   = "PROGRAMMER ERROR"
    Code    = 3
    Message = "Can't invert empty matrix"
    File    = GenericTransform.cpp
    Line    = 139

There is nothing in the output log files that indicate which image(s) is causing the problem or what the problem actually is so it takes quite a bit of time and testing to determine what is going on.  I have found in the 2 cases I scrutinized that only one image in the input fromlist fails with this error when I run images through findfeatures one at time.  So one image in a long list of input images via fromlist is making the entire process fail.

I am running findfeatures on overlapping images which have overlap ratios >5%.  For the two cases I looked into closely the overlap ratio is just less than 6%.  For the two cases (which happen to overlap each other), the area of overlap is the top of one image and the bottom of the other.  The spice is decent and there is plenty of overlap as far as I'm concerned to place points - about 300 lines and across nearly all samples.

I have many images with similar or less overlaps that work just fine so I'm not clear on what the problem is.  Maybe those 5% overlaps are different in size and shape and somehow that's better?  I don't know because without going through an arduous process for every failure, I simply don't know which overlap in the input fromlist is causing the problem. Findfeatures shouldn't die because it is having an issue with one input image out of many.  Can't it just report on the offending image and move on?  

I have well over 173,000 images to process and now expect there will be many of these types of failures.  Some images may end up in other overlapping image networks because findfeatures has no issues with them, but if one other image in the fromlist is a problem, then no points for anyone. It's not clear to me how many image networks not being generated due to this problem will be ok in the end but it could have quite the impact depending on the total number of these types of all or nothing failures.

**How to reproduce**  
See files under /work/users/lweller/Isis3Tests/FindFeatures/EmptyMatrix/

The findfeatures command can be found in TC2S2B0_01_02980N041E0466_lq13_matcher_ff.cmd (can execute this file or copy and paste onto the command line) and it expects TC2S2B0_01_02980N041E0466_lq13_fromlist_ff.lis as input.

The output files from the failed run are
TC2S2B0_01_02980N041E0466_lq13_ff.log
print_TC2S2B0_01_02980N041E0466_lq13_ff.prt

The same command will successfully run if the image TC2W2B0_01_05147N056E0466 is removed from the input fromlist.

Command:
findfeatures algorithm=fast/brief maxthreads=7 match=/scratch/lweller/Kaguya_TC/Global/Morning/Lev0/TC2S2B0_01_02980N041E0466.lev0.cub fromlist=TC2S2B0_01_02980N041E0466_lq13_fromlist_ff.lis fastgeom=true geomtype=camera maxpoints=10000 epitolerance=3.0 ratio=0.9 hmgtolerance=3.0 filter=sobel networkid=TC2S2B0_01_02980N041E0466_lq13 pointid='TC2S2B0_01_02980N041E0466_lq13_?????' onet=TC2S2B0_01_02980N041E0466_lq13_ff.net tolist=TC2S2B0_01_02980N041E0466_lq13_cubes_ff.lis tonotmatched=TC2S2B0_01_02980N041E0466_lq13_notmatched_ff.lis description='Create image-image control network' debug=true debuglog=TC2S2B0_01_02980N041E0466_lq13_ff.log -log=print_TC2S2B0_01_02980N041E0466_lq13_ff.prt

**Additional context**  
Although the error message is different, I believe this is related to https://github.com/USGS-Astrogeology/ISIS3/issues/556

### LINKED ISSUE #4771: Improvements and bug fixes for findfeatures
findfeatures has FastGeom performance issues as well as a matrix inversion failure that will cause runtime failures

**Description**  

findfeature has several bugs that cause runtime failures.

Issues with findfeatures FastGeom may at times cause an invalid matrix which results in an attempt to invert an empty matrix (#4639). The error is trapped and causes the program to abort. This is typically caused by FastGeom correspondence mapping between MATCH and FROM/FROMLIST images. 

The FastGeom process was being deferred to the ImageSource class which impeded the ability to parameterize and constrain the grid-based algorithm. This would either exit early with insufficient corresponding valid latitude/longitude points between the images or it would run unbounded when correspondence wasn't possible. This resulted in ineffective matrix for warp/perspective transformations or matrices that could not be inverted.

**How to reproduce**  

It is difficult to reproduce as the application would abort without reporting the problem image. An exception is thrown with the message `Can't invert empty matrix.` or run for a very long time and then report `Failed to get geometry mapping for...` the images in ImageSource.

**Possible Solution**  

The matrix inversion error needs to be properly identified and properly handled while processing FROM/FROMLIST images. The FastGeom process needs to be improved.

A PR is forthcoming to address these issues.

**Additional context**

KrisBecker:
findfeatures matrix inversion issues are addressed in this PR.

## Description

Matrix inversion errors have been reported in #4639 that cause findfeatures to abort. This appears to be directly related to problems with the FastGeom procedure. Images that have very little to no overlapping common coverage in their FOVs are typically the source of problems in the ImageSource class.

This PR address the matrix inversion problem with better detection of an invalid matrix. Also, improvements to the FastGeom algorithm were required to detect and minimize the impact on successful use of lists of images, particularly. 

The FastGeom process was implemented in the ImageSource class that made corrections to the algorithm problematic because it lacks parameterization/control of the grid-based geometric mapping correspondence algorithm. Therefore, this algorithm, now referrred to as `Grid`, was reimplemented directly in the FastGeom class. This is an iterative algorithm to that will increase the density of the candidate geometry points in the MATCH image on each successive iteration and run unbounded until every pixel was compared and still failed. 

An additional FastGeom algorithm, called `Radial`, was implemented to compute radial patterns of corresponding geometry patterns with increasing point density on each circle. This algorithm is not iterative but can be configured to produce varying density and radial spacing outward from the center of the image.

As a consequence, these image failures are more quickly and efficiently identified and removed from the matcher list and optionally reported in a user specified file that contains a list of images that cannot be transformed or the matrix inversion fails. Hence, a new parameter to findfeatures was added called TONOGEOM that contains this list.

In addition, a known problem with construction of control points and measures is also addressed in this PR. At times, only the MATCH geometry control measure coordinate was set in the newly created control point. However, the corresponding geometry control measure in the FROM/FROMLIST images was never checked/validated. This caused downstream processing problems, particularly in jigsaw.

Lots of debugging information has been added to help in better understanding and track performance of FastGeom and creation of the output control network.

Here is the summary of changes contained in this PR:

* Fix matrix inversion error on empty matrix. (Fixes #4639)
* Identify images that fail FastGeom transform and exclude from matching.
* Add TONOGEOM parameter to write failed file loads or FastGeom error file list if the transform cannot be determined
* Improve FastGeom transform algorithm using new Radial point mapping scheme
* Add more debugging output to help diagnose problem images/procedures
* findfeatures modified to add Grid algorithm to FastGeom class
* Improved Grid algorithm by computing proper starting iteration to statisfy fastgeompoints request
* Added additional parameterization of Grid algorithm
* Reorganized mapping process to consolidate Radial and Grid algorithms in FastGeom
* Fixed fetching of target name in ImageSource that supports use of projected images/mosaics

## Related Issue

Fixes #4639.
Fixes #4771.

## Motivation and Context

findfeatures would abort on single images when applying FastGeom processing to computed a warp transform to better spatially compare the MATCH image to FROM/FROMLIST images. This allows use of OpenCV feature/detector algorithms that are not rotation and/or scale invariant.

Problems with a single image would abort findfeatures when they could simply be excluded from the matching process and continue processing images that are successfully processed.

## How Has This Been Tested?

These changes have been tested in for several months processing NEAR/MSI Eros and Rosetta/OSIRIS-NAC 67P images. Modifications to findfeatures affect only code used in the application and does not impact any other system classes or functionality.

## Types of changes

## Checklist:

## Licensing


isis/src/control/apps/findfeatures/FastGeom.h
```diff
@@ -75,6 +94,21 @@ class FastGeom {
...
+                       const double ymin, const double ymax)
+                       const {
+
+      // Test if in FOV consistent with ISIS image coordinates
+      if ( (point.x >= xmin ) && (point.x < xmax ) ) {
+        if ( ( point.y >= ymin ) && (point.y < ymax ) ) {
+          return (true);
+        }
+      }
+      return (false);
```
jessemapel: This can be reduced down to

```
return (point.x >= xmin ) && (point.x < xmax ) && ( point.y >= ymin ) && (point.y < ymax )
```

I assume the compiler will optimize this away, but it's much more readable than a double nested if that just returns a boolean.

isis/src/control/apps/findfeatures/GenericTransform.cpp
```diff
@@ -120,6 +122,35 @@ cv::Point2f GenericTransform::inverse(const cv::Point2f &point) const {
...
+ * @param matrix Square matrix to invert
+ * @param verify If true will throw an exception if not invertable
+ *
+ * @return cv::Mat Inverted matrix
+ */
+cv::Mat GenericTransform::computeInverse(const cv::Mat &matrix,
+                                         const bool verify) {
+  cv::Mat inverse;
+  double result = cv::invert(matrix, inverse);
+  if ( verify == true ) {
```
jessemapel: Just do `if (verify)` here

isis/src/control/apps/findfeatures/GenericTransform.cpp
```diff
@@ -120,6 +122,35 @@ cv::Point2f GenericTransform::inverse(const cv::Point2f &point) const {
...
+    if ( qFuzzyCompare(result+1.0, 0.0+1.0) ) {
```
jessemapel: Why do a fuzzy comparison? The OpenCV docs say that it will return exactly 0 if the matrix is singular.

https://docs.opencv.org/4.5.3/d2/de8/group__core__array.html#gad278044679d4ecf20f7622cc151aaaa2

isis/src/control/apps/findfeatures/GenericTransform.cpp
```diff
@@ -120,6 +122,35 @@ cv::Point2f GenericTransform::inverse(const cv::Point2f &point) const {
...
+      QString msg = "cv::Mat inverse matrix is not invertable";
```
jessemapel: Looks like there's an extra inverse and invertable should be invertible. Error message should be something like "Transformation matrix is not invertible"

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
  * @param train            Train image
  * @return ImageTransform* Pointer to FastGeom transform
  */
-ImageTransform *FastGeom::compute(MatchImage &query, MatchImage &train)  {
-  // std::cout << "\nQuery: " << query.source().name() << "\n";
-  // std::cout << "Train: " << train.source().name() << "\n";
+ImageTransform *FastGeom::compute(MatchImage &query, MatchImage &train,
+                                  QLogger logger) const {
+
+  // Shortcut...
```
jessemapel: What does this comment mean?

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+  double fg_q_line_tolerance = toDouble(m_parameters.get("FastGeomQueryLineTolerance", "0.0"));
+  double fg_sample_tolerance = toDouble(m_parameters.get("FastGeomSampleTolerance", "0.0"));
+  double fg_line_tolerance = toDouble(m_parameters.get("FastGeomLineTolerance", "0.0"));
+
+  // Train line/sample map restrictions
+  double q_min_samp = 0.5 - fg_q_sample_tolerance;
+  double q_max_samp = q_samps + 0.4999 + fg_q_sample_tolerance;
+  double q_min_line = 0.5 - fg_q_line_tolerance;
+  double q_max_line = q_lines + 0.4999 + fg_q_line_tolerance;
+  FGFov  q_fov(q_min_samp, q_min_line, (q_max_samp-q_min_samp+1.0), (q_max_line-q_min_line+1.0) );
```
jessemapel: Why add 1 to the FoV sizes?

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+  QString fg_algorithm = m_parameters.get("FastGeomAlgorithm", "Radial").toLower();
+  bool radial_method = ( "grid" != fg_algorithm );
+  logit <<   "  FastGeomAlgorithm:       " << fg_algorithm << "\n";
+  logit <<   "  FastGeomPoints:          " << m_fastpts << "\n";
+  logit <<   "  FastGeomTolerance:       " << m_tolerance << "\n";
+  logit <<   "  FastGeomSampleTolerance: " << fg_sample_tolerance << "\n";
+  logit <<   "  FastGeomLineTolerance:   " << fg_line_tolerance << "\n\n";
+
+  // Run the requested algorithm
+  if ( true == radial_method) {
```
jessemapel: no need to compare against true, just `if (radial_method)`

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    logit << "--> Using Radial Algorithm train-to-query mapping <--\n";
+
+    // Compute maximum radial track at the center of the image to
+    // the corner and scaling parameters
+    double fg_max_radius    = std::sqrt( (q_samps * q_samps) + ( q_lines * q_lines) );
```
jessemapel: This is the maximum diameter, not the maximum radius. Should be divided by 2 to get the distance from the center of the image to a corner.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    double fg_radius_factor = toDouble(m_parameters.get("FastGeomRadiusFactor", "0.035"));
+    double fg_point_factor  = toDouble(m_parameters.get("FastGeomPointFactor", "0.10"));
+
+    // No check for explict specs
+    // Allow user explicit radius delta increments - minimum is the length of the
+    // diagonal of 1 pixel ( 2 * sqrt(1^2 + 1^2))
```
jessemapel: This the diagonal length of 2 pixels. One pixel is just sqrt(2) diagonally.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+        double theta = r_angle * DEG2RAD;
+        double v_cos = std::cos(theta);
+        double v_sin = std::sin(theta);
+
+        // Rotation matrix relative to center of image. This is the
+        // reference vector, (v_x0, v_y0), rotated counter-clockwise,
+        // scaled by the radial pixel distance from the center of
+        // the image at (c_x, c_y).
+        double q_x = ( ((v_x0 * v_cos) + (v_y0 * -v_sin)) * fg_radius ) + c_x;
+        double q_y = ( ((v_x0 * v_sin) + (v_y0 *  v_cos)) * fg_radius ) + c_y;
```
jessemapel: You don't need a reference vector here as it cannot be parameterized, and it's already implicit in the sin and cos functions. These lines can just be

```
double q_x = c_x + (v_cos * fg_radius)
double q_y = c_y + (v_sin * fg_radius)
```

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+
+        // Check to see if the point is in the image FOV
+        FGPoint q_coord(q_x, q_y);
```
jessemapel: Don't bother defining q_x and q_y. Just define the point collectively. This function already has a lot of variable floating around. This will require 239 to use q_coord.x and q_coord.y.

isis/src/control/apps/findfeatures/FastGeom.cpp
jessemapel: This variable is entirely unused

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+  double q_max_samp = q_samps + 0.4999 + fg_q_sample_tolerance;
+  double q_min_line = 0.5 - fg_q_line_tolerance;
+  double q_max_line = q_lines + 0.4999 + fg_q_line_tolerance;
+  FGFov  q_fov(q_min_samp, q_min_line, (q_max_samp-q_min_samp+1.0), (q_max_line-q_min_line+1.0) );
+
+  double t_min_samp = 0.5 - fg_sample_tolerance;
+  double t_max_samp = t_samps + 0.4999 + fg_sample_tolerance;
+  double t_min_line = 0.5 - fg_line_tolerance;
+  double t_max_line = t_lines + 0.4999 + fg_line_tolerance;
+  FGFov  t_fov(t_min_samp, t_min_line, (t_max_samp-t_min_samp+1.0), (t_max_line-t_min_line+1.0) );
```
jessemapel: Same as q_fov, this is entirely unused

isis/src/control/apps/findfeatures/FastGeom.h
```diff
@@ -52,19 +53,37 @@ class ImageTransform;
...
+ *   @history 2022-02-02 Kris Becker Added Grid implementation resulting in
+ *                                     consolidation/abstraction of Radial
+ *                                     algorithm code
  */
 
 class FastGeom {
   public:
     typedef GenericTransform::RectArea   RectArea;
+    typedef cv::Point2d                  FGPoint;
+    typedef cv::Rect2d                   FGFov;
```
jessemapel, line 69: This typedef is only used for the unused FoV variables and should be removed.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+            // Test for those falling in train FOV
+            bool inFOV = in_fov(t_coord, t_min_samp, t_max_samp, t_min_line, t_max_line);
+            if ( inFOV ) {
+              n_train_fov++;
+
+              // Save the good points
+              q_infov_points.push_back(q_coord);
+              t_infov_points.push_back(t_coord);
+            }
+            t_inFOV.push_back(inFOV);
```
jessemapel: Put this before the if

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    double c_y = q_lines / 2.0;
+
+    // The Circle Loop - compute points along a circle at radial
+    // distance (pixels) from center
+    double points_per_radius = 0.0; // Gets the center point on 1st iteration
+    double fg_radius = 0.0;
+    while ( fg_radius <= fg_max_radius ) {
+
+      // Compute the number points / radial
+      double r_points = qMax(1.0, points_per_radius);
```
jessemapel: Why not just define `points_per_radius` as 1 above and then ditching this qMax call?

This could require an explicit check that "FastGeomPointIncrement" is non-negative. As of right now, this will just cause it to always use a single point if "FastGeomPointIncrement" is negative.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    if ( m_parameters.exists("FastGeomRadiusIncrement") ) {
+      fg_radius_inc = toDouble(m_parameters.get("FastGeomRadiusIncrement"));
+    }
+
+    // Allow user explicit point/radius increments. A 0 increment would only
+    // get a single point at all radii at the end of the reference vector.
+    // Probably bad, but we will allow it as an extreme.
+    double fg_point_inc = qMax( 0.0, (fg_radius_inc * fg_point_factor) );
+    if ( m_parameters.exists("FastGeomPointIncrement") ) {
+      fg_point_inc = toDouble(m_parameters.get("FastGeomPointIncrement"));
```
jessemapel: Why are both the factor and increments user settable? Is there a use case where you want to set the factor over the increment?

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+      points_per_radius += fg_point_inc;
+    }
+  }
+  else { // true == grid_method
+
+#if USE_OLD_ALGORITHM
+    mapper = train.source().getGeometryMapping(query.source(),
+                                               m_fastpts,
+                                               m_tolerance);
+#endif
```
jessemapel: Remove this as we don't set compiler flags. If users would reasonably want to use the old algorithm then it should be an option like grid vs ring.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+          }
+        }
+
+        // Update angle of rotation to the next point
+        r_angle += d_theta;
+      }
+
+      // Update radial distance from image center and points/circle
+      fg_radius         += fg_radius_inc;
+      points_per_radius += fg_point_inc;
```
jessemapel: What is the goal for the point distribution on the rings? Should points be equally spaced on the ring (equal arc length between then), equally spaced spatially (equal euclidean distance between them), something less regular?

With the pattern as implemented, you'll get a big variety of spatial point densities.

isis/src/control/apps/findfeatures/FastGeom.cpp
jessemapel: Would you want to floor this and make r_points an int so that the points are equally spread around the ring? With this implementation, you'll get "fractional" points in the lower right corner of the image.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+#if USE_OLD_ALGORITHM
+    mapper = train.source().getGeometryMapping(query.source(),
+                                               m_fastpts,
+                                               m_tolerance);
+#endif
+
+    logit << "--> Using Grid Algorithm train-to-query mapping <--\n";
+
+     // Compute increment
+    int   fg_minpts = qMax(m_fastpts, 16);
```
jessemapel: Odd spacing here

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    logit <<   "  FastGeomGridIterationStep:  " << fg_grid_iter_step << "\n";
+    logit <<   "  FastGeomGridSaveAllPoints:  " << toString(fg_save_all) << "\n";
+    logit <<   "  FGPointIncrement:           " << increment << "\n";
+
+
+    bool done = false;
+    int iteration = fg_grid_start_iter;
+    int currinc(0);
+    double v_q_nsamps = q_max_samp - q_min_samp + 1.0;
+    double v_q_nlines = q_max_line - q_min_line + 1.0;
```
jessemapel: Why +1 on these? q_min/max_samp/line do not appear to be inclusive based on how they are defined.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    int n_iterations(0);
+    while ( (iteration < fg_grid_stop_iter) && ( !done) ) {
+
+      currinc = increment + ( iteration * 2 );
+      n_iterations++;
+
+      double sSpacing = qMax( 1.0, v_q_nsamps/(currinc*1.0) );
+      double lSpacing = qMax( 1.0, v_q_nlines/(currinc*1.0) );
+
+      if ( qMax( sSpacing, lSpacing ) <= 1.0 ) done = true; // Last possible loop
```
jessemapel: The max here seems needlessly complex, just check if they are both 1.0

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+    mapper = train.source().getGeometryMapping(query.source(),
+                                               m_fastpts,
+                                               m_tolerance);
+#endif
+
+    logit << "--> Using Grid Algorithm train-to-query mapping <--\n";
+
+     // Compute increment
+    int   fg_minpts = qMax(m_fastpts, 16);
+    int increment = (int) (std::sqrt(std::max(24.0, (double) (fg_minpts - 1))) + 1.0);
```
jessemapel: Use a ceil call instead of subtracting and adding 1 before casting to an int. This will make it much easier to understand what this is trying to do. So, it would look like:

```
int increment = ceil(std::sqrt(std::max(24.0, fg_minpts)));
```

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -62,24 +67,398 @@ FastGeom::~FastGeom() { }
...
+        for ( int s = 0 ; s < currinc ; s++ ) {
+
+          // Total point count
+          n_total_points++;
+
+          double q_y = (lSpacing / 2.0 + lSpacing * l + 0.5) + q_min_line;
+          double q_x = (sSpacing / 2.0 + sSpacing * s + 0.5) + q_min_samp;
+
+          // Check to see if the point is in the image FOV
+          FGPoint q_coord(q_x, q_y);
```
jessemapel: Same as the ring algorithm, you may as well just define the point directly.

isis/src/control/apps/findfeatures/FastGeom.cpp
```diff
@@ -98,43 +477,98 @@ ImageTransform *FastGeom::compute(MatchImage &query, MatchImage &train)  {
...
  // If a mapper has not been allocated, allocate mapping to query image size
  // (as cam2cam does).
   if ( fastg.isNull() ) {
     // std::cout << "FastGeom cam2map map created...\n";
-    fastg.reset(new GenericTransform("FastGeomCamera", t_to_q, qSize.size()));
+    fastg.reset(new GenericTransform("FastGeomCamera", mapper, qSize.size()));
   }
 
   return ( fastg.take() );
 }
```
jessemapel, line 248: This is a really big function doing a lot of things and has a lot of inputs. It would be good to split out some of the algorithms so that they can be properly unit tested and we can ensure that we're properly checking and handling all of the parameterization.

isis/src/control/apps/findfeatures/findfeatures.cpp
```diff
@@ -88,12 +90,69 @@ namespace Isis{
...
+   * Errors are trapped here if the file cannot be loaded or the geometry proess
+   * in FastGeom fails. In this case, false is returned without adding the image
+   * to the matcher.
+   *
+   * @author 2021-10-29 Kris J. Becker
+   *
+   * @param matcher    The source reponsible for managing the matcher process
+   * @param trainfile  File name to load
+   * @param fastgeom   Apply FastGeom if provded. Errors will return false with
+   *                      no adding to the matcher
```
jessemapel: Looks like a typo, should be "return false without adding to the matcher"

isis/src/control/apps/findfeatures/findfeatures.cpp
```diff
@@ -88,12 +90,69 @@ namespace Isis{
...
+        // the query image.
+        fastgeom->apply(matcher.query(), t_image, QLogger(logger));
+      }
+
+      // If alls good...expection should be thrown above otherwise
+      matcher.addTrainImage( t_image );
+      isGood = true;
+    }
+    catch ( IException &ie ) {
+      isGood = false;
```
jessemapel: Why not just return false here and return true at the end?

isis/src/control/apps/findfeatures/findfeatures.cpp
```diff
@@ -327,15 +433,22 @@ namespace Isis{
       cnet.SetDescription(best->matcher()->name());
       cnet.SetCreatedDate(findfeatures_runtime);
       QString target = ( ui.WasEntered("TARGET") ) ? ui.GetString("TARGET") :
-                                                     best->target();
+                                                      best->target();
       cnet.SetTarget( target );
       ID pointId( ui.GetString("POINTID"), ui.GetInteger("POINTINDEX") );
-      matcher.setDebugOff();
+      // matcher.setDebugOff();  TURN ON DEBUG OUTPUT FOR NETWORK GENERATION!
```
jessemapel: Why is this here? If it needs to be on just delete this line

isis/src/control/apps/findfeatures/findfeatures.cpp
```diff
@@ -327,15 +433,22 @@ namespace Isis{
...
       PvlGroup cnetinfo = matcher.network(cnet, *best, pointId);
 
       if ( cnet.GetNumPoints() <= 0 ) {
         QString mess = "No control points found!!";
         logger->dbugout() << mess << "\n";
+
+        // Get total elapded time
```
jessemapel, line 482: typo, elapsed time

jessemapel (commented):
Overall looks good. I have a few questions about constant choices and what the desired matching patterns are. Most of the comments are just readability and clean-up.

The point distribution and matching algorithm is quite complex, it would be really helpful to separate it out so that it can be unit tested more easily.

isis/src/control/apps/findfeatures/FastGeom.h
KrisBecker: I completely removed this method in favor of OpenCV Rect::contains. It has the same comparison.

isis/src/control/apps/findfeatures/GenericTransform.cpp
KrisBecker: Done.

isis/src/control/apps/findfeatures/GenericTransform.cpp
KrisBecker: I actually looked at the [code](https://github.com/opencv/opencv/blob/8b4fa2605e1155bbef0d906bb1f272ec06d7796e/modules/core/src/lapack.cpp#L843-L1047) and it returns a boolean. I assumed it would be the determinant, hence the fuzzy compare. I changed it as you suggested.

isis/src/control/apps/findfeatures/GenericTransform.cpp
KrisBecker: Reworded the error message to actually make sense.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Its a shortcut (reference) to the actual stream operator. I expanded the comment to explain this.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Thats a bug, which is fixed. I retained the variable in favor of minimizing the number of variables and use of the OpenCV Rect type to test for points in the image FOV.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Done.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Thats a bug. Fixed.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: The comment is incorrect. I want to avoid testing the same pixel (corners). So I set it as indicated in the code and updated the comment.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Modified as suggested.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Ok...

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Corrected and retained the FGFov variable to test for points in the image.

isis/src/control/apps/findfeatures/FastGeom.h
KrisBecker: It is now used to test for points in the FOV so its retained.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Ok.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: This works as intended. On the first iteration you get exactly one point. On the next iteration you get fg_point_inc points and multiples of the increment on subsequent rings/circles. If fg_point_inc <= 0, you get a line of single points extending from the center of the image to the right. Values 0 < fg_point_inc < 1 might be interesting, but will still work. qMax is an important element of this algorithm.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Removed. This was for reference only.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Thats the point of this algorithm - to provide density when needed. Note the radial method does not loop and reconfigure the density as the grid algorithm does. You got one shot to get enough points.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: That is an implementation choice. This is something I can test but I think it would produce strange patterns due to other floating point elements of the algorithm.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Fixed/reworked.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Fixed.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Ok.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Done.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: I made separate methods for the grid, radial and dump portions of the compute() function and made them public scope so they can be directly called.

isis/src/control/apps/findfeatures/findfeatures.cpp
KrisBecker: Fixed.

isis/src/control/apps/findfeatures/findfeatures.cpp
KrisBecker: Ok

isis/src/control/apps/findfeatures/findfeatures.cpp
KrisBecker: This indicates a behavioral change. It was explicitly turned off in the commented code and I left it in the initial PR to indicate that change. I have since removed it.

isis/src/control/apps/findfeatures/findfeatures.cpp
KrisBecker: Fixed.

isis/src/control/apps/findfeatures/FastGeom.cpp
KrisBecker: Allowing both to be user settable provides for a lot of flexibility. This is useful for irregular bodies and approach observations to provide point density.

KrisBecker:
I am currently testing these modifications and will be ready to push them to the branch tomorrow unless you have additional comments/concerns.

KrisBecker:
Thanks @jessemapel for your very helpful feedback. Your comments compelled a deeper dive in testing and evaluation of the modifications. This identified additional issues and errors that produced a rather large refactor.

I pushed those updates into my branch that is ready for review. Clearly this app begs fo an overarching guide to help make it easier to use effectively and productively. Its something on my todo list for sure.

Please let me know if you have any other concerns.

jessemapel:
Code looks good. For testing, I think the best way to handle this is to have simple tests for findfeatures that ensure it's calling the correct algorithm between grid and ring. Then, the grid and ring algorithms can be unittested to ensure that all of the parameters are being passed in and used correctly.

jlaura:
@KrisBecker @jessemapel What is the status on this? Looks like it is waiting on tests?

jessemapel:
Correct this just needs tests and it's ready to go

rfergason:
This pull request is still active and should remain open.  Thanks!

KrisBecker:
Ok, I think these modifications address the remaining issues. Could someone please review this.

Thanks for your patience.

KrisBecker:
Pinging this PR for review/merge. 

Thanks...

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -1543,6 +1589,33 @@ End
...
+                <p>
+                    There is very little robust error detection or any confirmation
+                    that the parameter names are valid. Any misspelled parameters
+                    are not detected and ill-formed parameter strings may not
+                    result in errors, but passthrough and mangle other parameter
+                    construction.
+                </p>
+            </description>
+          <internalDefault>None</internalDefault>
+      </parameter>
```
acpaquette: This seems extremely specific to power users of find features. Without any kind of error checking or other mechanisms it seems like people could really shoot themselves in the foot here.

Plus what variables does this expose? If I dig into the findfeatures code I can find them but they aren't exposed in a way that would even indicate they are present. As a user on the surface I would have no idea how this worked or what it's for.

It seems like this is handled by the `Parameters` input in the application as per the tests, but from the documentation it would seem like the fastgeom parameters should not be part of the `Parameters` file.

I am sure there are a couple ways this could be addressed, but I think adding a `fastgeomparamters` file option if `FastGeom` is enabled would be nice. Either that or this `Global` parameter should only be for `FastGeom` parameters.

More than happy to iterate on a solution here

acpaquette (commented):
Overall tests look good and test the new FastGeom code. I am a little worried about the new `Global` parameter introduced. See comment below for details.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: I agree. 

This parameter is very useful for users mainly because it saves them from having to provide parameters in a file - this is very inconvenient when running in batch - as there is no other way to alter the parameters otherwise.

After a scan of the code and documentation, I see there really is no description how to use FASTGEOM. Note that I did [post](https://github.com/DOI-USGS/ISIS3/issues/4771#issuecomment-1029125719) a detailed analysis of these changes and how they impact the process (it will probably go away when [AstroDiscuss](https://astrodiscuss.usgs.gov) goes away. I can add much of this content to the findfeatures documentation, perhaps there is a way I can add this document somewhere in ISIS docs (convert to MD?).

I will also add a section that describes all the parameters in findfeatures that can be modified using this `GLOBAL` parameter. I can also add some templates for both Grid and Radial algorithms in appdata/findfeatures.

I will also add code to dump the full content of the global parameters used in findfeatures during the run in a separate group in the app log (print.prt). I see there is no report of the final parameters used. This will allow users to confirm the desired parameters are being applied.

If its the name of the parameter that is a concern, I chose this name because it does affect the global variable table beyond just the FASTGEOM algorithm. I figured the name `GLOBAL` is a good indicator of that scope. This parameters is required for efficient batch use. If you have a suggest for a more appropriate name/approach, we can certainly discuss that.

Let me know if these suggestions address your concerns or if you have some thoughts.

acpaquette: Maybe something to consider in the future is reworking findfeatures parameter tuning to be a little more consistent. It seems like one could make any change they want from the `GLOBAL` parameter and again end up doing unintended things by overriding variables in a passed in parameter file. For new users it's going to be overwhelming if they have to understand PVL parameter files as well as passing in variables from the `GLOBAL` parameter. At which point they would probably just pick whatever path gives them the least resistance.

Some potential in the future would be having different fields for different parameter sets, getting away from parameter files and moving more towards this key, value text parsing that the `GLOBAL` parameter is doing.

I think adding some context for `FASTGEOM` into the XML for the app somewhere would be ideal, basically context around the FASTGEOM parameters and describing the `GLOBAL` parameter. Seems like a straight forward place for that information to live. 

The other additions also sound useful and would get this closed out!

acpaquette:
@KrisBecker Do you have an idea when some of the above discussion will be added to this PR?

KrisBecker:
I am currently working on this along with other distractions. I am going to be on vacation later this week into next week so it will be at least two weeks away.

If you are planning a release soon, would it be possible to merge this PR and I submit another one when I have completed this work?

Thanks...

acpaquette:
@KrisBecker I think we can make that happen, I will open another issue to get docs on this once we merge this PR

KrisBecker:
@acpaquette that would be awesome!

Thank you!

AustinSanders:
@KrisBecker I'm in the process of peeling a release at the moment, and I'm in the final stages of merging PRs so that they can be included in 8.0.0 LTS.  I brought this PR up during our standup, and the team is somewhat on the fence about merging this without addressing some of the comments in the most recent review, particularly with respect to missing documentation.

That being said, I recognize that it's important to get this merged.  I can delay the release build until tomorrow morning, but any later than that would make me worried about the release timeline. If there's anything that I can do to help move this along before then, please let me know, and I'll do my best to help so that this can be included in the release-in-progress.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: The revisions I pushed hopefully resolves these issues/concerns.

Please review and let me know if there is anything remaining.

KrisBecker:
Thank you for your patience. 

I have updated the documentation and added some additional examples that addresses concerns. Tables have been added the describes all the parameters that can be applied to modify findfeatures behavior. The examples demonstrate use this feature. 

Please review and let me know if anything remains to get this PR merged.

Thanks for all your suggestions and help with this.

isis/appdata/templates/findfeatures/mosaics/00_basic_proc.sh
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash
```
acpaquette: Just to verify, is there a particular reason to include this chunk of bash scripts under templates? It seems strange to include them here rather that the `isis/scripts` area. Or potentially under some other section for tutorials. ISIS really hasn't shipped these kinds of pipelines in the past. It's something we could do but I am struggling to see a benefit for a new user.

Kelvinrr:
I think going forward we need to split these sort of PRs into smaller more manageable chunks. Large PR's almost guarantee this sort of slow progress in the merge process as it's more like merging a large feature branch. For now we still need to focus on wrapping this one up before or during the next support sprint.

KrisBecker:
I agree this one got much more difficult over time.

If you think it would help we can arrange for an interactive review. This might help it move along quicker toward merge.

I do see a few typos that I should fix. I may need to restructure the change log entry (its all over the place), so if you have any suggestions there that would be great.

Let me know how I can help get this merged.

Thanks...

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -3,431 +3,465 @@
...
+      matchers</a>. Some algorithms work well in a wide range of scenarios
+      (SURF and SIFT are both well tested and very robust), while others are
+      highly specialized. The following section provides guidelines to
+      determine which algorithms and parameters to use.
+    </p>
+    <p>
+      <i>findfeatures</i> gives users a wide range of options to alter
+      algorithms functionaility. Such a broad range of choices/combinations
+      can be intimidating for users who are unfamiliar with FBM. The following
+      are some sugestions to help make reasonable decision. First,
```
acpaquette: Typo, suggestions instead of sugestions

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -3,431 +3,465 @@
...
+      If at all possible, always use the FASTGEOM parameter. The
+      majority of problems when using FBM arise when trainer and query images
+      have inconsistent spatial geometry and/or significant variations in
+      observing conditions during image acquisition. The FASTGEOM parameter
+      helps minimize these challenges by using a priori ephemeris SPICE data
+      to create and apply a perspective matrix warp projection using the
+      query and training geometry. Combining the FASTGEOM option with
+      algorithms that are designed for speed (the FAST descriptor and BRIEF
+      extractor are reasonable options) will quickly produce a high quality
+      control network.
```
acpaquette: This seems like a very general and difficult to prove statement. To what standard is the control network high quality?

Maybe changing some wording to say, "...will quickly produce a control network. This process typically creates robust control networks but should be verified by the user"

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -3,431 +3,465 @@
...
+      the images. These transformations can help improve match quality in
+      specific scenarios. The <b>FASTGEOM</b>, <b>GEOMTYPE</b>, and
+      <b>FASTGEOMPOINTS</b> parameters allow for reprojection of the trainer
+      images into the query image's geometry prior to FBM. (This algorithm is
+      very similar to processing by the ISIS <i>cam2cam</i> application.
+      FASTGEOM differs in that it is not as robust or accurate as
+      <i>cam2cam</i> because FASTGEOM applies a single matrix to warp the
+      entire image - <i>cam2cam</i> uses a rubber sheet projection which is
+      significantly more robust and accurate but can take significantly
+      longer to project images.) These parameters can be used to achieve
```
acpaquette: The information on cam2cam here is decent but seems irrelevant. Largely because nothing in the cam2cam documentation mentions how it operates or that users have any control over that operation.

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -3,431 +3,465 @@
...
+      available to apply a more accurate filter. These filters allows for
+      improved detection when using edge based detectors (FAST, AGAST, BRISK,
+      etc...). If an edge based detector is not detecting a sufficient number
+      of key points or the key points are not sufficienty distinct, these
+      filters may increase the number of successful matches.
+    </p>
+    <p>
+      The OpenCV methods used in the outlier rejection process have several
+      options that can be set along with the algorithms. The available
+      parameters for are robust marcher algorithms are listed in the following
```
acpaquette: Typo, for are robust marcher -> for the robust matcher algorithms

Not really sure, but that seems to be logically consistent with the rest of the paragraph

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -730,7 +764,7 @@
           the query and trainer images are very similar, then the BRIEF
           algorithm offers unmatched speed. This speed comes at a cost, the
           BRIEF algorithm behaves very poorly if the query and trainer images
-          vary significantly. The <b>FASTGEOM</b> parameter can help improve
+          vary significantly. The FASTGEOM parameter can help improve
```
acpaquette: Should FASTGEOM be left  bolded here?

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -886,216 +920,816 @@
...
+      matching specification command line overrides the contents of
+      the file specified in the PARAMETERS program parameter. This allows
+      runtime specification of mainly the ALGOSPECFILE PVL-type file
+      contents to alter behavior of this component of the feature matching
+      algorithm.
+    </p>
+    <p>
+      Initial values of outlier matching parameters that can be specified in
+      the PARAMETERS program option are those found in the
+      <a href="#MatcherParameters">Robust Matcher Parameters</a> table.
```
acpaquette: Potential change from "...option are those found in the..." -> "...option can be found in the..."

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -886,216 +920,816 @@
...
+      These keywords are SaveRenderedImages, SavePath, RootSift,
+      MinimumFundamentalPoints, RefineFundamentalMatrix, and
+      MinimumHomographyPoints. Use of PARAMETERS excludes the need for these
```
acpaquette: If we link to the table why mention the keywords by name here?

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -886,216 +920,816 @@
...
+      ISIS inherently provides all the necessary cartographic capabilities
+      that make it possible to preprocess the images to eliminate, as much
+      as possible, scale and rotation variances in images. By using a priori
+      geometry provided by SPICE data, <i>findfeatures</i> constructs
+      transformation matrices for each trainer image that matches the
+      geometric properties of the query image. Applying the transformation
+      matrix to each trainer image results in a (fast) projection, or warp,
+      of the image into the image space of the query image, thus minimizing
+      scale and rotation invariance. This defines the FASTGEOM processing
+      objectives in <i>findfeatures</i>. Note, however, that translation,
```
acpaquette: Reads a little easier if ", however," is removed

isis/src/control/apps/findfeatures/findfeatures.xml
```diff
@@ -886,216 +920,816 @@
...
+      or spatial offsets, may likely exist between images due to the inherent
+      nature of a priori ephemeris data. And, for some image data sets,
+      particularly irregular bodies, this option may not work well,
+      especially for images acquired along the long axis of the target body.
```
acpaquette: The sentence starting with "And," is difficult to understand. Could this change to something like:

"This option is known to have problems with image sets of irregular bodies, especially for images acquired along the long axis of the target body."

acpaquette (commented):
Hi @KrisBecker, the documentation provided looks good! Couple of things to change here and there. I am not sure about the control network generation tutorial being housed under findfeatures template section as well as the tutorial being placed in the examples of the findfeatures app.

I think a better place to put the tutorial scripts is `isis/scripts/findfeatures_mosaic_tutorial/`, then we can add a new tutorial page under the ISIS wiki in the advanced section called something like "Making Mosaics Using findfeatures"

isis/appdata/templates/findfeatures/mosaics/00_basic_proc.sh
KrisBecker: There are likely several areas where these scripts could be located. I can see why some may argue this is a tutorial rather than a template. Here is my reasoning for including them in findfeatures documentation as an example and why they are in the templates directory for findfeatures:

1. Including this as an example in findfeatures provides context. A lot of effort has been put in this work and this example demonstrates how it is used to produce a small, regional mosaic comprised of nine MESSENGER MDIS images. It explains and demonstrates why many of the design decisions were made (e.g., parameterization, an issue you raised in a previous review).
2. It is a very explicit, small example using a specific dataset. The example explains, with visuals, why the MATCH image was selected and effective application of parameters to fully utilize findfeatures capabilities and features to analyze, evaluate and understand this application. It connects context where this information is provided - in the application documentation.
3. I think putting the scripts in the `isis/scripts` directory may give the false impression they are general/ready enough to use in other circumstance. They are not. For example, the photometric correction is specific to MDIS images. However, they do provide a starting point to gain experience with findfeatures. Users are encouraged to copy, modify to suite and experiment. Putting the scripts in the template directory associates/isolates it to this application as intended.
4. We intend to include, in PRs, at least three broad, comprehensive tutorials with our ISIS APEX work when it is submitted into ISIS. These tutorials will make use of findfeatures and provide a more general, comprehensive discussion/explanation of image processing techniques that can be applied using these features and capabilities.

I hope you have had a chance to see the rendered HTML documentation. I think it is a nice addition to the findfeatures documentation to explain its role in producing mosaics.

This example is provided to help users to better understand findfeatures in a broader context. It provides a baseline configuration (i.e, template) to adjust/modify parameterization that enables quick/efficient evaluations and comparisons of its behavior.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: Fair enough.

There are at least two ways to assess network quality - numerically and visually.

jigsaw produces a lot of bundleout data that I have used for numerical assessment. I look for a standard residual of no more than 1 pixel in measures/points. I also look for image RMS less that a pixel. There is more analysis that can be done, but these are the fundamentals.

Visual inspection of the controlled mosaic is the ultimate indication of quality.

That particular comment is based upon personal experience. 

I do agree its misleading to imply the output result of findfeatures is directly a high quality network. A quality network may exist in that network, but application of other ISIS tools are typically needed to achieve high quality networks.

I will modify the wording.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: For those who understand what cam2cam does, this reference provides context.

Its unfortunate cam2cam documentation does not better explain how it works.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: Perhaps. I tried to be consistent. Too much bolding can be distracting from the topic being discussed/emphasized.

I will review.

isis/src/control/apps/findfeatures/findfeatures.xml
KrisBecker: Thanks, I will revise.

isis/appdata/templates/findfeatures/mosaics/00_basic_proc.sh
acpaquette: Thanks, this helps me understand this a little better and gives context. While reviewing I was thinking that looking over this rendered may also help see broader scope. I will do that and report back.

If we plan to introduce more comprehensive tutorials would we want to see this info eventually pointed at those or would this remain along side?

KrisBecker:
I believe all the comments/concerns have been addressed.

Please let me know if there is anything left to do se we can get this PR merged.

Thanks...

CHANGELOG.md
```diff
@@ -35,11 +35,20 @@ release.
...
+- Added two new examples demonstrating/documenting the use of FASTGEOM algorithm, parameterization using <b>GLOBALS</b> and how to produce a regional mosaic using <i>findfeatures</i> with batch scripts. [#4772](https://github.com/DOI-USGS/ISIS3/issues/4772)
+- Added new option <b>GEOMSOURCE=BOTH</b> to check both the MATCH and FROM/FROMLIST images for valid control measure geometry to produce better networks and prevent downstream processing errors. Ignore points that end up with no valid measures (but can retained with use of <b>PreserveIgnoredControl</b> via GLOBALS parameterization). [#4772](https://github.com/DOI-USGS/ISIS3/issues/4772)
+- Added new gtests for <i>findfeatures</i> that replaces all the old application tests. These tests are <i>FunctionalTestFindfeaturesFastGeomDefault</i>, <i>FunctionalTestFindfeaturesFastGeomRadialConfig</i>, <i>FunctionalTestFindfeaturesFastGeomGridDefault</i> and <i>FunctionalTestFindfeaturesFastGeomGridConfig</i>. [#4772](https://github.com/DOI-USGS/ISIS3/issues/4772)
```
Kelvinrr: As per our discussion at the TC meeting. The fact that his required multiple changelog lines I believe indicates that this should have been split up into multiple PRs.

CHANGELOG.md
```diff
@@ -48,6 +57,10 @@ release.
 ### Fixed
 - Updated History constructor to check for invalid BLOB before copying History BLOB to output cube [#4966](https://github.com/DOI-USGS/ISIS3/issues/4966)
 - Updated photomet MinnaertEmpirical model to support photemplate-style PVL format [#3621](https://github.com/DOI-USGS/ISIS3/issues/3621)
+- Fix matrix inversion errors in findfeatures due to bad FASTGEOM matrix transforms using a more robust implementation to detect these errors and throw exceptions. Images with these errors are captured and logged to the <b>TONOTMATCHED</b> file. Fixes [#4639](https://github.com/DOI-USGS/ISIS3/issues/4639)
+- Fixed use of projected mosaics with correct check for <b>TargetName</b> in the Mapping labels. [#4772](https://github.com/DOI-USGS/ISIS3/issues/4772)
+- Fixed a bug in the <i>cnetwinnow</i> test that did not clean/remove it during test runs.
+- Fixed instantiation and use of projection classes to correctly return geometry data from projected images and mosaics. [#4772](https://github.com/DOI-USGS/ISIS3/issues/4772)
```
Kelvinrr: Even further evidenced that this is a mix of both Bugfixes and feature adds that seem to affect multiple issues.  This would probably make for a good policy discussion.

isis/appdata/templates/findfeatures/mosaics/00_basic_proc.sh
Kelvinrr: @KrisBecker Do you have an answer to this? I think it's important to establish the purpose of these script as we need to know to what extent we need to support these scripts. I think there exists a bigger discussion as to how we want to support tutorials. And by merging these script in I believe might set a precedence we do not want to set right now as to how we want to go about supporting tutorial content.  

This also, I believe, further validates the concern of big PRs. It becomes easier to let these slip through the cracks and conflates the conversation with other PR concerns. 

For now I am gonna advocate for their removal from this PR before merging and re-introducing them in a later PR once we agree on a way to manage them.

Kelvinrr:
> I believe all the comments/concerns have been addressed.
> 
> Please let me know if there is anything left to do se we can get this PR merged.
> 
> Thanks...

@KrisBecker When I said ping the reviewer, I meant to @ them so they get pinged. Adam has been busy putting out some fires so I'm gonna take over review work with his permission. I agree it's good to go except for the scripts which is the odd man out.

Kelvinrr (changes requested):
I think this is good except we should remove those tutorial scripts. Or someone needs to make a convincing argument that it's fine. 

They can be reintegrated later down the line. My desire to see this merged is only slightly lower than my desire to avoid introducing more tech debt.

isis/appdata/templates/findfeatures/mosaics/00_basic_proc.sh
KrisBecker: @Kelvinrr Simply stated, this is an example. It is an example that shows how findfeatures can be used to produce an image-based network, a concept that is heavily relied upon in the control and production of larger, regional and global mosaics.

Larger scoped tutorials would certainly refer to (the scripts in) this example to help explain what an image-based control network is and describe the basics for generating them using findfeatures. However, the pipelines used to produce the mosaics cited in the example documentation implements findfeatures' creation of these networks differently.

Just like other applications and examples, this example provides a template for users to modified as needed. It also is useful to understand how to use/apply findfeatures and the effects of its parameters.

You certainly raise very good points about discussions/guidelines/policy that need to be had regarding these types of things that currently are not available or are ill-defined. Please open an issue/discussion so that your concerns can be addressed.

Per your request, I will remove the scripts in this example and scrub mention of them from the documentation.

CHANGELOG.md
KrisBecker: I agree this PR became much larger than I expected or even imagined. However, the scope was identified and explained in issue #4772 and in the context of this PR on Feb 3, 2022. At no time did anyone mention this concern even after the initial commits were made.

Clearly, the review process (you are now the third reviewer) has contributed greatly to the size of this PR. Its also clear it did get much larger than anyone expected. I also learned a lot about this process and will be mindful to take steps to prevent this from occurring again. For one thing, I will not let my PRs linger so long between reviews - definitely did not help at all.

I am grateful for the reviews as they have been very helpful and useful to make this PR better. The unfortunate consequence of this is it also made it much larger. Much of the size was in response to requests for documentation improvements/clarifications, which is always useful.

I am open for suggestions/discussion about how this PR could be broken up or made easier and more efficient somehow. Hindsight would indicate to not address two issues in one PR even if you think they are small and/or related.

CHANGELOG.md
KrisBecker: The only bug fix in the above entries that was outside the scope of this PR is line 62. It was something I discovered during testing and was fixed in the testing framework.

CHANGELOG.md
Kelvinrr: I think at this point, it's too late to break up this PR. It has been through the review process and has had pretty much everything already addressed. But something to consider as a policy to avoid it in the future. Something for the ISIS TC.  

Once the merge conflicts are resolved and the scripts are removed, I'll approve and merge this

KrisBecker:
@Kelvinrr updates have been completed and conflict resolved.

Let me know if there is anything remaining.

Thanks...

Kelvinrr (approved):

PR #4790, opened 2022-02-10
Cli last

gsn9:
## Description

Added a function to generate a new command line when an executable  is called with -last. 
## Motivation and Context

When using the command line and the -last option. The terminal output of the command was not showing the changed values. 

## Types of changes

## Checklist:

- [ x I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.

## Licensing


isis/src/base/objs/UserInterface/UserInterface.cpp
```diff
@@ -405,6 +405,20 @@ namespace Isis {
...
+ * @return QString returnString- the updated command line 
+ */
+QString UserInterface::buildNewCommandLine(){
+  QString returnString = "";
+  for(auto i:p_cmdline){
+    returnString += i;
+    returnString += " ";
+  }
+  return returnString;
+}
```
Kelvinrr: testing this locally, it only seems to be dittoing the command you entered instead of using the previous command. 

I get: 

```shell
(isisdev) $ cam2map FROM=test.cub TO=test_lev2 PIXRES=mpp RESOLUTION=160
(isisdev) $ cam2map -last RESOLUTION=20
cam2map -last RESOLUTION=20 
```

Im pretty sure the expected output is: 


```shell
(isisdev) $ cam2map FROM=test.cub TO=test_lev2 PIXRES=mpp RESOLUTION=160
(isisdev) $ cam2map cam2map -last RESOLUTION=20
cam2map FROM=test.cub TO=test_lev2 PIXRES=mpp RESOLUTION=20 
```

Additionally, I think there might be a way to make a test for this? Would building two UsterInterfaces back to back, one with `-LAST`, trigger the expected code in a way that is GTest-able?

Kelvinrr (changes requested):
Looks good to me, is there a way to test this? If you were to create two app calls using UserInterface objects, can we see the

isis/src/base/objs/UserInterface/UserInterface.cpp
```diff
@@ -517,6 +530,9 @@ namespace Isis {
         throw IException(e, IException::User, "Invalid command line", _FILEINFO_);
       }
 
+    } 
+    if( usedDashLast == true){
```
AustinSanders: I'm not sure that this results in the intended behavior.  From my testing, this prints out the string "<program_name> -last <new options>" instead of expanding -last into the previously specified arguments.

I ran:
$> lrowac2isis from= wac0000a1c4.img to= out.cub
$> lrowac2isis -last to=out2.cub
and got the output:
lrowac2isis -last to=out2.cub

I would expect to get the output:
lrowac2isis from=wac0000a1c4.img to=out2.cub

@scsides Can you confirm that my understanding of the intended behavior is correct?

AustinSanders (changes requested):

isis/src/base/objs/UserInterface/UserInterface.cpp
Kelvinrr: Im pretty sure this change needs to happen in `LoadHistory` (https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/UserInterface/UserInterface.cpp#L550 called from https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/UserInterface/UserInterface.cpp#L667)

scsides: gtestable: I don't think the gtests should ever work with or modify a developers personal files, accessing history once to save a command line then again to retrieve it would do just that. The only way I can see to test this would be to have a temp user, much like a temp dir/temp file, then the solution seems possible.

I don't believe LoadHistory is the correct place. It only loads the history. The current command line with the history needs to be spiced together before the cout happens.

isis/src/base/objs/UserInterface/UserInterface.cpp
scsides: your expectations are correct

isis/src/base/objs/UserInterface/UserInterface.cpp
scsides: One way to get the command line after the two have been merged might be to use the child class IsisAml.CommandLine(pvl) to get the current command as a pvl then loop through each keyword and construct the command line.

scsides (changes requested):

isis/src/base/objs/UserInterface/UserInterface.cpp
Kelvinrr: @scsides There seems to be some logic just for tests: https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/UserInterface/UserInterface.cpp#L670 I wonder if we can just change that boolean to be true when in a GTest or something?

scsides:
Oh yes, I forgot about that. Your idea is worth looking into.

isis/src/base/objs/UserInterface/UserInterface.cpp
```diff
@@ -408,13 +408,41 @@ namespace Isis {
...
+  // loop currentCommandLine and update returnString 
+  for(auto i:currentCommandLine.split(" ")){
+    QString paramVale = i;
+    
+    // loop values to update
+    for(auto j:valuesToUpdate){
+      QString updateVal = j;
+
+      // check to see if if the histoy param is the same as our input param
+      if(paramVale.split("=")[0].startsWith(updateVal.split("=")[0].toUpper())
```
scsides, line 436: I'm concerned about the parsing in here (split, toUpper). The UI already does all the parsing, casing, and comparisons. Once the data struct that holds user parameters has been updated (IsisAml), we should ask the Aml for the parameters and then stuff them into a string to return.  What about params that were defaulted in the previous one (i.e., not on the command line), but present in the new run?

scsides (changes requested):

PR #4792, opened 2022-02-11
Add a public API definition

### LINKED ISSUE #145: ISIS Project public API
We are having a good discussion in #140 about what constitutes a "breaking change."

Based on conversations in our TC meetings, we really wanted to try and have better written guidelines for what constitutes a breaking change in our codebase that would trigger a major version increment.  As we have all agreed, since the ISIS project is following Semantic Versioning, a "breaking change" is defined as a "backwards incompatible change introduced to the public API."  The difficultly, perhaps, is that we have not yet formally defined and agreed upon what the ISIS public API *is*, as required by the first element of the SemVer Specfication.

Such a definition is therefore a necessary first step before we should even be having a discussion about what constitutes a "breaking change" and then once we do, we still have to figure out how to deal with them from a policy and workforce standpoint (which is its own can of worms), but first things first:

# Proposed definition of the "ISIS Project public API"

The public API is considered to be all the interfaces that the ISIS project provides to the community from which the ISIS user base can build more complex tools.  This API therefore includes code-level structures accessible via software calls to libraries, but also more user-facing structures like command-line programs and graphical user interface (GUI) programs.  Any "interface" via which a user can reasonably interact with components of the ISIS project should be considered part of the "ISIS Project public API" for the purposes of Semantic Versioning.

Examples are provided below, but are not meant to be exhaustive or complete.  The above philosophy should guide our thinking in what the public API is.

**Code-level structures**: Function, class, and object names, argument signatures, and return signatures for functions, classes, and objects.

- C++ code exceptions would be those structures above marked with a "private" access specifier via the C++ language.
- Python code exceptions would be those structures above which do not begin with an underbar ("_", as per Python language tradition).

**User-facing command-line and GUI programs**: program name, argument list, argument defaults, configuration or preference names and their defaults, and output file structure and contents.

rbeyer:
## Description
Would close [ISIS_TC Issue 145](https://github.com/USGS-Astrogeology/ISIS_TC/issues/145).

There are a lot of issues around details regarding the definition of the public API.  Rather than trying to get everything possible covered in this document before accepting this PR, it would be great if we can focus on the general principles and just the examples listed to get this first PR landed.

For example, the above linked discussion talks about how to deal with control network files or camera plugins.  Those topics need to be tackled, but should probably be done under future, atomic PRs to this document.

## Types of changes

- Documentation change (update to the documentation; no code change)

## Checklist:

- I have read and agree to abide by the [Code of Conduct](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/Code-Of-Conduct.md)
- I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- My change requires a change to the documentation.
- I have updated the documentation accordingly.
- I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.

## Licensing

- I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.


rbeyer:
I didn't review the TC meeting notes, which indicated a file named `API.md`, before I wrote this PR naming the file `APIdefinition.md`.  I'm consistently bad at naming, and I'm happy to switch the filename if desired.

APIdefinition.md
```diff
@@ -0,0 +1,75 @@
...
+## General Guidance
+
+**Code-level structures**: Function, class, and object names,
+argument signatures, and return signatures for functions, classes,
+and objects.
+
+- C++ code exceptions would be those structures above marked with
+  a "private" access specifier via the C++ language.
+- Python code exceptions would be those structures above which begin 
+  with an underbar ("_", as per Python language tradition).
```
jessemapel: Maybe we should add a note that the ABI is not a part of the public API. Users can expect to compile against new versions of the library without making changes, but cannot expect to link against new versions of the library without recompiling. For example, we may add new arguments, but they will have a default value.

APIdefinition.md
```diff
@@ -0,0 +1,75 @@
...
+
+**User-facing command-line and GUI programs**: program name, argument
+list, argument defaults, configuration or preference names and their
+defaults, and for those programs with structured output files, the
+structure and contents of those output files are analogous to the
```
jessemapel: I would say structure, but not contents. We can say that the existence of a keyword is part of the public API, but the contents of the keyword is not.

APIdefinition.md
```diff
@@ -0,0 +1,75 @@
...
+
+Object, group, and keyword names are a part of the API.
+
+For example, the "minimum" keyword will exist in the "statistics"
+group. This also applies to the PVL labels in Cube files.
+
+New aggregates or keywords can be added (feature), but if any are changed
+or removed, that's a breaking change.  PVL text does have data types, and
+changing something from `MIN = 5` to `MIN = "5"` is changing the data type
+of the keyword MIN from an integer to a string.
```
jessemapel, line 103: We should add a Cube files sections. That can be another commit though.

jessemapel (commented):
Overall I am happy with this. I just have two small clarifications and something that I think should be addressed in a new PR.

APIdefinition.md
rbeyer: Can you provide suggested prose that I can paste in for what you'd like to see here?

APIdefinition.md
rbeyer: I disagree, perhaps because we have sliced the meaning of "contents" differently, I think it is important to keep both words in that sentence to make people think about possibilities, and I think they are mitigated by the presence of "are analogous" later on and the specific examples farther down the document.

I would agree with you that the value of a keyword is not at issue, but as in my PVL example, the simple enclosure of a number in quotes changes the "type" of that value from an int to a string.  You might conceive of that as "structure," but I consider it "contents" because it is the data type of the "contents" of the assigned variable.  If a function returned a dict, and the value of the key "foo" changed from int to string, that would be a problem, and that value is the "contents" of the dict, but maybe it is structure?

So we could go deep in this first para and strictly and exhaustively define terms, or we could leave it to the file-type examples later in the document to provide specific guidance about particular file types.

APIdefinition.md
jessemapel: Maybe this is best handled by modifying this section and adding an example.

```
**Code-level structures**: Public function, member and class names,
argument signatures, and return signatures.

- In C++ code, public structures are structures that can be accessed by including the distributed header files for libraries.
- In Python code, public structure are structures that do not being with an underbar ("_"), as described in PEP 8.
```

Then for the example:

```
### Adding new arguments

Adding new arguments to code level structures does not violate the existing public API as long as they provide a default value. For example changing a function doThing(int a, double b) to doThing(int a, double b, double c=1.0) does not violate the existing API for the doThing function as code that calls the doThing function will still compile without modification.

Adding new optional arguments or arguments with default values to command line programs does not violate the existing public API. For example, adding a new argument called "check" that has a default value of "false" to a command line program does not violate the program's public API as scripts and commands that were valid previously are still valid after the addition.
```

rbeyer: Pushed a commit with this language added.

APIdefinition.md
jlaura: I agree with @jessemapel here. The word 'content' is way too ambiguous. One could argue that the information content of 2 and "2" is identical. We all just read two. If this is about types, then we need to say types. If this is about semantic content, then we are talking about values which do not, in my opinion, belong in the API.

rbeyer: A human might argue that they are identical, but a standards-compliant PVL parser would interpret those differently.  I'm happy to strike "content" here as being potentially confusing, and allow us to be potentially more specific with later examples.

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
+# ISIS Project public API
+
+The public API is considered to be all the interfaces that the ISIS
+project provides to the community from which the ISIS user base can
+build more complex tools. This API therefore includes code-level
+structures accessible via software calls to libraries, but also
+more user-facing structures like command-line programs and graphical
+user interface (GUI) programs. Any "interface" via which a user can
```
jlaura, line 8: The user can not build a tool based on the GUIs. I would remove any GUI discussion.

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+reasonably interact with components of the ISIS project should be
+considered part of the "ISIS Project public API" for the purposes
+of [Semantic Versioning](https://semver.org).
```
jlaura, line 11: and here I agree, so perhaps just push the GUI part down into here.

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+**User-facing command-line and GUI programs**: program name, argument
+list, argument defaults, configuration or preference names and their
+defaults, and for those programs with structured output files, the
+structure of those output files are analogous to the return signatures
+of code-level structures.
+
+
+This guidance may be more straightforward to understand and apply
+for code-level structures which have more formalized argument and
+return signatures.  User-facing command-line and GUI programs should
```
jlaura, line 47: Starting with 'User Facing...', while I agree that this documentation is important, most of these outputs are not written against a specification. It is also not likely that this documentation will be generated in the near future. Therefore, I would suggest that this sentence should not part of an API discussion.

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+the *doThing* function as code that calls the *doThing* function will
+still compile without modification.
+
+Adding new optional arguments or arguments with default values to
+command line programs does not violate the existing public API. For
+example, adding a new argument called "check" that has a default
+value of "false" to a command line program does not violate the
+program's public API as scripts and commands that were valid
+previously are still valid after the addition.
+
```
jlaura, line 67: Perhaps another example: Adding a deprecation warning to a component of the API requires that a MINOR version tick occur. When the deprecated component of the API is removed from the code base, a MANJOR version tick occurs. [Source](https://semver.org/#how-should-i-handle-deprecating-functionality).

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Unstructured text files
+
+For anything of an inherently unstructured file-type, like .txt;
+there is no guarnatee of its contents; just that it will exist.
+
```
jlaura, line 77: Pedantic, but I want to be sure. By exist, `touch foo.txt` is sufficient?

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+### CSV output files
+
+Column names and data types are part of the API.
+
+For example, a program that writes out a CSV should clearly describe
+the output columns which include the fieldname, data type, and
+whether non-data-type values can be present (e.g. a column of
+floating point numbers named "minimum" that could contain the string
+"NULL" instead of a floating point value).
```
jlaura, line 87: This is fine, as long as it is not retroactively applied to existing code. I would love to start seeing PRs come in that add this type of documentation, but it simply does not exist yet.

APIdefinition.md
```diff
@@ -0,0 +1,96 @@
...
+changed or removed, that's a breaking change.
+
+
+### PVL files
+
+Object, group, and keyword names are a part of the API.
+
+For example, the "minimum" keyword will exist in the "statistics"
+group. This also applies to the PVL labels in Cube files.
+
```
jlaura, line 99: Clarifying question: if the column name is spelled incorrectly, I would consider this to be a bug (VERSION tick when fixed). Is that the shared understanding?

APIdefinition.md
jessemapel: I'm not sure about GUIs. I agree that we shouldn't support some kind of wild automation against the GUIs, but what about SOPs or cookbook style workflows. Would removing a button from a GUI be breaking for that application? It's very frustrating when I'm following a tutorial and it says to click on a button or a menu drop down that doesn't exist anymore.

APIdefinition.md
jessemapel: At a reductive level I think that is sufficient. A more realistic example is that we make no guarantee of the contents of the bundleout.txt files generated by jigsaw. They are purely human-readable output and not designed to be parsed programmatically. So, we can add sections, remove sections, change spacings on tables, etc.

APIdefinition.md
jessemapel: I think this whole document has to be applied retroactively. So, if we change the type of a column in a CSV or allow an "N/A" entry in a numeric field that would be API breaking. I don't think we are immediately compelled to generate this documentation. It would be very good to have, though, and can be something we include in all future app documentation.

@ladoramkershner This may be something good to consider for our documentation audit.

APIdefinition.md
jessemapel: I think that's still breaking. From a development perspective it is definitely a bug; but from a user perspective, any programmatic tools that interface with the file will need to be coded against that bug. So, fixing the issue will break any tool that references that column by name. So, I think from the perspective of breaking changes requiring modifications to existing consumers, this would be breaking.

jlaura: Based on [this discussion](https://github.com/planetarypy/pvl/issues/99), it seems that the perspective of the library can and should supersede the perspective of the user in this case.

APIdefinition.md
rbeyer: This is advice for best practice.  I think that there is a good reason to recommend that implementors *should* do this.

APIdefinition.md
rbeyer: Can we hold the avalanche of additions (of which I am certain there are many) for subsequent PRs, so they can be discussed more fully?  However, this one is straightforward, but its already detailed on SemVer, so do we have to repeat everything there verbatim?  That's gonna make a long document.

jessemapel: If it's already a part of SemVer I don't think we should mention it here

jlaura: This assumes people read SemVer, which, in my experience, they will not.

APIdefinition.md
jlaura: Sure. But not in an API definition. Put the into the contributing docs.

rbeyer: This is in a section labeled "General Guidance" and makes a recommendation for a best practice.  I'm not sure how a reader could construe that as part of the *definition* of the API, but if we think it doesn't belong here, I guess I can remove, but this also seemed like a good place to remind folks (and it should also go in the contributing docs, maybe the solution is to remove it from here when it is there).

jessemapel: Maybe we flip this and say "For command-line and GUI programs, the documentation describes the output structure that users can depend on." I view it as "If  it's documented, you can rely on it"

It's weaker than I like, but it's more accommodating with the current code base and documentation.

APIdefinition.md
jessemapel: I don't like reproducing docs that are else where in here, I'd rather just link them.

How about a generic statement: "This API definition is used with Semantic Versioning policies to determine when and how the ISIS version number changes. An FAQ and some examples can be found on the Semantic Versioning website (https://semver.org/)"

APIdefinition.md
jessemapel: I think that's a philosophical decision we need to make. If we do take the stand this such a change is a bug fix we need to be very careful about how we communicate it to users on release. Basically, say "this change is a bug fix but it has a chance to break existing tools if you coded against the previously faulty behavior." We likely should also call it out in a specific section of the release notes/changelog.

APIdefinition.md
jlaura: The TC is today and I expect we will discuss this. No one else has responded. I agree completely that it is frustrating if a GUI changes and a tutorial is not updated. Having said that, if the goal is to tractably maintain tutorials that point at GUIs, then we need to be able to CI test the tutorial as I think it is the tutorial's responsibility to maintain compatibility.

Counter argument: Maintaining compatibility without CI is as simple as tagging a tutorial to a version. If that is done, then GUI changes might warrant version ticking.

Counter-counter argument: The tutorial is release either with or after a particular version is released. So, it is hard to think of a situation where a tutorial says 'written using version x.y.z' and then that version suddenly lacks a GUI element.

This should likely be discussed from a cost-benefit perspective.

APIdefinition.md
rbeyer: If this can be another commit, please resolve.

APIdefinition.md
rbeyer: Would removing the phrase "and graphical user interface (GUI) programs." resolve this concern?

APIdefinition.md
rbeyer: Does the concept of a "GUI" need to be added to this sentence?

APIdefinition.md
rbeyer: Jay, what explicit changes do you need here to resolve this concern?

APIdefinition.md
rbeyer: The indicated sentence begins with "For example, ... should ..." which was intended to be for example purposes, and does not expect all of the existing code to have this characteristic, but as development moves forward, and these programs and documentation are revisited, they "should" be more explicit.

If there is a specific change you'd like to see here, please be specific, or resolve.

APIdefinition.md
rbeyer: This seems like a discussion to have on a particular PR that might then influence an example being added to this document.  Are there specific changes that need to be made to *this* document, or can this be resolved?

jessemapel:
This has stalled, I'm going to go ahead and merge this. Any outstanding discussion can be continued in a separate PR with proposed changes to this document.

jessemapel (approved):

PR #4835, opened 2022-03-10
Initial support for documentation versions

scsides:
Add ability to get to multiple versions of the ISIS documentation

## Description
Multiple versions of ISIS documentation can now live side-by-side. New versions will be placed in a dir named from the ISIS version.

## Motivation and Context

There have been many requests for old version of the docs.

## How Has This Been Tested?

This has been locally deployed and all of the pages hand tested

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,8 @@ install/
 
 # Unignore the documentation build
 !isis/src/docsys/build
+!isis/src/docsys/Object/build
+!isis/src/docsys/Application/build
 
 # Ignore vs code
 .vscode
--- a/isis/cmake/BuildDocs.cmake
+++ b/isis/cmake/BuildDocs.cmake
@@ -66,22 +66,30 @@ endfunction(copy_app_docs_info)
 # Build the top level of the documents directory
 function(build_upper_level)
 
-  # Copy existing folders to the install directory
-  copy_folder(${docBuildFolder}/assets ${docInstallFolder})
-  copy_folder(${docBuildFolder}/w3c    ${docInstallFolder})
-
   # Make new (empty) output folders
   set(newFolders UserDocs AboutIsis General Guides Installation TechnicalInfo)
   foreach(f ${newFolders})
-    file(MAKE_DIRECTORY "${docInstallFolder}/${f}")
+    file(MAKE_DIRECTORY "${docInstallFolder}/${docVersion}/${f}")
   endforeach()
+
+  # Copy the assets folder to the specific version directory
+  copy_folder(${docBuildFolder}/assets ${docInstallFolder}/${docVersion})
+  # Also copy the assest folder to the base area so the base index.html can use it
+  # TODO - eliminate this copy of assest by having a simple top index.html that redirects to the vx.y.z/index.html. The top index should contain html base so it uses vx.y.z assets
+  copy_folder(${docBuildFolder}/assets ${docInstallFolder})
+
+  # Copy other folders to the specific version directory
+  copy_folder(${docBuildFolder}/w3c    ${docInstallFolder}/${docVersion})
 
   # These folders are populated inside "build_documents_folder"
 
-  # Create index.html file
-  execute_process(COMMAND ${XALAN} ${XALAN_VALIDATE_OPTION} ${XALAN_PARAM_OPTION} menuPath \"\" ${XALAN_OUTFILE_OPTION} ${docInstallFolder}/index.html ${XALAN_INFILE_OPTION} ${docBuildFolder}/build/homepage.xml ${XALAN_XSL_OPTION} ${docBuildFolder}/build/main.xsl)
+  # Create the main index.html file aka the home page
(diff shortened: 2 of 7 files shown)


isis/cmake/BuildDocs.cmake
```diff
@@ -66,22 +66,30 @@ endfunction(copy_app_docs_info)
...
+  copy_folder(${docBuildFolder}/w3c    ${docInstallFolder}/${docVersion})
+
   # These folders are populated inside "build_documents_folder"
 
-  # Create index.html file
-  execute_process(COMMAND ${XALAN} ${XALAN_VALIDATE_OPTION} ${XALAN_PARAM_OPTION} menuPath \"\" ${XALAN_OUTFILE_OPTION} ${docInstallFolder}/index.html ${XALAN_INFILE_OPTION} ${docBuildFolder}/build/homepage.xml ${XALAN_XSL_OPTION} ${docBuildFolder}/build/main.xsl)
+  # Create the main index.html file aka the home page
+  execute_process(COMMAND ${XALAN} ${XALAN_VALIDATE_OPTION} ${XALAN_PARAM_OPTION} menuPath \"${docVersion}/\" ${XALAN_OUTFILE_OPTION} ${docInstallFolder}/index.html ${XALAN_INFILE_OPTION} ${docBuildFolder}/build/homepage.xml ${XALAN_XSL_OPTION} ${docBuildFolder}/build/main.xsl)
+  # Create a second main page inside the version numbered area for when the page above gets overwritten with a new version 
+  execute_process(COMMAND ${XALAN} ${XALAN_VALIDATE_OPTION} ${XALAN_PARAM_OPTION} menuPath \"\" ${XALAN_OUTFILE_OPTION} ${docInstallFolder}/${docVersion}/index.html ${XALAN_INFILE_OPTION} ${docBuildFolder}/build/homepage.xml ${XALAN_XSL_OPTION} ${docBuildFolder}/build/main.xsl)
```
ladoramkershner, line 89: So here, the plan is to make two identical doc builds that will get pushed up to the server. The first being the main homepage you navigate to when you go to isis.astrogeology.usgs.gov and the other being the homepage you navigate to if you click on the current version in the drop down menu?

isis/cmake/BuildDocs.cmake
```diff
@@ -464,8 +472,10 @@ endfunction(build_object_docs)
 function(build_docs)
 
   message("Building Isis Documentation...")
+  message(" Version is: " ${VERSION})
 
   # Set up output directory and a temporary directory for building
+  set(docVersion       7.0.0)
```
ladoramkershner: Ignoring, but noting (for later reviews)

isis/cmake/BuildDocs.cmake
```diff
@@ -464,8 +472,10 @@ endfunction(build_object_docs)
 function(build_docs)
 
   message("Building Isis Documentation...")
+  message(" Version is: " ${VERSION})
```
ladoramkershner: Where is VERSION set?

.gitignore
```diff
@@ -51,6 +51,8 @@ install/
 
 # Unignore the documentation build
 !isis/src/docsys/build
+!isis/src/docsys/Object/build
+!isis/src/docsys/Application/build
```
ladoramkershner, line 55: Is this in preparation for splitting up the documentation builds?

scsides: No, when I updated the conf files, git warned me that these were being ignore, and they should not have been.

isis/cmake/BuildDocs.cmake
scsides: At the top of the second from the bottom function. It's hardcoded for now, but that will change today if all goes well

isis/cmake/BuildDocs.cmake
scsides: Just two identical index.html files. One in the rood of isis.astro.usgs.gov and the other in v7.0.0. This way 7.0.0 never goes away and can be fully accessed, but the root index file gets renewed every time we push a new version.  I have ideas on how we can fix this, but haven't been able to test yet.

isis/cmake/BuildDocs.cmake
jessemapel: Use `${CMAKE_PROJECT_VERSION}` instead to get the version from ISIS.

isis/cmake/BuildDocs.cmake
jessemapel: Same as above, just set this to `${CMAKE_PROJECT_VERSION}` and do it prior to the message above so the message can just use `${docVersion}`

jessemapel (commented):
New setup seems reasonable to me. Are the changes in the doxyfiles important or just incoming changes for using a more up to date doxygen version?

scsides:
Many new doxygen versions cause us to do a doxygen -u to get rid of all the xyz is deprecated warning. This is not a functional difference, except that now we can see real errors without wading through all junk

ladoramkershner (approved):

PR #4875, opened 2022-03-22
Initial Lidar Network fixture

jessemapel:
Draft PR for adding a Lidar jigsaw test


isis/src/base/objs/LidarData/LidarData.cpp
```diff
@@ -50,9 +50,20 @@ namespace Isis {
     m_points.insert(point->GetId(), point);
   }
 
+  /**
+   * Gets a single LidarDataPoint by serial number
+   *
+   * @param serialNumber The serial number of the LidarDataPoint
```
jessemapel: This should be the point ID instead of the serial number. Points use IDs, measures and images use serial number.

PR #4926, opened 2022-04-21
fixes bug where cam2map doesn't project multiple bands on band dependent camera models (fixes #4635)

### LINKED ISSUE #4635: CAM2MAP Only Map-Projects First Layer in a CUB for Some Cameras (at least MARCI)
**ISIS version(s) affected**: 4.x.x (and later?)

**Description**  
Running CAM2MAP on a cube file with multiple bands will output the same number of bands map-projected, but only the first will have data.

**How to reproduce**  
Process an image, and run PHOCUBE on it.  Map-project the PHOCUBE file and take a look at it, and the different bands, in QVIEW.  You should be able to see what I mean.

**Possible Solution**  
I don't know ...

**Additional context**  
It seems like this is a bug since all of the bands or layers from the original cube are present in the output, just completely empty.  If this is expected behavior, that should be documented somewhere obvious, and you should probably remove the blank layers/bands.  I discovered this issue when working with MARCI data and trying to shift my order of operations, whether it's faster to do certain things before others, and while running PHOCUBE on the unprojected file and then map-projecting was much faster for larger lat/lon ranges, it isn't usable if the data aren't propagated through.  For now, I'll use a work-around and split those backplanes out and do the map-projection on them.  Which will still be faster than doing PHOCUBE on the map-projected version.

For the "... for some cameras (at least MARCI)" part, see the first two comments.

Kelvinrr:
## Description

When phocube is run on a Marci or LRO WAC image, then that image is run through cam2map, only the first band is projected and the rest have only NULL DNs. In the case of LRO WAC, it'll fail with an exception. 

This is because when trying to set camera parameters in the `SetBand` method overwritten in Marci and LRO WAC camera classes, the output of phocube has bands that the camera doesn't expect and doesn't properly set the camera parameters for the new band causing projections for any band number > 1 to fail the projection process.  

## Related Issue

fixes https://github.com/USGS-Astrogeology/ISIS3/issues/4635

## How Has This Been Tested?

The original image in the issue projected successfully. Still needs some tests added. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/cam2map/cam2map.cpp
+++ b/isis/src/base/apps/cam2map/cam2map.cpp
@@ -462,7 +462,9 @@ namespace Isis {
   bool cam2mapForward::Xform(double &outSample, double &outLine,
                              const double inSample, const double inLine) {
     // See if the input image coordinate converts to a lat/lon
-    if (!p_incam->SetImage(inSample,inLine)) return false;
+    if (!p_incam->SetImage(inSample,inLine)) {
+      return false;
+    }
 
     // Does that ground coordinate work in the map projection
     double lat = p_incam->UniversalLatitude();
@@ -568,6 +570,7 @@ namespace Isis {
   }
 
   void bandChange(const int band) {
+    // band dependant band change
     incam->SetBand(band);
   }
 }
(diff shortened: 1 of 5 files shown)


jlaura:
@Kelvinrr Is this still a work in progress? If not, can we get someone tagged to review?

Kelvinrr:
~~Pretty sure this just needs are review.~~ it's waiting on tests to be added. I can probably take care of this sometime next week.

gsn9 (approved):
Looks good to me, has CI run on this? @Kelvinrr

Kelvinrr:
@gsn9 ran it overnight, came back green

PR #4936, opened 2022-04-29
New push frame stitch application

### LINKED ISSUE #4924: Add application to stitch LRO WAC even and odd images together
**Description**  

In order to properly work with LRO WAC, and likely other push frame sensor, images in stereo matchers we need a tool that merges the even and odd framelet images back into a single image. The program will also need have the option to flip the framelets similar to how MRO MARCI images can be flipped.

In order to prevent complicating processing, the app will also need to strip the camera information off of the cube.

jessemapel:
## Description

Adds a new application for stitching push frame images back together.

## Related Issue

resolves #4924 

## Motivation and Context

We need this application for processing push frame imagery such as LRO WAC outside of ISIS. This will allow us to ingest the images into ISIS and do processing such as radiometric calibration and then export a single image with all of the image data in it. It will also allow us to flip the frame so that adjacent ground points are adjacent in the image and we can use matchers without fear of ending up on frame boundaries.

## How Has This Been Tested?

Tested on an LRO WAC UV and VIS image set. There is also extensive automated testing.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ release.
 
 ## [Unreleased]
 - Updated the LRO photometry application Lronacpho, to use by default the current 2019 photometric model (LROC_Empirical). The model's coefficients are found in the PVL file that exists in the LRO data/mission/calibration directory. If the old parameter file is provided, the old algorithm(2014) will be used. This functionality is desired for calculation comparisons. Issue: [#4512](https://github.com/USGS-Astrogeology/ISIS3/issues/4512), PR: [#4519](https://github.com/USGS-Astrogeology/ISIS3/pull/4519)
-- Added new application, framestitch, for stitching even and odd push frame images back together prior to processing in other applications. [4924](https://github.com/USGS-Astrogeology/ISIS3/issues/4924)
+- Added a new application, framestitch, for stitching even and odd push frame images back together prior to processing in other applications. [4924](https://github.com/USGS-Astrogeology/ISIS3/issues/4924)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ release.
 
 ## [Unreleased]
 - Updated the LRO photometry application Lronacpho, to use by default the current 2019 photometric model (LROC_Empirical). The model's coefficients are found in the PVL file that exists in the LRO data/mission/calibration directory. If the old parameter file is provided, the old algorithm(2014) will be used. This functionality is desired for calculation comparisons. Issue: [#4512](https://github.com/USGS-Astrogeology/ISIS3/issues/4512), PR: [#4519](https://github.com/USGS-Astrogeology/ISIS3/pull/4519)
+- Added new application, framestitch, for stitching even and odd push frame images back together prior to processing in other applications. [4924](https://github.com/USGS-Astrogeology/ISIS3/issues/4924)
+
 ### Changed
 
 ### Added
--- a/isis/src/base/apps/framestitch/framestitch.cpp
+++ b/isis/src/base/apps/framestitch/framestitch.cpp
@@ -1,0 +1,220 @@
+#include "framestitch.h"
+
+#include <QString>
+
+#include "Brick.h"
+#include "Buffer.h"
+#include "Cube.h"
+#include "CubeAttribute.h"
+#include "Camera.h"
+#include "Histogram.h"
+#include "IException.h"
+#include "IString.h"
+#include "LineManager.h"
+#include "ProcessByBrick.h"
+#include "PvlGroup.h"
+#include "PvlKeyword.h"
+#include "Statistics.h"
+#include "UserInterface.h"
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+
+  /**
+   * Helper function to count the size of frames in a cube
+   */
+  int computeFrameSize(Cube *cube) {
+    // Use a histogram to accumulate frame counts
+    Histogram frameSizes(0.0, 1024.0, 1024);
+    LineManager cubeLine(*cube);
+    int currentFrameSize = 0;
+    for (int line = 1; line <= cube->lineCount(); line++) {
+      cubeLine.SetLine(line);
+      cube->read(cubeLine);
+      Statistics lineStats;
+      lineStats.AddData(cubeLine.DoubleBuffer(), cubeLine.size());
+      // If the line is all NULL add it to the current frame count
+      if (lineStats.TotalPixels() == lineStats.NullPixels()) {
+        ++currentFrameSize;
+      }
+      // If the line has non-NULL pixels and we have previously counted
+      // some all null lines add the count to our histogram and reset
+      else if (currentFrameSize > 0) {
+        frameSizes.AddData(currentFrameSize);
+        currentFrameSize = 0;
+      }
+    }
+
+    // If the last line is part of a NULL frame handle it now
+    if (currentFrameSize > 0) {
+      frameSizes.AddData(currentFrameSize);
+    }
+
+    // Special case where there are no all NULL lines
+    if (frameSizes.TotalPixels() == 0) {
+      return 0;
+    }
+
+    // Use the median to handle any outliers
+    return frameSizes.Median();
+  }
+
+  /**
+   * Compute the stats for an ISIS cube. This is the programmatic interface to
+   * the ISIS stats application.
+   *
+   * @param ui The User Interface to parse the parameters from
+   */
+  void framestitch(UserInterface &ui) {
+    ProcessByBrick process;
+
+    // It is very important that the even cube gets added first as later on
+    // we'll use frameNumber % 2 to index the input cube vector
+    QString evenCubeFile = ui.GetCubeName("EVEN");
+    Cube* evenCube = process.SetInputCube(evenCubeFile,
+                                          CubeAttributeInput(evenCubeFile));
+    QString oddCubeFile = ui.GetCubeName("ODD");
+    Cube* oddCube = process.SetInputCube(oddCubeFile,
+                                         CubeAttributeInput(oddCubeFile));
+
+
+    // Check that all the inputs are valid
+    if (evenCube->sampleCount() != oddCube->sampleCount() ||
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ release.
 
 ## [Unreleased]
 - Updated the LRO photometry application Lronacpho, to use by default the current 2019 photometric model (LROC_Empirical). The model's coefficients are found in the PVL file that exists in the LRO data/mission/calibration directory. If the old parameter file is provided, the old algorithm(2014) will be used. This functionality is desired for calculation comparisons. Issue: [#4512](https://github.com/USGS-Astrogeology/ISIS3/issues/4512), PR: [#4519](https://github.com/USGS-Astrogeology/ISIS3/pull/4519)
+- Added new application, framestitch, for stitching even and odd push frame images back together prior to processing in other applications. [4924](https://github.com/USGS-Astrogeology/ISIS3/issues/4924)
```
oleg-alexandrov: Should be "a new application".

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "Statistics.h"
+#include "UserInterface.h"
+
+using namespace std;
+using namespace Isis;
+
+namespace Isis {
+
+  /**
+   * Helper function to count the size of frames in a cube
```
oleg-alexandrov: It looks that the comment is not aligned with what the function does. The function appears to find the median framelet size rather than counting. The function name may also need to be renamed to computeMedianFrameSize() or computeMedianFrameletSize(), and somewhere it should be explained what a framelet is.

I think this computes in fact framelet height, rather than its 2D size.

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+    if (frameSizes.TotalPixels() == 0) {
+      return 0;
+    }
+
+    // Use the median to handle any outliers
+    return frameSizes.Median();
+  }
+
+  /**
+   * Compute the stats for an ISIS cube. This is the programmatic interface to
```
oleg-alexandrov: This comment also appears to not say what is actually happening. This function does the stitching, rather than the stats.

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+      frameSize = ui.GetInteger("FRAMESIZE");
+    }
+    else {
+      // User didn't pass the size of the frames so attempt to infer it from
+      // the cubes
+      int evenFrameSize = computeFrameSize(evenCube);
+      int oddFrameSize = computeFrameSize(oddCube);
+
+      if (evenFrameSize == 0) {
+        QString msg = "Failed to find any NULL frames in even cube when attempting "
```
oleg-alexandrov: I think this should error message should say "non-NULL" rather than "NULL". Or even better, maybe, "failed to find any valid data in even cube".

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+                      "to compute frame size.";
+        throw IException(IException::User, msg, _FILEINFO_);
+      }
+      if (oddFrameSize == 0) {
+        QString msg = "Failed to find any NULL frames in odd cube when attempting "
```
oleg-alexandrov: Same thing here.

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // first frame in the odd cube is now all NULL.
+    //
+    //  Before   --Flip-->  After
+    // Even  Odd          Even  Odd
+    // 0000  ####         ####  0000
+    // ####  0000         0000  ####
+    // 0000  ####         ####  0000
+    // ####  0000         0000  ####
+    //
+    int numFrames = evenCube->lineCount() / frameSize;
```
oleg-alexandrov: I think this implicitly assumes that all frames have the same number of lines (frameSize). I am also a little confused here. I would think that lineCount() includes both the lines which have nulls (which look black) and the lines which have valid data, no?

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,220 @@
...
+        QString msg = "Failed to find any NULL frames in even cube when attempting "
+                      "to compute frame size.";
+        throw IException(IException::User, msg, _FILEINFO_);
+      }
+      if (oddFrameSize == 0) {
+        QString msg = "Failed to find any NULL frames in odd cube when attempting "
+                      "to compute frame size.";
+        throw IException(IException::User, msg, _FILEINFO_);
+      }
+      if (evenFrameSize != oddFrameSize) {
```
oleg-alexandrov: This implicitly assumes that all frames in both even and odd cubes have the same size. If they don't, given that the frame size is found using a median, these two medians will likely disagree.

isis/src/base/apps/framestitch/framestitch.xml
```diff
@@ -0,0 +1,124 @@
...
+  <brief>
+    Stitches even and odd frames from push frame <def link="Cube">cubes</def> back into a single cube.
+  </brief>
+
+  <description>
+    <p>
+      This program combines the even and odd frames from push frame <def link="Cube">cubes</def> back
+      into a single interwoven <def link="Cube">cube</def>. It can also optionally flip the frame order
+      so that adjacent ground features appear adjacent between frames. This program
+      is designed to be ran before processing the data in another software package
```
oleg-alexandrov: should be "to be run".

isis/src/base/objs/ProcessByBrick/ProcessByBrick.cpp
```diff
@@ -660,6 +660,82 @@ void ProcessByBrick::SetOutputRequirements(int outputRequirements) {
     omgrs.clear();
   }
 
+
+  /**
+   * Starts the systematic processing of the input cube by moving an arbitrary
```
oleg-alexandrov: should be "arbitrarily-shaped".

isis/src/base/apps/framestitch/framestitch.cpp
jessemapel: Yes this is true. All frames will have the same number of lines. We want to include all of the lines, including the NULL frames, because we will need to iterate through every frame.

isis/src/base/apps/framestitch/framestitch.cpp
jessemapel: You can't stitch together two images if you have two different frame sizes as the frame boundaries won't line up.

oleg-alexandrov: Then finding the median is not helpful as all values should be equal. Instead, when, for a given even or odd cube the height of each frame is found, it should be verified right there that all obtained frames in a cube have the same height, before comparing the frame heights between the cubes.

isis/src/base/apps/framestitch/framestitch.cpp
oleg-alexandrov: I see. So then the frame size includes both valid and invalid lines. I think that was not clear from the comment on top of that function. I assumed only valid lines are counted.

isis/src/base/apps/framestitch/framestitch.cpp
jessemapel: I went with the median to account for any weird situations where there's random lines of NULL in the image that are not associated with missing even/odd frames. I could see just validating in the counting function being better. I'll switch to that. If there's a situation where there are random all NULL lines, then the user can just manually enter the frame size.

oleg-alexandrov: Sounds good. That was more for my own understanding.

jessemapel:
Updated the docs, re-worked the error handling in the automatic frame height computations, and replaced all instances of frame size with frame height to be more clear.

oleg-alexandrov (dismissed):

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -0,0 +1,221 @@
...
+    }
+    if (frameHeights.Minimum() != frameHeights.Maximum()) {
+      QString msg = "Found different frame heights between [" + toString((int)frameHeights.Minimum())
+                  + "] and [" + toString((int)frameHeights.Maximum()) + "] lines in cube ["
+                  + cube->fileName() + "]. Please manually enter the frame height.";
+      throw IException(IException::User, msg, _FILEINFO_);
+    }
+
+    // Use the median to handle any outliers
+    return frameHeights.Average();
```
AustinSanders, line 74: Should this be average or median?

isis/src/base/objs/ProcessByBrick/ProcessByBrick.cpp
```diff
@@ -660,6 +660,82 @@ void ProcessByBrick::SetOutputRequirements(int outputRequirements) {
     omgrs.clear();
   }
 
+
+  /**
+   * Starts the systematic processing of the input cube by moving an arbitrarily-shaped
+   * brick through the cube. This method allows multiple input and output
+   * cubes.
+   *
+   * @deprecated Please use ProcessCubeInPlace, ProcessCube, or ProcessCubes
```
AustinSanders: Are we adding deprecated functionality here?

isis/src/base/apps/framestitch/framestitch.cpp
jessemapel: I switched to a statistics object instead of a histogram and forgot to update the comment. It doesn't matter because there should only be 1 value accumulated by the time we get here

isis/src/base/objs/ProcessByBrick/ProcessByBrick.cpp
jessemapel: These deprecated tags need to be removed, the functional signatures are brand new.

oleg-alexandrov (approved):
I have no further comments. Whenever the other reviewer approves this can be pulled in.

AustinSanders (approved):

PR #4950, opened 2022-05-13
Incremental bugfix for stitching WAC framelets

oleg-alexandrov:
## Description
This is an incremental improvement upon the issues noted in https://github.com/USGS-Astrogeology/ISIS3/issues/4944 and https://github.com/USGS-Astrogeology/ISIS3/issues/4945, which state that each WAC framelet has the same data in its last several rows as the earliest rows in the next framelet, and moreover, some of those rows have pixel values flagged as invalid (null) even though they look reasonably valid.

The proposed fix wipes several top and bottom lines in each framelet before stitching. With this fix, the *framestitch* tool is called as:

framestitch even = img.even.cub odd = img.odd.cub to = stitched_img.cub flip = true num_lines_overlap = 2

With this choice of num_lines_overlap, 1 line is wiped at the top and 1 at the bottom of each framelet. 

Here's an illustration of the stitched image before (left) and after (right) this fix. Note how black pixels and repeated lines go away.

framestitch

The camera model is adjusted accordingly (https://github.com/USGS-Astrogeology/usgscsm/pull/397).

## How Has This Been Tested?

The produced images were tested very carefully with stereo in ASP and the resulting terrain models have a lot less artifacts. The triangulated rays intersect better as well, hence the intersection error map has lower values. The image-to-ground and ground-to-image operations after this fix agree better than before (see the CSM pull request above).

This does not eliminate the problem fully. Framelets can overlap by more than 2 pixels, depending on the image, and the amount of overlap is likely terrain-dependent (lower-lying terrain should result in more overlap). 

For the purposes of stereo it is likely best to first mapproject the obtained stitched images. This reduces any remaining repetition of rows even further. This was tried and works decently, but some residual artifacts still can be seen, though things improve quite a bit in that mode as well. 

More work and research needs to be done here but there's no more time in this cycle. 

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/framestitch/framestitch.cpp
+++ b/isis/src/base/apps/framestitch/framestitch.cpp
@@ -122,11 +122,6 @@ namespace Isis {
       }
     }
 
-    QString outCubeFile = ui.GetCubeName("TO");
-    Cube *outCube = process.SetOutputCube(
-          outCubeFile, CubeAttributeOutput(outCubeFile),
-          evenCube->sampleCount(), evenCube->lineCount(), evenCube->bandCount());
-
     int frameHeight = 1;
     if (ui.WasEntered("frameHeight")) {
       frameHeight = ui.GetInteger("FRAMEHEIGHT");
@@ -144,15 +139,23 @@ namespace Isis {
 
       frameHeight = evenFrameHeight;
     }
+    int numFrames = evenCube->lineCount() / frameHeight;
 
-    int numLinesOverlap = 2; // The default value
+    int numLinesOverlap = 0; // The default value
     if (ui.WasEntered("NUM_LINES_OVERLAP"))
       numLinesOverlap = ui.GetInteger("NUM_LINES_OVERLAP");
 
-    if (numLinesOverlap %2 != 0 || numLinesOverlap < 0) {
+    if (numLinesOverlap % 2 != 0 || numLinesOverlap < 0) {
       QString msg = "Expecting a non-negative and even value for NUM_LINES_OVERLAP.";
       throw IException(IException::User, msg, _FILEINFO_);
     }
+
+    int reducedFrameHeight = frameHeight - numLinesOverlap;
+    
+    QString outCubeFile = ui.GetCubeName("TO");
+    Cube *outCube = process.SetOutputCube(
+          outCubeFile, CubeAttributeOutput(outCubeFile),
+          evenCube->sampleCount(), numFrames * reducedFrameHeight, evenCube->bandCount());
     
     // If there's an even number of frames and the inputs are flipped, we have to
     // swap even and odd because the first frame in the even cube is valid and the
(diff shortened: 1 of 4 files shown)


isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -189,34 +201,81 @@ namespace Isis {
...
+      srcBuff.SetBasePosition(1, src + 1, 1);
+      dstBuff.SetBasePosition(1, dst + 1, 1);
+
+      outCube->read(srcBuff);  // read into srcBuf
+      dstBuff.Copy(srcBuff);   // copy from srcBuff to dstBuff
+      outCube->write(dstBuff); // write into outCube
+    }
+    
+    // Wipe the last lines of the image after they are transferred
+    // TODO: Figure out how to resize the cube.
```
jessemapel: The resize needs to be done before any data is copied to it. Ideally this is done when we call process.SetOutputCube

jessemapel (commented):
A few things:

1. This needs to be added to the tests
2. The cube needs to be properly resized. The simplest way to do this is to create a temp cube that is deleted on completion and copy the raw frames into that; then create the true output cube and copy the trimmed frames into this.
3. The number of lines trimmed out should be added to the image label next to the information about flipping. This way we can pull it into the CSM model later.

isis/src/base/apps/framestitch/framestitch.cpp
oleg-alexandrov: When it comes to tests, will it be enough to just tweak the tests you added when this tool was added? I did not work with ISIS tests before. I assume there's no need for me to add new files, but likely just some more lines of code where appropriate in the "Fixtures" code. Is that correct?

If the cub file is resized, the number of lines in it won't correspond to what the CSM model thinks the number of lines is. Is that a problem? Likely not, but may be confusing. In the CSM model there is logic for having the reduced number of lines, which would agree with the number or actual lines in the cub file. 

> The number of lines trimmed out should be added to the image label next to the information about flipping. This way we can pull it into the CSM model later.

OK. This will need some logic changes in the CSM model as well then. I will get to this tomorrow.

jessemapel: The existing tests will need updating because of the default 2 line trim breaking them. You'll also need to add a test that sets the trimming and then checks that it's doing the correct thing.

I don't think the fixtures will need to be updated as they handle input data. These tests are where the updates will need to go https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/tests/FunctionalTestsFrameStitch.cpp

If you have any questions or anything let me know. I can get on a call and go over some stuff or we can just go back and forth here.

oleg-alexandrov: I fixed the tests, by setting to 0 the number of lines to trim by default, as for a general application that should be the case. For WAC no single number works all the time, it has to be 2 or 4 depending on the cub.

I added a test where the number of lines to trim is 2.

The cube is resized on writing, so no empty lines are written.

The number of lines that were trimmed are written to the output cube. Next step will be making changes to ALE and do CSM to query this. 

I tested this very carefully with two cub files and the result is as before, except the intended changes as above.

oleg-alexandrov: Here's a problem though. This tool will write the number of images lines to be trimmed in the stitched cube. But later, when ALE reads in this cube, it won't have kernels. So ALE needs to read a cube before being stitched, but that won't have the number of image lines to trimp. 

Can framestitch write the kernels so it is an honest cube?

jessemapel: I strip the kernels off because the stitched image is no longer valid for the ISIS camera model. I could try something else to prevent it from creating a camera model.

isis/src/base/apps/framestitch/framestitch.xml
```diff
@@ -127,6 +132,30 @@
...
+          <p>
+          The estimated total number of lines in each frame which have
+          the same pixel values as frames above and below it. During
+          processing, half this amount of lines will be removed from
+          each of the top and bottom of the frame. Must be an even
+          non-negative number.
+          </p>
+          <p>
+            If not entered, the program will assume the value 0. A value
+			of 2 or 4 works reasonably for WAC images.
```
jessemapel: for LRO WAC images

isis/src/base/apps/framestitch/framestitch.xml
```diff
@@ -127,6 +132,30 @@
...
+          </p>
+        </description>
+        <internalDefault>Camera</internalDefault>
```
jessemapel: Change this to an actual default of 0

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -144,7 +139,24 @@ namespace Isis {
 
       frameHeight = evenFrameHeight;
     }
+    int numFrames = evenCube->lineCount() / frameHeight;
+
+    int numLinesOverlap = 0; // The default value
+    if (ui.WasEntered("NUM_LINES_OVERLAP"))
+      numLinesOverlap = ui.GetInteger("NUM_LINES_OVERLAP");
```
jessemapel, line 146: If you set an actual default of 0 for NUM_LINES_OVERLAP you don't need to worry about the default here or checking if it was entered.

isis/src/base/apps/framestitch/framestitch.cpp
```diff
@@ -144,7 +139,24 @@ namespace Isis {
...
+
+    if (numLinesOverlap % 2 != 0 || numLinesOverlap < 0) {
```
jessemapel, line 148: Same here, the minimum on the UI already enforces >= 0

oleg-alexandrov: I think such redundant checks give the reader the impression that a thorough job was undertaken in validating the inputs.

isis/src/base/apps/framestitch/framestitch.cpp
oleg-alexandrov: Same as below, having extra redundant checks likely can't hurt.

oleg-alexandrov: I made the default to be 0 and tested that the tool can handle being run with this value not explicitly set when it then uses the default.

isis/src/base/apps/framestitch/framestitch.xml
oleg-alexandrov: I made the default to be 0 and tested that the tool can handle being run with this value not explicitly set when it then uses the default.

isis/src/base/apps/framestitch/framestitch.xml
oleg-alexandrov: Changed.

oleg-alexandrov:
Issues addressed.

jessemapel (approved):

jessemapel:
Running CI just to double check then this is good to go

PR #4968, opened 2022-05-31
added tests for importing with new eis template

gsn9:
Created tests for testing Isisimport with the EIS template. 

## Description

Added needed XML and IMG files to tests/data/.. for testing.
Added 4 tests to cover the new template in #4965

## Types of changes

## Checklist:

## Licensing


isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+    Object = OriginalXmlLabel
+      Name      = IsisCube
+      StartByte = 10297634
+      Bytes     = 13238
+      ByteOrder = Lsb
+    End_Object
+    End
+  )");
+  QString dataFileName = " data/eis2isis/nacFrame/nac000xxx_2022145t000000_0000000001_frame_raw02.xml";
```
scsides: Shouldn't these files be in a dir named isisimport rather than eis2isis? Hopefully eis2isis will go away in the near future.

scsides (dismissed):
The file name comment goes for the other data/?????/ files as well. 

@jessemapel, can you think of a way to create the IMG files on the fly? It would be somewhat convoluted, but we could do it with "makecube" with detached labels, and a rename.

isis/tests/FunctionalTestsIsisImportEis.cpp
jessemapel: I agree, we should call the directory something more generic. Maybe EISPDS

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,375 @@
...
+  QString templateFile = "../appdata/import/PDS4/ClipperEIS.tpl";
+  QString cubeFileName = tempDir.path() + "/nacFrame.cub";
+  QVector<QString> args = {"from=" + dataFileName, "to=" + cubeFileName, "template=" + templateFile};
```
jessemapel: We should test both with the direct template call and without.

jessemapel (commented):
I agree the IMG files are too large to just add to the repo. We should either make them very small and adjust the XML labels appropriately or generate them dynamically. One way to generate them dynamically would be to copy the xml label into a temp directory and then just write 0s to an IMG file in the temp directory. We could also modify the XML files to point to Dev 0 for the img file.

isis/tests/FunctionalTestsIsisImportEis.cpp
gsn9: @jessemapel Can you expand on this a little more?

jessemapel: We should run isisimport with the template parameter and without the template parameter to check that the automatic template selection is working with EIS. So with it would be like it is right now. Without, it would be `QVector<QString> args = {"from=" + dataFileName, "to=" + cubeFileName};`. Then we would just check that against the result for with.

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -83,10 +86,94 @@ TEST_F(TempTestingFiles, FunctionalTestIsisImportEisNacFrame){
...
-  QString dataFileName = " data/eis2isis/nacFrame/nac000xxx_2022145t000000_0000000001_frame_raw02.xml";
+  QString dataFilePath= " data/isisimport/eispds/nacFrame/nac000xxx_2022145t000000_0000000001_frame_raw02.xml";
+  QString dataFileName = "nac000xxx_2022145t000000_0000000001_frame_raw02.xml";
+  QString imageFilePath = " data/isisimport/eispds/nacFrame/nac000xxx_2022145t000000_0000000001_frame_raw02.img";
+  QString imageFileName = "nac000xxx_2022145t000000_0000000001_frame_raw02.img";
   QString templateFile = "../appdata/import/PDS4/ClipperEIS.tpl";
   QString cubeFileName = tempDir.path() + "/nacFrame.cub";
-  QVector<QString> args = {"from=" + dataFileName, "to=" + cubeFileName, "template=" + templateFile};
+
+  // hard coding these values for now, but we should use some sort of xml parser to grab these values from the labels xml
```
jessemapel: I don't know if this needs to be dynamically parsed out of the XML. We also have them hard coded into the truth label.

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -83,10 +86,94 @@ TEST_F(TempTestingFiles, FunctionalTestIsisImportEisNacFrame){
...
+  int samples = 3832;
+  int lines = 1335;
+  int bytes = 2;
+
+  // create a temp img file and write data to it
+  QFile tempImgFile(tempDir.path() + "/" + imageFileName);
+  if(!tempImgFile.open(QFile::WriteOnly | QFile::Text)){
+      qDebug() << " Could not open file for writing";
```
jessemapel: fail here if we can't open the file. Just change qDebug() to FAIL()

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -83,10 +86,94 @@ TEST_F(TempTestingFiles, FunctionalTestIsisImportEisNacFrame){
...
+      return;
+  }
+  QTextStream out(&tempImgFile);
+  char writeToFile[samples * bytes];
+  for(int i=0; i<samples * bytes; i++){
+    writeToFile[i] = '0';
```
jessemapel: I don't think we want to write text '0's to the file. Rather just 0 bytes. You should be able to use a a [qDataStream](https://doc.qt.io/qt-6/qdatastream.html) to do this.

jessemapel: If you really want to properly test the data ingest, then you could write MSB short ints to the datastream. I don't know if that' actually needed though.

jessemapel (commented):
It looks like you axed the xml labels too

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,579 @@
...
+  QString cubeFileName = tempDir.path() + "/nacFrame.cub";
+
+  int samples = 1335;
+  int lines = 3832;
+  int bytes = 2;
+
+  // create a temp img file and write data to it
+  QFile tempImgFile(tempDir.path() + "/" + imageFileName);;
+
+  (tempDir.path() + "/" + imageFileName);
```
jessemapel: This looks extraneous

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,579 @@
...
+  QString imageFileName = "nac000xxx_2022145t000000_0000000001_frame_raw02.img";
+  QString templateFile = "../appdata/import/PDS4/ClipperEIS.tpl";
+  QString cubeFileName = tempDir.path() + "/nacFrame.cub";
+
+  int samples = 1335;
+  int lines = 3832;
+  int bytes = 2;
+
+  // create a temp img file and write data to it
+  QFile tempImgFile(tempDir.path() + "/" + imageFileName);;
```
jessemapel: extra semi-colon

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,579 @@
...
+  // generate lines 
+  QByteArray writeToFile = QByteArray();
+  short int fill = 0;
+  for(int i=-1; i<(samples * bytes); i++){
+    writeToFile.append(fill);
+  }
+
+  // write the lines to the temp file 
+  for(int i=0; i<lines; i++){
+    QTextStream out(&tempImgFile);
```
jessemapel: Change to a QDataStream and you won't need to fill a ByteArray. You can just pipe `fill` straight into the stream and it will go to the file.

Right now, this is getting converted to text https://doc.qt.io/qt-6/qtextstream.html#operator-lt-lt-15

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,579 @@
...
+          Type       = SignedWord
+          ByteOrder  = Lsb
+          Base       = 0.0
+          Multiplier = 1.0
+        End_Group
+      End_Object
+
+      Group = Instrument
+        SpacecraftName   = "Europa Clipper"
+        InstrumentId     = "EIS NAC PB"
```
jessemapel: This should be `EIS NAC FC` for framing camera

isis/tests/FunctionalTestsIsisImportEis.cpp
```diff
@@ -0,0 +1,579 @@
...
+        End_Group
+
+        Group = Pixels
+          Type       = SignedWord
+          ByteOrder  = Lsb
+          Base       = 0.0
+          Multiplier = 1.0
+        End_Group
+      End_Object
+    End_Object
```
jessemapel, line 422: Missing the Instrument, BandBin, and Kernel groups

jessemapel (commented):
The XML files appear to have been untracked and need to be added back.

jessemapel (approved):

Issue #4931, opened 2022-04-26
hical not falling back to ZeroDark when ZeroDarkRate is enabled but matrix file is not available.


rbeyer:
**ISIS version(s) affected**: 6.0.0

**Description**  
When the new ZeroDarkRate module is enabled for ISIS `hical` it uses a set of `DarkRate*.csv` files in $ISISDATA/mro/calibration/matrices for calculations.  These files have names like `DarkRate_RED2_1_TDI128_BIN4_ADC54_hical_0002.csv` and there could be a file for each combination of CCD, channel, TDI rate, binning value, and ADC setting (which is a potentially open-ended and infinite number of combinations).

However, if `hical` attempts to process an image for which the appropriate `DarkRate*.csv` does not exist, it errors.

**How to reproduce**  
1. First, set up appropriate .conf file to enable the ZeroDarkRate module, `hical-darktest.conf`.
2. Download http://hirise-pds.lpl.arizona.edu/PDS/EDR/ESP/ORB_020200_020299/ESP_020242_0945/ESP_020242_0945_RED0_1.IMG
3. Then:

```
$> hi2isis fr= ESP_020242_0945_RED0_1.IMG to= ESP_020242_0945_RED0_1.cub
$> hical conf= hical-darktest.conf fr= ESP_020242_0945_RED0_1.cub to= ESP_020242_0945_RED0_1.hical.cub
**USER ERROR** Failed in ZeroDarkRate module.
**ERROR** No existing files found with a numerial version matching [DarkRate_RED0_1_TDI128_BIN4_ADC54_hical_????.csv] in [/Users/rbeyer/opt/isisdata/mro/calibration/matrices].
```

**Possible Solution**  
When ZeroDarkRate is enabled and no appropriate `DarkRate*.csv` file can be found, fall back to ZeroDark behavior.

Alternately, leave as-is, since it is a legit error, given the specification, and leave the remedy up to the user to use just the ZeroDark behavior.

**Additional context**
Not all possible combinations of `DarkRate*.csv` exist because there is insufficient calibration data to create them.

AustinSanders:
I haven't yet replicated this issue, but I'm tracking this issue and plan to address this (or at least triage) with continuous support hours.

jlaura:
[...]

PR #4986, opened 2022-07-07
Force hical to fall back to ZeroDark if ZeroDarkRate fails on init

AustinSanders:
## Description
Added hical functionality to fall back to ZeroDark if ZeroDarkRate module fails on initialization.  

This logic is dependent on the configuration file, and can be toggled ON by specifying Fallback = True within the ZeroDarkRate profile.  By default (i.e. the parameter is not present in the config file) this functionality is turned off.  In this case, hical will fail, and an error message will be displayed.

## Related Issue
#4931

## Motivation and Context
Motivation and context provided in #4931 

## How Has This Been Tested?
Added a test and ran manually with 
``` hical conf= hical-darktest.conf fr= ESP_020242_0945_RED0_1.cub to= ESP_020242_0945_RED0_1.hical.cub ``` . hical-darktest.conf points to a file that is not on my system.

The first run of hical demonstrates default usage with no fallback parameter in the ZeroDarkRate profile.  The second run (after vim) demonstrates the new functionality.
hical

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/mro/apps/hical/hical.cpp
+++ b/isis/src/mro/apps/hical/hical.cpp
@@ -188,6 +188,7 @@ namespace Isis {
       else {
         hiconf.selectProfile("ZeroDarkRate");
         hiprof =  hiconf.getMatrixProfile();
+
         if ( !SkipModule(hiprof) ) {
           // Make sure we aren't applying ZeroDark and ZeroDarkRate
           if ( !SkipModule(hiconf.getMatrixProfile("ZeroDark")) ){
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by acpaquette** [comment]
With the discussion happening in #4931 I'm not sure how this work should proceed. It seems like making a parameter that enables fall back would be ideal with some error/"warning" message in two cases:

First, if fallback is not enable, throw an error that the image is not able to be calibrated.

Then, if fallback is enabled, just add a warning to the output of the program to tell the user that the image is falling back to the ZeroDark file. That just seems to be the best option and keeps the work that has been done.

File: isis/src/mro/apps/hical/hical.cpp [hunk 1]
```diff
@@ -217,11 +199,42 @@ namespace Isis {
...
+              zdr.Dump(hiconf.getMatrixSource("DumpModuleFile",hiprof));
+            }
+          }
+          catch(IException e){
+            if (hiprof.exists("Fallback")){
+              if (IsTrueValue(hiprof, "Fallback")){
+                zdrFallback = true;
+                calVars->add(hiconf.getProfileName(), HiVector(nsamps, 0.0));
+                std::cout << "Falling back to ZeroDark implementation. Unable to initialize ZeroDarkRate "
+                          << "module with the following error:" << std::endl << e.what() << "\nContinuing..."<< std::endl;
```
Kelvinrr: A warning might still want to be printed via cerr. Or maybe [qWarning](https://doc.qt.io/qt-6/qtglobal.html#qWarning)()?

Kelvinrr (commented):
simple comment in line

isis/src/mro/apps/hical/hical.cpp
```diff
@@ -217,11 +199,31 @@ namespace Isis {
...
+            if ( hiprof.exists("DumpModuleFile") ) {
+              zdr.Dump(hiconf.getMatrixSource("DumpModuleFile",hiprof));
+            }
+          }
+          catch(IException e){
+            if (hiprof.exists("Fallback") && IsTrueValue(hiprof, "Fallback")){
+                zdrFallback = true;
+                calVars->add(hiconf.getProfileName(), HiVector(nsamps, 0.0));
+                std::cerr << "Falling back to ZeroDark implementation. Unable to initialize ZeroDarkRate "
+                          << "module with the following error:" << std::endl << e.what() << "\nContinuing..."<< std::endl;
```
acpaquette, line 215: Typically we would throw a "Warning" in the logging pvl output, not sure if that is something we want to do here

acpaquette (commented):
One comment to address, other than that this looks good!

isis/src/mro/apps/hical/hical.cpp
AustinSanders: I wasn't sure about that either.  I wanted to print something directly to the user so that it was immediately evident that the process wasn't _exactly_ what they specified in the config -- IMO the user shouldn't have to access logs to learn that they got ZeroDark instead of ZeroDarkRate.

I think logging might still be a good addition, my only question is: is it overkill to include a warning message, a warning log, and a history message for the same condition?

acpaquette: I think as is right now is good. If we want to throw it in the actual app logs later on I don't think that will be very hard

acpaquette (approved):

Issue #4943, opened 2022-05-06
Segfault when footprintinit with Phobos DSK


scsides:
**ISIS version(s) affected**: 7.0

**Description**  

Many Mex HRSC SRC images segfault in footprintinit when using the Phobos DSK.
Segfault looks like it is in: NaifDskShape::setLocalNormalFromIntercept()

Image: HC545_0031_SR2.lev0.cub
spiceinit from=HC545_0031_SR2.lev0.cub tspk='(base/kernels/spk/de405.bsp, tspk/MAR097.bsp)' IK=MEX_HRSC_V09.ti extra=spk/MEX_STRUCT_V01.bsp shape=user model=dsk/PHOBOS_K137_DLR_V01.bds iak=iak/hrscsrcAddendum002.ti

This example uses some kernels from the mission team, and the modified camera model from PR #4910. All are documented as better than the defaults stored in ISISDATA

**How to reproduce**  

footprintinit from=HC545_0031_SR2.lev0.cub

**Possible Solution**  

**Additional context**  

The camera model has been checked to see if the changes caused this issue. The model appears to be working correctly.
The pointing for all of the images of Phobos is a little off, but that's nothing out of the ordinary.
footprintinit also fails if Bullet is used as the ray trace engine, but in a different way

scsides:
@lwellerastro, FYI this is the new issue for Mex HRSC SRC

PR #4997, opened 2022-07-15
Added intersectSurface to NaifDskShape so the intercept point would b…

scsides:
Override NaifDskShape::intersectSurface so the intercept is calculated.

## Description

footprintinit segfaulted when setGround was called on the subSpacecraftPoint because the intercept destructor had been called after exiting a scope

## Related Issue
#4943

## Motivation and Context
Stop the segfault

## How Has This Been Tested?
footprintinit no longer segfaults and a test was added to make sure the intercept was being calculated

## Types of changes

## Checklist:

- [X] I have added any user impacting changes to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document.

## Licensing


acpaquette (dismissed):
Overall this looks good to me but having @KrisBecker and @jessemapel review would be ideal

isis/src/base/objs/NaifDskShape/unitTest.cpp
```diff
@@ -80,6 +80,18 @@ int main(int argc, char *argv[]) {
     success = shapeModelFromPvlShape.intersectSurface(obsPos, lookDir);
     qDebug() << "Intersection successful?     " << toString(success);
 
+    qDebug() << "Try to intersect surface at obsPos (1000,0,0) and lookDir (-1,0,0)";
+    obsPos[0]  = 1000.0;  obsPos[1]  = 0.0;   obsPos[2]  = 0.0;
+    lookDir[0] = -1.0;  lookDir[1] = 0.0;  lookDir[2] = 0.0;
+    success = shapeModelFromPvlShape.intersectSurface(obsPos, lookDir);
```
jessemapel: This is calling the wrong intersectSurface function, it should call the ground point + observer position version.

jessemapel (commented):
I worked with Stuart on the change, so I don't have any input on the functional change. Just the test change.

KrisBecker (dismissed):
This code is difficult to follow, top to bottom.

This looks like a good addition. Note that it costs two ray traces in this mod. One to determine the local radius at the lat/lon location and then a ray trace from observer to the surface point. That looks like a total of three ray traces/point so expect it to be a bit slow.

An alternative would be to call [ShapeModel::intersectSurface(lat,lon,obs,backcheck)](https://github.com/USGS-Astrogeology/ISIS3/blob/d8092c11645cab74c017e51c8c43ccb37db9bedc/isis/src/base/objs/ShapeModel/ShapeModel.cpp#L86-L108). This costs one ray trace but may result in a different intercept. Not sure it would be a different intercept in this context, but you would not need this update for this approach. Though, it would require a modification to [Camera::SetGround()](https://github.com/USGS-Astrogeology/ISIS3/blob/d8092c11645cab74c017e51c8c43ccb37db9bedc/isis/src/base/objs/CameraGroundMap/CameraGroundMap.cpp#L68-L100). 

Also note that the back projection check is ignored in this implementation. It may not be returning the same/valid intercept point as the surface point because of occlusion.

isis/src/base/objs/NaifDskShape/unitTest.cpp
scsides: Called other function

jessemapel (dismissed):

jessemapel (approved):

PR #5072, opened 2022-11-10
CSM Camera Skymap

acpaquette:
Attempt at getting the CSM camera to work with `skymap`

## Description
These changes are related to getting the MSL Mastcam data projected using ISIS but met a hard block once the `setSky` function was hit. Within ISIS this uses the camera look vector to generate a focalplaneX/Y from a line/sample but the CSM doesn't support data within the camera reference frame.

There is still room for experimentation but it seems like we will need a specific program for creating projections from rover data.

## Related Issue
Ground based sensor work (See USGSCSM IAA Proposal)

Has not been validated as the changes are incomplete

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@ release.
 
 ## [Unreleased]
 
+### Added
+
+- CSMCamera can now read and use the body rotation from ALE produced ISDs [#5072](https://github.com/DOI-USGS/ISIS3/pull/5072)
+- CSMSkyMap added to CSMCamera for use with local rover projections in ISIS [#5072](https://github.com/DOI-USGS/ISIS3/pull/5072)
+
 ## [8.1.0] - 2023-12-05
 
 ### Changed
--- a/isis/src/base/apps/csminit/csminit.cpp
+++ b/isis/src/base/apps/csminit/csminit.cpp
@@ -29,8 +29,10 @@ find files of those names at the top level of this repository. **/
 #include "Pvl.h"
 #include "PvlGroup.h"
 #include "PvlKeyword.h"
+#include "SpiceRotation.h"
 
 using namespace std;
+using json = nlohmann::json;
 
 namespace Isis {
 
@@ -424,27 +426,32 @@ namespace Isis {
     if (ui.WasEntered("ISD")) {
       QString isdFilePath = ui.GetFileName("ISD");
       std::ifstream isdFileStream(isdFilePath.toStdString());
-      json isd = json::parse(isdFileStream);
+      try {
+        json isd = json::parse(isdFileStream);
 
-      if (isd.contains("body_rotation")) {
-        int bodyFrameCode = isd["body_rotation"]["time_dependent_frames"][0];
-        if (isd["body_rotation"].contains("constant_frames")) {
-          bodyFrameCode = isd["body_rotation"]["constant_frames"][0];
+        if (isd.contains("body_rotation")) {
+          int bodyFrameCode = isd["body_rotation"]["time_dependent_frames"][0];
+          if (isd["body_rotation"].contains("constant_frames")) {
+            bodyFrameCode = isd["body_rotation"]["constant_frames"][0];
+          }
+
+          SpiceRotation bodyRotation(bodyFrameCode);
+          bodyRotation.LoadCache(isd["body_rotation"]);
+
+          Table bodyTable = bodyRotation.Cache("BodyRotation");
+          bodyTable.Label() += PvlKeyword("Description", "Created by csminit");
+          bodyTable.Label() += PvlKeyword("SolarLongitude",
+            "10"/**toString(spice.solarLongitude().degrees())**/ );
+          cube->write(bodyTable);
+          if (isd.contains("naif_keywords")) {
+            json aleNaifKeywords = isd["naif_keywords"];
+            PvlObject naifKeywords = PvlObject("NaifKeywords", aleNaifKeywords);
+            *cube->label() += naifKeywords;
+          }
         }
-
-        SpiceRotation bodyRotation(bodyFrameCode);
-        bodyRotation.LoadCache(isd["body_rotation"]);
-
-        Table bodyTable = bodyRotation.Cache("BodyRotation");
-        bodyTable.Label() += PvlKeyword("Description", "Created by csminit");
-        bodyTable.Label() += PvlKeyword("SolarLongitude",
-          "10"/**toString(spice.solarLongitude().degrees())**/ );
-        cube->write(bodyTable);
-        if (isd.contains("naif_keywords")) {
-          json aleNaifKeywords = isd["naif_keywords"];
-          PvlObject naifKeywords = PvlObject("NaifKeywords", aleNaifKeywords);
-          *cube->label() += naifKeywords;
-        }
+      }
+      catch (json::parse_error &e) {
+        // continue if we hit a parse error
       }
     }
 
--- a/isis/src/base/apps/csminit/csminit.cpp
+++ b/isis/src/base/apps/csminit/csminit.cpp
@@ -417,6 +417,37 @@ namespace Isis {
     blobLabel += PvlKeyword("PluginName", pluginName);
     cube->write(csmStateBlob);
 
+    // Try to attach rotations
+    // If a camera fails to generate the original BodyRotation
+    // table will overwrite this one (if there was an original
+    // BodyRotation table)
+    if (ui.WasEntered("ISD")) {
+      QString isdFilePath = ui.GetFileName("ISD");
+      std::ifstream isdFileStream(isdFilePath.toStdString());
+      json isd = json::parse(isdFileStream);
+
+      if (isd.contains("body_rotation")) {
+        int bodyFrameCode = isd["body_rotation"]["time_dependent_frames"][0];
+        if (isd["body_rotation"].contains("constant_frames")) {
+          bodyFrameCode = isd["body_rotation"]["constant_frames"][0];
+        }
+
+        SpiceRotation bodyRotation(bodyFrameCode);
+        bodyRotation.LoadCache(isd["body_rotation"]);
+
+        Table bodyTable = bodyRotation.Cache("BodyRotation");
+        bodyTable.Label() += PvlKeyword("Description", "Created by csminit");
+        bodyTable.Label() += PvlKeyword("SolarLongitude",
+          "10"/**toString(spice.solarLongitude().degrees())**/ );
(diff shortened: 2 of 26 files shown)


isis/src/base/apps/csminit/csminit.cpp
```diff
@@ -417,6 +419,43 @@ namespace Isis {
     blobLabel += PvlKeyword("PluginName", pluginName);
     cube->write(csmStateBlob);
 
+    // Try to attach rotations
+    // If a camera fails to generate the original BodyRotation
```
AustinSanders: It's a little tough for me to understand what's being said here. Is this saying that if a camera isn't generated, then this uses the original BodyRotation table?

isis/src/base/apps/skymap/main.cpp
```diff
@@ -88,7 +88,9 @@ void IsisMain() {
   userGrp.addKeyword(PvlKeyword("EquatorialRadius", toString(1.0)), Pvl::Replace);
   userGrp.addKeyword(PvlKeyword("PolarRadius", toString(1.0)), Pvl::Replace);
   userGrp.addKeyword(PvlKeyword("LatitudeType", "Planetocentric"), Pvl::Replace);
```
AustinSanders, line 90: May be outside the scope of this PR, but do all the "replace" calls need to be conditional?

isis/src/base/apps/skymap/main.cpp
```diff
@@ -247,6 +252,13 @@ bool sky2map::Xform(double &inSample, double &inLine,
   if(p_incam->Sample() > p_inputSamples + 0.5) return false;
   if(p_incam->Line() > p_inputLines + 0.5) return false;
 
+  p_incam->SetImage(p_incam->Sample(), p_incam->Line());
+  
+  // Check that the new line and sample is projects close to the input Right Ascension 
+  // and Declination
+  if (abs(p_incam->RightAscension() - lon) > 0.1) return false;
```
AustinSanders, line 259: Where does 0.1 come from?

isis/src/base/apps/skymap/skymap.xml
```diff
@@ -48,6 +49,14 @@
     Finally, the user has the option to override the pixel resolution and the ground range (latitude/longitude).  These changes can be
     facilitied through the parameters SCALE, SRA, ERA, SDEC, and EDEC.
   </p>
+
+  <p>
+    In addition skymap can be used to produce landed mosaics from rovers like MSL. This can be done using ALE to generate an ISD
+    then that ISD can be attached to the cube via <def link="csminit">csminit</def>. The user would then have to update there map file
```
AustinSanders: there -> their

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -246,6 +270,107 @@ namespace Isis {
...
 
+  /**
+   * Given the ra/dec compute the look direction.
+   *
+   * @param ra    Right ascension in degrees (sky longitude).
+   * @param dec   Declination in degrees (sky latitude).
+   *
+   * @return @b bool True if successful.
+   */
+  bool CSMCamera::SetRightAscensionDeclination(const double ra, const double dec) {
```
AustinSanders, line 265: Does this actually set the RA/DEC anywhere, or just test if they're valid?

isis/src/base/objs/CSMCamera/CSMCamera.cpp
```diff
@@ -990,14 +1116,15 @@ namespace Isis {
    * positive east, ocentric).
    *
    * This is not supported for CSM sensors because we cannot get the position
-   * of the sun, only the illumination direction.
+   * of the sun, only the illumination direction. To prevent Qview from erroring
+   * this function returns 0, 0
```
AustinSanders: I think the wording in this documentation is a little misleading.  The function isn't "returning" values as much as setting lat/lon by reference.  Not sure how much difference it makes, though.

isis/src/base/objs/CSMSkyMap/CSMSkyMap.cpp
```diff
@@ -0,0 +1,45 @@
+/** This is free and unencumbered software released into the public domain.
```
AustinSanders, line 1: Do we want to merge #5069 before this PR so that we can get a clean file history?  Not sure how we want to handle the similar PRs.

isis/src/base/apps/csminit/csminit.cpp
acpaquette: Somewhat, we need to pull the body rotation into ISIS to compute the `RightAscension`/`Declination`

isis/src/base/apps/skymap/main.cpp
acpaquette: This was a limitation in skymap. Basically all skymap'ed images were only allowed to be PositiveWest. To get the mosaic correct in MSL I had to be able to set the Longitude direction to PositiveEast

isis/src/base/apps/skymap/main.cpp
acpaquette: It is a fairly minimal check to make sure that back projecting works, and removes the duplicate image that appears at the inverse lat lon. Basically if you have a 0-360 image set, an image at lon 90 would also appear at 180 without this check

isis/src/base/objs/CSMCamera/CSMCamera.cpp
acpaquette: In typical ISIS fashion, this takes an RA and DEC and sets the look direction but rather than call it `SetLookDirection` we call it `SetRAandDEC`

isis/src/base/objs/CSMSkyMap/CSMSkyMap.cpp
acpaquette: I think we can merge this in favor of #5069, this PR should be a super set of that PR

isis/src/base/apps/skymap/main.cpp
AustinSanders: I understand the check, I don't understand the reason for this particular threshold, i.e. is 0.1 appropriate?  0.1 in terms of lat/lon seems fairly course-grained, but if it works, then I have no issue with it 🤷‍♂️

acpaquette: I think if we go too course grain we end up not being able to project at all. I could try something like 0.01 or 0.001 but really we just need to make sure we are not on the opposite side of the projection

AustinSanders (dismissed):

AustinSanders (commented):
I think this needs a changelog entry and then we're good to go

AustinSanders (dismissed):

AustinSanders (dismissed):

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@ release.
 
 ## [Unreleased]
 
+### Added
+
+- CSMCamera can now read and use the body rotation from ALE produced ISDs [#5072](https://github.com/DOI-USGS/ISIS3/pull/5072)
+- CSMSkyMap added to CSMCamera for use with local rover projections in ISIS [#5072](https://github.com/DOI-USGS/ISIS3/pull/5072)
```
Kelvinrr, line 41: We don't normally allow more than one changelog but this is an old PR so we can probably let it slide.

isis/src/base/apps/csminit/csminit.cpp
```diff
@@ -417,6 +419,43 @@ namespace Isis {
...
+            bodyFrameCode = isd["body_rotation"]["constant_frames"][0];
+          }
+
+          SpiceRotation bodyRotation(bodyFrameCode);
+          bodyRotation.LoadCache(isd["body_rotation"]);
+
+          Table bodyTable = bodyRotation.Cache("BodyRotation");
+          bodyTable.Label() += PvlKeyword("Description", "Created by csminit");
+          bodyTable.Label() += PvlKeyword("SolarLongitude",
+            "10"/**toString(spice.solarLongitude().degrees())**/ );
```
Kelvinrr: weird commented out line

Kelvinrr (commented):
I think we also need tests for the new CSMCamera stuff?

isis/src/base/apps/csminit/csminit.cpp
acpaquette: This was leftover from the original effort to add the body table. Since we don't have a sun position in CSM we can't get the solar longitude unless we extract the sun position table onto the cube like we did with the body rotation. It may make more sense to leave the longitude at 0 instead of 10

acpaquette:
@Kelvinrr Tests are present, I will double check add potentially add some for coverage

isis/src/base/apps/csminit/csminit.cpp
acpaquette: From some final rounds of testing, we don't need to rotate the look vector into the body fixed reference frame so this was remove

Kelvinrr (dismissed):

Kelvinrr:
`skymap_app_test_user` seems to be failing

isis/src/base/apps/skymap/main.cpp
```diff
@@ -247,6 +252,13 @@ bool sky2map::Xform(double &inSample, double &inLine,
   if(p_incam->Sample() > p_inputSamples + 0.5) return false;
   if(p_incam->Line() > p_inputLines + 0.5) return false;
 
+  p_incam->SetImage(p_incam->Sample(), p_incam->Line());
+  
+  // Check that the new line and sample projects close to the input Right Ascension 
+  // and Declination
+  if (abs(p_incam->RightAscension() - lon) > 0.1) return false;
+  if (abs(p_incam->Declination() - lat) > 0.1) return false;
```
acpaquette, line 260: @Kelvinrr These lines were added to avoid dual projection, changing the values in the skymap test. Basically any look vector projects into two points in right ascension and declination, so if you project to 0, 0 you also project onto 0, 180. This checks avoids look vectors projecting into the inverse right ascension, declination

Kelvinrr: Did you update the test data? We just got to make sure we aren't adding another failing test if we dont have to.

acpaquette: I have updated data, we usually check it in once the PR is merged

Kelvinrr (approved):

isis/src/base/apps/skymap/main.cpp
Kelvinrr: kk just making sure

PR #5086, opened 2022-11-28
Adds mapping group to automos track cube

### LINKED ISSUE #4810: MAP2MAP with Tracking Layer Doesn't Work Properly
**ISIS version(s) affected**: 6.0.0-

**Description**  
After map-projecting a mosaic that has an additional _tracking.cub file, the _tracking.cub will not itself map-project to the new projection and QVIEW will not properly map back to it.

**How to reproduce**  
I was examining a mid-latitude mosaic that I had constructed with AUTOMOS with track=true.  Mid-latitudes in equirectangular projection have some significant distortions that make it hard to tell sometimes if the control was good.  I used MAP2MAP to convert my image mosaic to a Lambert Azimuthal with a local central lat/lon, which works much better for looking for anomalies.  I then tried to project the _tracking.cub layer with MAP2MAP but got an error that it couldn't be projected ("**ERROR** Unable to find PVL group [Mapping] in file [MC04_PDS202106_50mpp_tracking.cub].").  Meanwhile, loading the image mosaic in QVIEW, it seemed like it still knew where things were, but as I would move from one image that was part of the mosaic to another, it did not update properly.  I would move off the mosaic in the Lambert projection and it would still say I was on an input cube, leading me to realize that it was not doing some dynamic, on-the-fly projection of the tracking cube.

**Possible Solution**  
See next section.

**Additional context**  
With that all said, I don't know if this is a bug per se or simply a missing feature?  Perhaps there is a bug that the _tracking.cub _should_ include a mapping group in the header but it's not, so it's a problem with AUTOMOS?  Or perhaps the feature request is that it be included so MAP2MAP works on it?  Or perhaps the bug or feature is that QVIEW be able to dynamically search for lat/lon from the image cube and map that onto the tracking cube rather than do a line/sample lookup?

AustinSanders:
Adds mapping group to automos track cube to resolve #4810

## Description
Adds mapping information to track cube via processmosaic

## Related Issue
Fixes #4810 

Added a test and verified that track cube can be map projected

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/FunctionalTestsAutomos.cpp
+++ b/isis/tests/FunctionalTestsAutomos.cpp
@@ -142,6 +142,47 @@ TEST_F(DefaultCube, FunctionalTestAutomosPriority) {
   EXPECT_DOUBLE_EQ(oCubeStats->StandardDeviation(), 79.757668686375951);
 }
 
+TEST_F(DefaultCube, FunctionalTestAutomosTracking) {
+  QString outPath = tempDir.path() + "/mosaic.cub";
+  QString singleCubeListPath = tempDir.path() + "/newCubeList.lis";
+  QString trackingPath = tempDir.path() + "/mosaic_tracking.cub";
+
+  FileList singleCubeList;
+  singleCubeList.append(projTestCube->fileName());
+  singleCubeList.write(singleCubeListPath);
+
+  QVector<QString> args = {"fromlist=" + singleCubeListPath, "mosaic=" + outPath,
+                           "priority=band", "highsat=true", "lowsat=true", "null=true", "track=true"};
+
+  UserInterface options(APP_XML, args);
+
+  automos(options);
+
+  Cube mos(outPath);
+  Cube track(trackingPath);
+  Pvl label = *mos.label();
+  Pvl trackingLabel = *track.label();
+
+  PvlGroup mapping = label.findObject("IsisCube").findGroup("Mapping");
+  PvlGroup tMapping = trackingLabel.findObject("IsisCube").findGroup("Mapping");
+
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, (QString)mapping.findKeyword("ProjectionName"), (QString)tMapping.findKeyword("ProjectionName"));
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, (QString)mapping.findKeyword("TargetName"), (QString)tMapping.findKeyword("TargetName"));
+  EXPECT_EQ((double)mapping.findKeyword("UpperLeftCornerX"), (double)mapping.findKeyword("UpperLeftCornerX"));
+  EXPECT_EQ((double)mapping.findKeyword("UpperLeftCornerY"), (double)tMapping.findKeyword("UpperLeftCornerY"));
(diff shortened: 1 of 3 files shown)


isis/tests/FunctionalTestsAutomos.cpp
```diff
@@ -142,6 +142,47 @@ TEST_F(DefaultCube, FunctionalTestAutomosPriority) {
...
+  EXPECT_EQ((double)mapping.findKeyword("Scale"), (double)tMapping.findKeyword("Scale"));
+  EXPECT_EQ(mapping.findKeyword("LatitudeType")[0], tMapping.findKeyword("LatitudeType")[0]);
+  EXPECT_EQ(mapping.findKeyword("LongitudeDirection")[0], tMapping.findKeyword("LongitudeDirection")[0]);
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("EquatorialRadius"), (double)tMapping.findKeyword("EquatorialRadius"));
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("PolarRadius"), (double)tMapping.findKeyword("PolarRadius"));
+  EXPECT_EQ((int)mapping.findKeyword("LongitudeDomain"), (int)tMapping.findKeyword("LongitudeDomain"));
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("MinimumLongitude"), (double)tMapping.findKeyword("MinimumLongitude"));
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("MaximumLongitude"), (double)tMapping.findKeyword("MaximumLongitude"));
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("MinimumLatitude"), (double)tMapping.findKeyword("MinimumLatitude"));
+  EXPECT_DOUBLE_EQ((double)mapping.findKeyword("MaximumLatitude"), (double)tMapping.findKeyword("MaximumLatitude"));
```
acpaquette, line 183: Could this be replaced by the `AssertPvlGroupEqual` function? Should look something like:
`EXPECT_PRED_FORMAT2(AssertPvlGroupEqual, mapping, tMapping);`

acpaquette (changes requested):
Overall simple and straight forward. Small change could be made to how the pvl groups are being compared

isis/tests/FunctionalTestsAutomos.cpp
AustinSanders: It looks like AssertPvlGroupEqual performs a string compare, and it's complaining about things like 0 vs 0.0.  Any ideas to make the comparison work?

acpaquette: Unfortunately PVL in ISIS has no concept of typing in the comparison. The only way I could see making it work would be to modify the comparison function. 

First we try a string compare, if that fails we then compare the two values as doubles. I have tried to make this change before but you run into issues if the value of a keyword is an array.

We can waylay this for now, keep what we current PR is and just make an issue.

AustinSanders: Is this good to merge then?

acpaquette (approved):

Issue #5110, opened 2023-01-16
ISISDATA Evaluation and Verification


KrisBecker:
**ISIS version(s) affected**: All

# ISISDATA Evaluation and Verification

[Recent changes](https://astrodiscuss.usgs.gov/t/replacing-the-usgs-isis-rsync-server/924) to maintenance and installation procedures for the ISISDATA ancillary data area has resulted in noticeable impact on users. Of particular concern is the size of the ISISDATA installation has more than doubled rather suddenly. This is because of at least three reasons: 1) The OSIRIS-REx mission kernel set has been added to ISIS, 2) the way in which ISISDATA mission SPICE kernel archives are managed and downloaded has changed, and 3) moving the ISISDATA public download resource to Amazon's AWS S3 storage breaks symbolic links and results in additional copies of the linked file. In addition, there are reports of users that encounter problems processing ISIS mission data that are related to ISISDATA issues ([#5107](https://github.com/USGS-Astrogeology/ISIS3/issues/5107), [#5103](https://github.com/USGS-Astrogeology/ISIS3/issues/5103),[#5093](https://github.com/USGS-Astrogeology/ISIS3/issues/5093), [#5024](https://github.com/USGS-Astrogeology/ISIS3/issues/5024), etc...).

There are some efforts underway to address some of the issues related to ISISDATA. Because ISIS SPICE maintenance procedures have been changed to downloading and installing mission SPICE kernels directly from the NAIF SPICE archives (and other archives) without filter, the size of the ISIS installation has significantly increased in number of files and full installation size. I have developed a rather aggressive [rclone filter file](https://github.com/USGS-Astrogeology/ISIS3/issues/5024#issuecomment-1343764219) to exclude a rather large number of files that are not used by ISIS. There are numerous related issues ([#5109](https://github.com/USGS-Astrogeology/ISIS3/pull/5109), [#5014](https://github.com/USGS-Astrogeology/ISIS3/issues/5104), [#5105](https://github.com/USGS-Astrogeology/ISIS3/issues/5105)).

[...]

PR #5109, opened 2023-01-13
add filters for irrelevant files in ISISDATA

antonhibl:
## Description
added filters for the following files in the ISISDATA download script, these are being imminently removed:

/a_older_versions/*
/kernels/*/former_versions/
/corrupt_files/*
/zzarchive/**
/kernels/*/original
gallileo/kernels/ck/prime_mission//extended_mission/
gallileo/kernels/*/prime_mission/
galileo/kernels/ck/GEM/
clementine1/kernels/ck/save
chandrayaan1/kernels/spk/SAVE_SCS_2017-11-22
/kernels/fk/release./*
cassini/kernels/fk/Archive

## Related Issue
Data storage concerns in ISISDATA area

## How Has This Been Validated?
tested downloadIsisData script on my machine and it works fine with these changes applied

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation change (update to the documentation; no code change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Infrastructure change (changes to things like CI or the build system that do not impact users)

## Checklist:
- [x] I have read and agree to abide by the [Code of Conduct](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/Code-Of-Conduct.md)
- [x] I have read the [**CONTRIBUTING**](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CONTRIBUTING.md) document.
- [x] I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json) document.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

Issue #5105, opened 2023-01-10
Some way to generate a table of kernel sizes for the current ISIS Data Area. 


Kelvinrr:
**Description**

There seems to be some stuff being downloaded nightly that are either not kernels, in order to find every instance and make this easier to diagnose in the future, might be worth creating some summary of the sizes of every mission category to get an idea of what is happening at a glance that can be generated with the nightly downloads. This is something that can be documented internally in the isis3mgr scripts repo. 

**Suggested Edit**

Before trying to write some custom script to do this, AWS probably has something we can leverage via cloud watch and/or log groups to get an idea of what is happening nightly and what the changes are over time (https://docs.aws.amazon.com/efs/latest/ug/logging-monitoring.html). If it's "go to the AWS and press these buttons", it's still worth writing down.

foobarbecue:
I found `rclone ncdu` works really well for this. You can get something like this, and when you press enter on a row it goes into it and gives you sizes:

![image](https://user-images.githubusercontent.com/854789/224862014-d516f420-99e4-4012-80b9-2d252d7c27ad.png)

To use, set up your rclone config file so it looks like this:

```
(isis) aacurtis@MT-405357:~$ cat /home/aacurtis/.config/rclone/rclone.conf
[isis]
type = s3
provider = AWS
region = us-west-2
location_constraint = us-west-2
```
and then run ` rclone ncdu isis:asc-isisdata/usgs_data/`

Kelvinrr:
I have some work to redo the way S3 files are synced being done, we can incorporate that rclone command into those and publish them somehow.

PR #5111, opened 2023-01-16
isisdataeval - ISISDATA Evaluation and Verification

KrisBecker:
## Description

Motivation for this tool is to validate the ISISDATA installation on a local processing system. [Recent changes](https://astrodiscuss.usgs.gov/t/replacing-the-usgs-isis-rsync-server/924) to how the USGS provides ISISDATA to ISIS users separates the kernel configuration (USGS/AWS) from the kernel download resources (typically NAIF).

This can be a useful tool to quickly determine if <em>spiceinit</em> failures may be due to missing files or improper local ISIS installation. It can also be used to validate a local ISISDATA install where kernel file download filters have been applied to ISISDATA installation processing. For example, AWS S3 storage does not support symbolic links to files, so a great deal of effort has been made to identify and report these types of files (they lead to copies/redundancy of the linked file). When given the $ISISDATA directory (the default), this tool will inspect all files and directories contained in the installation for a valid ISISDATA configuration. As it pertains to **isisdataeval**, a valid ISISDATA installation is confirmed when all the contents of the kernel database (DB) and configuration files are translated into files that exist in the system. **isisdataeval** **does not**, however, confirm all necessary or required files have been downloaded to the local system, but it can be verified by comparing results of the application on each install of ISISDATA.

I apologize I do not have tests developed for this. Honestly, I am open for suggestions on how to test this application.

## Related Issue

Related Issues:
#5110
#5109 
#5105 

## Types of changes

## Checklist:

## Licensing


Kelvinrr:
Testing it locally, this might be useful to run during the nightly syncs to diagnose data issues.

KrisBecker:
My suggestion is to run this application after every ISISDATA update to your sources and make them available to users. Run the app on each $ISISDATA/{mission} directory independently and make them available as well.

This provides hashes for all data and each mission after every source sync and users can now check this resource and determine any timeline of ISISDATA integrity.

Some additional (perhaps Python) tools to run comparisons of the data would be super helpful. (NOTE the app contains the ability to generate a JSON file for output, too.)

isis/src/system/apps/isisdataeval/isisdataeval.xml
```diff
@@ -0,0 +1,867 @@
...
+      kernel DB files and evaluated as described below. Note these file names
+      are strongly recommended <b>not</b> to be of the generic form
+      <b>kernel.????.db</b> to minimize confusion and erroneous behavior.
+    </p>
+    <p>
+      After the configuration files are processed, a second search is made
+      for kernel DB files using the form <b>kernels.????.db</b>. The
+      highest version of these files are selected and its contents are read.
+      For every <b>File</b> specification in a <b>Selection</b> group,
+      the highested version is determined and then checked for existance.
```
Kelvinrr: highested -> highest

isis/src/system/apps/isisdataeval/isisdataeval.xml
```diff
@@ -0,0 +1,867 @@
...
+End_Group
+</PRE>
+          The files labeled <em>missing</em> are kernel files contained in
+          either a kernels.????.db or kernels.????.conf ISIS kernel database
+          configuration file. Files with <em>empty</em> status have no
+          content or do not specify any kernels. This may be needed to
+          satisfy spiceinit requirements. A status of <em>external</em>
+          indicate the file <em>as specified</em> in the db/conf file
+          does not exist in the ISISDATA inventory. For the occurances
+          in the above results indicate an extra '/' in the file name.
```
Kelvinrr: just to be clear, external means that it is a file NOT mentioned in any .db and therefore essential a dead kernel to ISIS?

Kelvinrr (commented):
some small comments for now. Still messing with it. 

For testing: If it is possible to test on a subset, we can test it by creating temporary kerne.db files from string literals to a temp dir similar to how we test ISDs. Then you can create empty kernel files to test things like missing, external etc. If it helps I put up boilerplate of what that would look like. But you can see some examples in [tests/Fixtures.cpp ](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/tests/Fixtures.cpp)

isis/src/system/apps/isisdataeval/isisdataeval.xml
KrisBecker: No. _external_ file status as determined in **isisdateaveal** has several meanings that may include:

- A file that is referred to in a kernel DB/conf file, but is not found in the full ISISDATA file inventory (it can still exist but the full path to the file must contain the expanded/specified $ISISDATA path or its deemed external). These situations are common when ISISDATA == DATADIR.
- A misspelled kernel file specification that contains an extra / in its file name (as seen in Voyager SPK kernel configs in #5110). The file is present and can be read in ISIS with the file specification, but it's still identified as an issue nonetheless.
- When you choose DATADIR as a mission directory in $ISISDATA, e.g., DATADIR=$ISISDATA/osirisrex, any file specified in a kernel DB or conf file that is not contained within the $ISISDATA/osirisrex directory hierarchy (i.e., _inventory_), is considered external. This identifies ISISDATA dependencies outside the mission directory even if the file is in $ISISDATA but not in DATADIR - a quite useful feature. And these cases may not indicate an error, but it is good to know from a management perspective (if your ISIS mission kernel spec has outside dependencies).

Note its difficult to make assumptions in **isisdataeval** regarding a dead file because all files in ISISDATA are considered inventory, including calibration files, which are typically not part of the kernel infrastructure.

isis/src/system/apps/isisdataeval/isisdataeval.xml
KrisBecker: I'll fix this, thanks.

isis/src/system/apps/isisdataeval/isisdataeval.xml
Kelvinrr: This seems kinda broad, does it make sense to make these different classifications? It seems useful to flag files that match case 2 and 3 specifically rather than being pooled together.

KrisBecker: What classification do you suggest?

Case 2 is difficult to identify specifically as a misspelled file in the config file and may have different OS behaviors. It is determined by string comparisons of the environment variable expansion and the result of absolute path by OS resources via [Qt::QFileInfo](https://doc.qt.io/qt-6/qfileinfo.html#absoluteFilePath) which seems to clean up the path.

I can add more documentation based upon this narrative if that will be helpful.

Kelvinrr: I think for now the extra text in the XML docs that breaks down these cases is good for now.

isis/tests/data/isisdata/mockup/hayabusa/kernels/spk/hay_osbj_050911_051118_v1n.bsp
```diff
@@ -0,0 +1,9 @@
+{
+    "source": "$ISISDATA/hayabusa/kernels/spk/hay_osbj_050911_051118_v1n.bsp",
+    "filesize": 81458176,
+    "createtime": "2022-08-12T22:29:59.000000",
+    "modifiedtime": "2022-11-23T23:37:37.651359",
+    "md5hash": "bbc5587a59fd5d5d9683ac5d957649ce",
+    "sha256hash": "8577f45d89d376a9bb1351b89e53c8d4e26327e465cc70b3bf361c975db53515",
+    "sha1hash": "072e199a7b2697b97e79c1fa6aa1a186b5e03dc3"
+}
```
Kelvinrr, line 9: Do the tests depend on having real binary kernels? Could the test setup not simply create an empty kernel? Part of the application is comparing hashes, is this hash comparison critical to the apps tests?

isis/src/system/apps/isisdataeval/tools/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+isisdata_mockup.py --saveconfig --isisdata /opt/isis/data/voyager1  --outpath $PWD/mockisisdata/voyager1 --ghostdir '$ISISDATA/voyager1'
+```
+Processing times can be significant since this script is computing 3 different hash values per file.
+
+
+# <a name="#isisdatamockupinstall">ISISDATA Mockup Test Data Preparation</a>
+
+With this tool and the files listed in `isisdataeval_isisdata_mockup_files.lis`, the test for isisdataeval can be recreated from any ISISDATA installation. Note that from time to time, the files used in this ISISDATA test could change which would cause failures. Here are the commands to create the `isisdataeval` test ISISDATA mockup directory - from the Git install directory:
+
```
Kelvinrr, line 89: Appreciate the thorough instructions.  Is it possible to simple make this mocked ISISDATA area static and not rely on the real data for comparison? From the tests it seems it's comparing hashes, my first thought is that we don't need to compare hashes from the original file? probably something I am missing.

Kelvinrr (commented):
Just some questions inline.

isis/tests/data/isisdata/mockup/hayabusa/kernels/spk/hay_osbj_050911_051118_v1n.bsp
KrisBecker: > Do the tests depend on having real binary kernels?

Only to compute (meaningful) hash values, which is why there is a reference to the real kernel/file.

> Could the test setup not simply create an empty kernel?

It could. But that does not test the validation aspect of this application. You could (and highly recommended to) pick kernels which you know its file name, size and contents should not change or file not exist (all part of verification) after any download/install. Modify the test to only compute hashes for these kernels/files and it automatically validates your environment. I don't know how to make a better/useful test.

> Part of the application is comparing hashes, is this hash comparison critical to the apps tests?

Yes. This validates $ISISDATA. By definition, this framework should also be completely consistent with conditions in any valid (i.e., fully downloaded/populated) ISIS test environment since it requires $ISISDATA to run the tests.

There are two things I cannot/do not test: file times and symbolic links since support/results have proven to be inconsistent across OSes for these file/directory/device characteristics. However, these data have proven to be useful on local systems, particularly when creating/configuring new mission kernel support systems in ISIS.

isis/src/system/apps/isisdataeval/tools/README.md
KrisBecker: Sure its possible to make data up, but as I mentioned above, this does not validate $ISISDATA. That would tell us nothing about ISISDATA or `isisdataeval` unless some additional effort or procedures are pursued on real data. These tests use selected files from four very stable $ISISDATA directories that should not change. This approach integrates verification and validation of required data for proper and more thorough testing of the ISIS system. 

Secondarily, but just as important, this provides tools to that can help address ongoing problems with ISISDATA encountered by ISIS users by first ensuring the problem is not related to their local ISISDATA installation.

Kelvinrr: I agree on the second part, after this is merged I plan to add it to the nightly ISISDATA script and find a good way to report results in AWS such that we can get a better idea of the changes other than did download/did not download. 

I think there might be a way to test the hash comparisons without real data, as the goal here is to test the app not the data itself, the data itself would be tested elsewhere. But there is already a ton of good work here and I'd rather get it in than split hairs on testing.

Kelvinrr (approved):
tests look good on Jenkins.

isis/src/system/apps/isisdataeval/tools/README.md
KrisBecker: Great. Can you hold off for a bit on the merge.I have an update coming.

It might address some of the things you mentioned.

PR #123, opened 2018-04-02
ORex SamCam and MapCam won't use PolyCam focus ID. Fixes #5213

jessemapel:
The value is still translated from the label, but will be unused.

PR #5149, opened 2023-03-10
Autolabeler for PRs

chkim-usgs:
## Description

Currently, the Github action `3pointer/check-pr-labels-on-push-action@v1.0.3` in the `autopr_bugfix.yml` workflow checks if the commit in the push to `dev` includes a PR and if that PR has labels.

The `pr_label_checker.py` script does the following steps:
1. Checks if the commit in the push to `dev` includes a PR
2. In the body of the PR, if there are any linked issues (matched via regex pattern, e.g., `#123`)
3. Get the labels from those linked issues and add to the PR
4. Outputs if the PR has a `bug` label (`boolean` output)

In the following PRs, you can see that the labels were added by the `github-actions` bot.
Also note below that the I modified the `release_pull_request` job for testing to `echo` something if the python script outputted `true` that the PR is a bugfix.

 Here are some scenarios I've been able to test on my fork:
- Merging PR with a linked issue that has `bug` and `enhancement` labels
  - PR: https://github.com/chkim-usgs/ISIS3/pull/12
  - Workflow: https://github.com/chkim-usgs/ISIS3/actions/runs/4386720455
  - Result: Two labels were applied to the PR and the `release_pull_request` job was run.
- Merging PR with a linked issue that has two non-bug labels:
  - PR: https://github.com/chkim-usgs/ISIS3/pull/11
  - Workflow: https://github.com/chkim-usgs/ISIS3/actions/runs/4386683098
  - Result: Two labels were added to the PR and the `release_pull_request` job was _not_ run.
- Merging PR with a linked issue that has no label:
  - PR: https://github.com/chkim-usgs/ISIS3/pull/10
  - Workflow: https://github.com/chkim-usgs/ISIS3/actions/runs/4386659535
  - Result: No label was added to the PR and the `release_pull_request` job was _not_ run.
- Merging PR with two linked issues that has two labels each (one `bug` label) :
  - PR: https://github.com/chkim-usgs/ISIS3/pull/13
  - Workflow:  https://github.com/chkim-usgs/ISIS3/actions/runs/4387156985
  - Result: Four labels were added to the PR and the `release_pull_request` job was run.

(Definitely need to squash commits when merging)

## Types of changes

- [x] Infrastructure change (changes to things like CI or the build system that do not impact users)

## Checklist:

## Licensing


.github/pr_label_checker.py
```diff
@@ -0,0 +1,168 @@
...
+from requests import Response
+from requests.exceptions import HTTPError, RequestException
+import sys
+
+# Constants
+GITHUB_TOKEN=os.environ["GITHUB_TOKEN"]
+GITHUB_API_URL=os.environ["GITHUB_API_URL"]
+GITHUB_SHA=os.environ["GITHUB_SHA"]
+
+BASE_URL=f'{GITHUB_API_URL}/repos/chkim-usgs/ISIS3'
```
AustinSanders: BASE_URL should be updated before merge

.github/pr_label_checker.py
```diff
@@ -0,0 +1,168 @@
...
+    except RequestException as re:
+        raise RequestException("Unable to retrieve list of PRs associated with commit.", re)
+
+def get_pr_attributes(response: Response) -> tuple:
+    """
+    Get necessary PR attributes.
+    """
+    pull_response_json = response.json()
+    if len(pull_response_json) == 0:
+        # No PRs in commit
```
AustinSanders: I'm a little confused by what this means.  Generally I view PRs as a list of commits, not the other way around.  Does this mean that the PR has no commit information?

.github/pr_label_checker.py
```diff
@@ -0,0 +1,168 @@
...
+        issue_labels = issue_response_json.get("labels")
+        for issue_label in issue_labels:
+            # Get name of each label object
+            label_name = issue_label.get("name")
+            if label_name not in combined_issue_labels:
+                # Add label if it does not exist
+                combined_issue_labels.append(label_name)
+    if not combined_issue_labels:
+        # No labels to return
+        print(False)
```
AustinSanders, line 100: It may be useful to print a more verbose message here or just remove the print?  I'm not really sure.

As an outside user I wouldn't really know what "False" meant in the logs without digging into the code.

.github/pr_label_checker.py
```diff
@@ -0,0 +1,168 @@
...
+        sys.exit(0)
+    return combined_issue_labels
+
+def convert_issue_list_to_dict(combined_issue_labels: list) -> dict:
```
AustinSanders: If this is only used in one area we might consider removing this function and just inlining the logic.  I think this is essentially just ```labels_data = {"labels": combined_issue_labels}```.  Up to you -- this might just be a personal preference.

.github/workflows/autopr_bugfix.yml
AustinSanders, line 1: Should this action be performed as part of the automatic PR process, or should it be performed independently? I'm not really sure.

I'd advocate for making the labeling functionality a separate action, because the automatic PR functionality will probably be moved to Jenkins.

AustinSanders (changes requested):

.github/pr_label_checker.py
AustinSanders: Or is this how the process communicates information...?

.github/pr_label_checker.py
chkim-usgs: Not too sure about the specifics of the commits in the PR, maybe if they're being squashed we wouldn't see the commits? But here's the [official GitHub doc](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-pull-requests-associated-with-a-commit) on that endpoint, which is what the `check-pr-labels-on-push-action` action used ([here](https://github.com/shioyang/check-pr-labels-on-push-action/blob/master/src/index.ts#L36)) before it was replaced with the python script.

.github/pr_label_checker.py
chkim-usgs: Yes, the following job in the Github workflow needs the output of this job, so this job echos the output into a variable that can be used within the workflow. The boolean has to be printed to be fed into the `echo` call but this approach does limit ways to print helpful messages. Open to different methods to output so we can have better logging.

.github/pr_label_checker.py
chkim-usgs: Good catch

.github/pr_label_checker.py
chkim-usgs: I'll update it once all the conversations are resolved just in case I need to test again on my fork.

.github/pr_label_checker.py
AustinSanders: Ahh -- no that makes sense.  It occurred to me after I left the comment that that might be how it worked.  If that's the standard way of facilitating IPC for GH actions then no reason to change it.

AustinSanders (approved):

PR #5255, opened 2023-08-08
fixes rclone recopying existing kernels

### LINKED ISSUE #5245: downloadIsisData and rclone Problems

amystamile-usgs:
## Related Issue

fixes #5245 

Ran script twice and confirmed there was no longer recopying of files.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/config/rclone.conf
+++ b/isis/config/rclone.conf
@@ -50,7 +50,7 @@ remote = asc_s3:asc-isisdata/usgs_data/lro/
 
 [tgo_naifKernels]
 type = combine
-upstreams = kernels=esa:/pub/naif/EXOMARS2016/kernels/
+upstreams = kernels=esa:/data/SPICE/ExoMars2016/kernels/
 
 [tgo_usgs]
 type = alias
@@ -68,9 +68,17 @@ upstreams = kernels=naif:/pub/naif/DAWN/kernels/
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/dawn/
 
-[dawn] 
-type = union 
-upstreams = dawn_naifKernels: dawn_usgs:
+[dawn_pck]
+type = combine
+upstreams = pck=sbn:dawn/fc/DWNCHSPG_2/GEOMETRY/
+
+[dawn_kernel_combine]
+type = combine
+upstreams = kernels=dawn_pck:
+
+[dawn]
+type = union
+upstreams = dawn_kernel_combine: dawn_naifKernels: dawn_usgs:
 
 [cassini_naifKernels]
 type = combine
@@ -86,7 +94,7 @@ upstreams = cassini_usgs: cassini_naifKernels:
 
 [hayabusa2_naifKernels]
 type = combine
-remote = kernels=jaxa:/pub/hayabusa2/spice_bundle/kernels/
+upstreams = kernels=jaxa:/pub/hayabusa2/spice_bundle/spice_kernels/
 
 [hayabusa2_usgs]
 type = alias
@@ -314,7 +322,7 @@ upstreams = smart1_usgs: smart1_naifKernels:
 
 [viking1_naifKernels]
 type = combine
-remote = kernels=naif:/pub/naif/VIKING/kernels/
+upstreams = kernels=naif:/pub/naif/VIKING/kernels/
 
 [viking1_usgs]
 type = alias
@@ -326,7 +334,7 @@ upstreams = viking1_usgs: viking1_naifKernels:
 
 [viking2_naifKernels]
 type = combine
-remote = kernels=naif:/pub/naif/VIKING/kernels/
+upstreams = kernels=naif:/pub/naif/VIKING/kernels/
 
 [viking2_usgs]
 type = alias
@@ -338,7 +346,7 @@ upstreams = viking2_usgs: viking2_naifKernels:
 
 [voyager1_naifKernels]
 type = combine
-remote = kernels=naif:/pub/naif/VOYAGER/kernels/
+upstreams = kernels=naif:/pub/naif/VOYAGER/kernels/
 
 [voyager1_usgs]
 type = alias
@@ -350,7 +358,7 @@ upstreams = voyager1_usgs: voyager1_naifKernels:
 
 [voyager2_naifKernels]
 type = combine
-remote = kernels=naif:/pub/naif/VOYAGER/kernels/
+upstreams = kernels=naif:/pub/naif/VOYAGER/kernels/
 
 [voyager2_usgs]
 type = alias
--- a/isis/config/rclone.conf
+++ b/isis/config/rclone.conf
@@ -18,15 +18,15 @@ url = http://www.darts.isas.jaxa.jp/
 
 [sbn]
 type =http
-url =https://sbnarchive.psi.edu/pds3/
+url = https://sbnarchive.psi.edu/pds3/
 
 [base]
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/base/
 
 [legacybase]
-type = union
-upstreams = asc_s3:asc-isisdata/usgs_data/legacy_base/
+type = alias
+remote = asc_s3:asc-isisdata/usgs_data/legacy_base/
 
 [control]
 type = alias
@@ -74,7 +74,7 @@ upstreams = dawn_naifKernels: dawn_usgs:
 
 [cassini_naifKernels]
 type = combine
-upstreams = kernels=naif:/pub/naif/CASSINI/kernels
+upstreams = kernels=naif:/pub/naif/CASSINI/kernels/
 
 [cassini_usgs]
 type = alias
@@ -86,7 +86,7 @@ upstreams = cassini_usgs: cassini_naifKernels:
 
 [hayabusa2_naifKernels]
 type = combine
-remote = kernels=jaxa:/pub/hayabusa2/spice_bundle/kernels
+remote = kernels=jaxa:/pub/hayabusa2/spice_bundle/kernels/
 
 [hayabusa2_usgs]
 type = alias
@@ -157,7 +157,7 @@ type = alias
 remote = asc_s3:asc-isisdata/usgs_data/mex/
 
 [mex]
-type = alias 
+type = union 
 upstreams = mex_usgs: mex_naifKernels:
 
 [hayabusa_naifKernels]
@@ -174,7 +174,7 @@ upstreams = hayabusa_usgs: hayabusa_naifKernels:
 
 [chandrayaan1_naifKernels]
 type = combine
-upstreams = kernels=esa:/data/SPICE/CHANDRAYAAN-1/kernels
+upstreams = kernels=esa:/data/SPICE/CHANDRAYAAN-1/kernels/
 
 [chandrayaan1_usgs]
 type = alias
@@ -238,7 +238,7 @@ upstreams = messenger_usgs: messenger_naifKernels:
 
 [mgs_naifKernels]
 type = combine
-upstreams = kernels=naif:/pub/naif/pds/data/mgs-m-spice-6-v1.0/mgsp_1000/data/kernels
+upstreams = kernels=naif:/pub/naif/pds/data/mgs-m-spice-6-v1.0/mgsp_1000/data/kernels/
 
 [mgs_usgs]
 type = alias
@@ -284,7 +284,7 @@ remote = asc_s3:asc-isisdata/usgs_data/osirisrex/
 type = union
 upstreams = osirisrex_usgs: osirisrex_naifKernels:
 
-[rolo_usgs]
+[rolo]
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/rolo/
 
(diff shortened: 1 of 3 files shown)


isis/config/rclone.conf
```diff
@@ -20,256 +20,354 @@ url = http://www.darts.isas.jaxa.jp/
...
+remote = kernels=jaxa:/pub/hayabusa2/spice_bundle/kernels
 
 [hayabusa2_usgs]
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/hayabusa2/
 
+[hayabusa2]
+type = union 
+upstreams = hayabusa2_usgs: hayabusa2_naifKernels:
+
```
Kelvinrr, line 106: with these new unions, I think the script needs to be updated to use them over the existing keywords of with the `_naif` `_usgs` suffixes.

acpaquette:
@amystamile-usgs We may want to drop the last commit here if the other PR gets merged with the same fix

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
 
+[cassini]
+type = union 
+upstreams = cassini_usgs: cassini_naifKernels:
+
 [hayabusa2_naifKernels]
-type = alias
-remote = jaxa:/pub/hayabusa2/spice_bundle/spice_kernels/
+type = combine
+remote = kernels=jaxa:/pub/hayabusa2/spice_bundle/kernels/
```
Kelvinrr: This needs to be: 

```
upstreams = kernels=jaxa:/pub/hayabusa2/spice_bundle/spice_kernels/
```

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
 
+[smart1]
+type = union 
+upstreams = smart1_usgs: smart1_naifKernels:
+
 [viking1_naifKernels]
-type = alias
-remote = naif:/pub/naif/VIKING/kernels/
+type = combine
+remote = kernels=naif:/pub/naif/VIKING/kernels/
```
Kelvinrr: remote needs to be changed to upstreams

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
 
+[viking1]
+type = union 
+upstreams = viking1_usgs: viking1_naifKernels: 
+
 [viking2_naifKernels]
-type = alias
-remote = naif:/pub/naif/VIKING/kernels/
+type = combine
+remote = kernels=naif:/pub/naif/VIKING/kernels/
```
Kelvinrr: another remote -> upstreams

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
 
+[viking2]
+type = union
+upstreams = viking2_usgs: viking2_naifKernels:
+
 [voyager1_naifKernels]
-type = alias
-remote = naif:/pub/naif/VOYAGER/kernels/
+type = combine
+remote = kernels=naif:/pub/naif/VOYAGER/kernels/
```
Kelvinrr: remote -> upstreams

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
 
+[voyager1]
+type = union 
+upstreams = voyager1_usgs: voyager1_naifKernels:
+
 [voyager2_naifKernels]
-type = alias
-remote = naif:/pub/naif/VOYAGER/kernels/
+type = combine
+remote = kernels=naif:/pub/naif/VOYAGER/kernels/
```
Kelvinrr: remote -> upstreams

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
-[lro_usgs]
+[lro]
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/lro/
 
 [tgo_naifKernels]
-type = alias
-remote = esa:/data/SPICE/ExoMars2016/kernels/
+type = combine
+upstreams = kernels=esa:/pub/naif/EXOMARS2016/kernels/
```
Kelvinrr: This should be reverted to `esa:/data/SPICE/ExoMars2016/kernels/`

isis/config/rclone.conf
```diff
@@ -18,258 +18,356 @@ url = http://www.darts.isas.jaxa.jp/
...
+type = combine
+upstreams = kernels=naif:/pub/naif/DAWN/kernels/
 
 [dawn_usgs]
 type = alias
 remote = asc_s3:asc-isisdata/usgs_data/dawn/
 
-[dawn_pck]
-type =alias
-remote = sbn:dawn/fc/DWNCHSPG_2/GEOMETRY/
```
Kelvinrr: I think we want to keep dawn_pck, then include it in the union. 

problem is, the only way I know how to do this within the confines of rclone is something like: 

```
[dawn_naifKernels]
type = combine
upstreams = kernels=naif:/pub/naif/DAWN/kernels/

[dawn_usgs]
type = alias
remote = asc_s3:asc-isisdata/usgs_data/dawn/

[dawn_pck]
type = combine
upstreams = pck=sbn:dawn/fc/DWNCHSPG_2/GEOMETRY/

[dawn_kernel_combine]
type = combine
upstreams = kernels=dawn_pck:

[dawn]
type = union
upstreams = dawn_kernel_combine: dawn_naifKernels: dawn_usgs:
```

isis/scripts/downloadIsisData
```diff
@@ -189,33 +184,24 @@ def main(mission, dest, cfg_path, parsedArgs, kwargs):
...
-    supported_missions = OrderedDict({})
-    for source in sorted(config_sources, key=lambda x: x.split("_")[-1]):
-        parsed_name = source.split("_")
-        # If it is a mission, it should be in the format <mission_nam>_<source_type>
-        if len(parsed_name) == 2 and parsed_name[1] in ["usgs:", "naifKernels:"]:
-            remotes_mission_name = parsed_name[0]
-            supported_missions[remotes_mission_name] = supported_missions.get(remotes_mission_name, []) + [source]
+    supported_missions = []
+    for source in config_sources:
+        if not any(x in source for x in ["asc_s3:", "esa:", "naif:", "jaxa:", "sbn:", "usgs:", "naifKernels:", "legacybase"]):
```
Kelvinrr: `["usgs", "naifKernels"]` can probably be removed from the list. 

I think including legacy base in this is also causing it to not be downloaded. Seems to be downloading base directly rather than downloading the legacybase target to the base folder.

isis/scripts/downloadIsisData
```diff
@@ -123,17 +123,12 @@ def create_rclone_arguments(destination, mission_name, parsedArgs, rclone_kwargs
...
     log.debug(f"Creating RClone command for {mission_name}")
-    mission_dir_name, source_type = mission_name.replace(":", "").split("_")
     
-    if (mission_dir_name == "legacybase"):
+    if (mission_name == "legacybase"):
         # We still want things to go into base
-        mission_dir_name = "base"
-        
-    log.debug(f"Mission_dir_name: {mission_dir_name}, source_type: {source_type}")
+        mission_name = "base"
```
Kelvinrr: aforementioned legacybase download issue might actually be coming from here instead?

isis/scripts/downloadIsisData
amystamile-usgs: I wanted to keep the supported_missions list to be only the names of the mission and not include `<mission>_usgs` or `<mission>_naifKernels` purely for user entry. I believe there is an error later down that displays this list to the user and I don't want any confusion. I will look into the legacybase issue.

amystamile-usgs: Just had it look for `_` instead since there are multiple with underscores that we don't want to add to the list. Removed `legacybase` since we want users to know that is an option. Removed it from list when user chooses `all`.

Kelvinrr (approved):

PR #5345, opened 2023-11-17
ShapeModel Normal State Fix

### LINKED ISSUE #5197: Camstats output doesn't make sense
**ISIS version(s) affected**: 7.1.0, 7.2.0
I get expected/sensible output under 5.0.2, 6.0.0, 7.0.0

**Description**  
Before recovering lost Kaguya TC data under 7.2.0, I compared the caminfo output of one image to that created under isis5.0.2 and found the camstats information to be quite different.  cubediff run on the old image and this new image shows them to be identical. However, the caminfo/camstats reporting of emission and incidence angle information is quite different.

under 5.0.2 and 6.00
    MinimumEmission          = 16.423659427042
    MaximumEmission          = 19.293244403461
    MinimumIncidence         = 59.443819396967
    MaximumIncidence         = 61.109253930639

no caminfo output under 7.0.0 because of an introduced bug in caminfo

7.1.0 and 7.2.0 report
    MinimumEmission                          = 1.0513035024623
    MaximumEmission                          = 41.396795355947
    MinimumIncidence                         = 35.210026880994
    MaximumIncidence                         = 85.765737110116

These values make no sense for an equatorial image from a mapping mission.  Additionally, basic geometry information in the image .lbl file includes 
INCIDENCE_ANGLE                      =  60.288 <deg>
EMISSION_ANGLE                       =  16.580 <deg>

which is similar 5.0.2 and 6.0.0 output.
Note that the values for oblique resolution are also different between the versions and everything else is the same.

**How to reproduce**  
See data under my work area Isis3Tests/Camstats/.  
  top level directory has output for isis7.2.0
  Isis5.0.2/ contains data for that version for comparison
  Isis7.0.0 contains data for that version for comparison

Here's the processing history of the test image (img and lbl in directory for use).  I removed path info for the shape model.
```
kaguyatc2isis from=TC1W2B0_01_00607S370E2634.lbl to=TC1W2B0_01_00607S370E2634.lev0.cub setnullrange=yes nullmin=-32768 nullmax=-0.00001 

spiceinit from=TC1W2B0_01_00607S370E2634.lev0.cub spksmithed=true shape=user model=Lunar_LRO_LOLAKaguya_DEMmerge_Global_512ppd_radius_demprep.cub

camstats from=TC1W2B0_01_00607S370E2634.lev0.cub attach=true linc=100 sinc=100 

footprintinit from=TC1W2B0_01_00607S370E2634.lev0.cub increaseprecision=FALSE inctype=vertices numvertices=40

caminfo from=TC1W2B0_01_00607S370E2634.lev0.cub to=TC1W2B0_01_00607S370E2634.lev0.pvl isislabel=true geometry=true originallabel=true statistics=true camstats=true linc=100 sinc=100 uselabel=true
```
I confirmed the differences exist at the camstats step when I output to a pvl file and make comparisons. Although caminfo/camstats was not generating output under 7.0.0, that version of camstats matches the output of isis5.0.2 which are the expected values.

I don't know if these differences are seen for other data sets.

**Additional context**

acpaquette:
## Description
The `m_normal` variable in ISIS was being used to store both the "local" normal (normal obtained from more detailed shape models) and the ellipsoid surface normal. While they could remain the same it is best that the local normal be stored in one variable (`m_localNormal`) and the ellipsoid normal retain the current behavior and store its value in `m_normal`. This ensures a cleaner state within the shapemodel hierarchy but requires resetting both normals when necessary.

## Related Issue
Fixes #5197 

Validated through existing tests

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/ShapeModel/ShapeModel.cpp
```diff
@@ -474,7 +504,7 @@ namespace Isis {
 
 
   /**
-   * Sets the normal for the currect intersection point.
+   * Sets the surface normal for the currect intersection point.
    * Note: This method will throw an error if this ShapeModel doesn't have
    * and intersection. Use the hasIntersection() method to verify before
```
chkim-usgs: Spelling typo: "and intersection" to "an intersection"
Applies to the `setLocalNormal()` description below too (and a couple more funcs below as well)

isis/src/base/objs/ShapeModel/ShapeModel.cpp
```diff
@@ -495,9 +525,31 @@ namespace Isis {
...
+   * @param normal Three dimensional local normal vector.
+   *
+   */
+  void ShapeModel::setLocalNormal(const std::vector<double> normal) {
+    if (m_hasIntersection) {
+      m_localNormal = normal;
+      m_hasLocalNormal = true;
+    }
+    else {
+      QString message = "No intersection point in known.  A local normal can not be set.";
```
chkim-usgs: Should the message say "No intersection point _is_ known"? If so, this should be addressed in the `setNormal()` func above too. Edit: This may need to be addressed in the following couple funcs below too.

acpaquette:
@KrisBecker Some input here may be useful, I know this is not how you fixed this issue but if you can think of any pitfalls of this approach that would be ideal

KrisBecker:
@acpaquette is it possible to get local incidence and emission angles from apps like campt and camstats? Is this the desired behavior?

Can you explain why the unit tests changed so much? Were they returning local angles before your modification?

acpaquette:
@KrisBecker Campt and camstats have never explicitly exposed local emission and incidence angles like phocube. They were unintentionally returning local angles after #4600 was merged into ISIS due to the shape models sharing one variable for the normal. This PR separates the normal and local normal into distinct variables that need to be requested by the program/programmer, thus the change to unittests to return the `localNormal` vs `Normal` when the "local" normal is expected.

If we want local emission and local incidence exposed in camstats and campt, that is something we could do, but is not currently available. Also in #4600 the "Functional" tests truth values were changed, stating that the local emission and incidence for camstats and campt was the "truth". This PR rectifies that and the new values are extremely close to the old values. Why they aren't exact I am not sure. 

Whether users want the "local" normal vs the ellipsoid normal is somewhat unclear. To some degree it's down to user expectation. If I ask for a normal with a dem or bullet shape model should I expect the "local" normal or the ellipsoid normal? Should users be able to more directly change the shapemodel of a cube without having to run spiceinit again? Some of these question may trickle into PSRMTS but may ultimately remain an issue for ISIS to tackle. Another possible avenue is that if we want all shapemodels to have access to the ellipsoid normal, it may make more sense for users to have to ask for that rather than the "local" normal.

KrisBecker:
Unfortunately, none of the current ISIS tests for this functionality include any small, irregular bodies. In fact, there are no ISIS tests for NAIF DSK, Bullet or Embree shape models that would evaluate the impact of these changes regarding tessellated shape models. The tests used for these cases are Mars global DEM 2.5D shape models in ISIS cube projection format. The impact of modifications to ISIS shape model code for small, irregular bodies (i.e., tessellated plate shape models) lacks representation.

Triaxial ellipsoid representations generally do not adequately model small, irregular bodies. Tessellated plate models are required for more accurate cartography functionality. The implementations for ray tracing shape models we have are decidedly different from what is currently in ISIS. The emphasis is on accuracy and flexibility.

I encourage you to run more comprehensive tests that at least compare the current ISIS implementation to your changes. These tests could include campt, camstats, caminfo, footprintinit, phocube, orthorectified projections (cam2map, cam2cam), mappt, etc..., using small bodies and tessellated shape models. I would also compare results from each ray tracing system - NAIF DSK, Bullet, Embree and ellipsoids as well. Some instruments/datasets candidates would be NEAR/MSI and Eros, Hayabusa1/AMICA and Itokawa. Bennu, although small, is not that irregular. But it does have a larger variety of shape models.

I do hope we can have discussions about small, irregularly shaped body support in ISIS and the ShapeModel design specifically.

acpaquette:
@KrisBecker I agree that those would be useful tests, I don't think this PR is a good place to add them. I will make a new issue and we can address it in future support sprints

chkim-usgs (dismissed):
Should we add a changelog entry signifying the updated ShapeModel differentiating a local normal and an ellipsoid surface normal? Looks good either way.

acpaquette:
@chkim-usgs Yes, this should have a changelog entry. I will add that

AustinSanders (dismissed):

amystamile-usgs:
@acpaquette conflicting changelog

Kelvinrr (approved):
fixed the merge conflict

PR #5398, opened 2024-01-08
8.1.0_RC2 Release Prep

chkim-usgs:
Cherry-pick (commit hash c1cc08a) necessary changes from `dev` into `8.1`:

* Update csm plugins path in IsisPreferences

* Update changelog

* Add trailing slash

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,7 @@ release.
 
 ## [Unreleased]
 
-### Added
-- Added new csm plugins path to IsisPreferences [#5397](https://github.com/DOI-USGS/ISIS3/pull/5397)
-
-## [8.1.0] - 2023-12-08
+## [8.1.0] - 2024-01-08
 
 ### Changed
 - Cnetthinner has been refactored to be callable; old Makefile tests have been removed and replaced by gtests. Issue: [#5354](https://github.com/USGS-Astrogeology/ISIS3/issues/5354), 
@@ -67,6 +64,7 @@ release.
 - Added ability to delegate calculation of nadir pointing to ALE [#5117](https://github.com/USGS-Astrogeology/ISIS3/issues/5117)
 - Added --no-kernels flag to downloadIsisData [#5264](https://github.com/DOI-USGS/ISIS3/issues/5264)
 - Added notice for depreciation to PvlObject::addLogGroup function which will be depreciated in favor of Application::appendAndLog(https://github.com/DOI-USGS/ISIS3/issues/5310)
+- Added new csm plugins path to IsisPreferences [#5397](https://github.com/DOI-USGS/ISIS3/pull/5397)
 
 ### Deprecated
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ release.
 
 ## [Unreleased]
 
+### Added
+- Added new csm plugins path to IsisPreferences [#5397](https://github.com/DOI-USGS/ISIS3/pull/5397)
+
```
acpaquette: Can this go under the Added section of 8.1.0 instead of unreleased?

code.json
```diff
@@ -91,22 +91,22 @@
     "name": "ISIS3",
     "organization": "U.S. Geological Survey",
     "description": "GitHub code repository for the Integrated Software for Imagers and Spectrometers (ISIS)",
-    "version": "8.1.0_RC1",
+    "version": "8.1.0_RC2",
```
acpaquette, line 94: Can this be a separate code.json entry?

recipe/meta.yaml
```diff
@@ -18,12 +18,12 @@
...
 #       A Release Candidate for ISIS3.6.1:                     {% set version = "3.6.1_RC" %}
 #       A custom build of ISIS3.6.1 for the CaSSIS mission:    {% set version = "3.6.1_cassis" %}
-{% set version = "8.1.0_RC1" %}
+{% set version = "8.1.0_RC2" %}
 
 # This is the build number for the current version you are building. If this is the first build of
 # this version, the build number will be 0. It is incremented by 1 with every consecutive build of
 # the same version.
-{% set build_number = "0" %}
+{% set build_number = "1" %}
```
acpaquette: I think this should be set to 0?

acpaquette (commented):
A couple small changes to address

recipe/meta.yaml
chkim-usgs: Sounds good, does a consecutive build mean something like if I need to rebuild `8.1.0_RC1` and overwrite the existing RC1 conda package in Anaconda?

code.json
chkim-usgs: Should we be appending for the code.json for every new release?

acpaquette: Yes, it gets a little annoying but every release needs a code.json entry

recipe/meta.yaml
acpaquette: Correct, in this case we are releasing a different version so we reset the build number to 0. If we went back to release RC1 for some reason we would increment the build number instead

code.json
chkim-usgs: Np, I think I was just confused since when I've looked at previous release PRs, they would overwrite the existing code.json entry and also the current code.json entry only shows `8.0.0` entry but nothing prior.

acpaquette (approved):

Issue #5612, opened 2024-09-12
Isisminer does not properly handle the presence of bad (e.g. self-intersecting) polygons.


kledmundson:
**ISIS version(s) affected**: x.y.z
dev

**Description**  

The _footprintinit_ application will sometimes create invalid, self-intersecting polygons. The _isisminer_ application detects bad polygons, reports them, and terminates instead of attempting to repair them. 

**Solution**  

In many cases, self-intersecting polygons can be repaired by creating a buffer of size 0 around the polygon. This is analogous to stretching a rubber band around all points in the polygon (see slide below). The buffer(0) approach will likely fix many (though possibly not all) self-intersecting polygons. While the fix should be incorporated into _footprintinit_, it should also be implemented in _isisminer_. 

**Additional context**  

The slide below shows a polygon with a highlighted self-intersection in blue. The self-intersection was repaired (shown in orange) by applying a buffer of 0.1 around all polygon vertices. Note that typically a buffer of size 0 is applied. Here, a buffer of 0.1 was used for illustration purposes only. 

<img width="1331" alt="selfIntersectingPolygons" src="https://github.com/user-attachments/assets/269712bb-1df7-4bf3-81f3-750089de63cb">

acpaquette:
@kledmundson Couple insights/questions. Is there a planned fix from the yourself or Kris coming in for this? Second, is seems related to #5553

kledmundson:
Yes, fix is coming.

Relation to #5553 I'm not sure about, will have to look.

KrisBecker:
This fix is not directly related to #5553 because [GisGeometry](https://github.com/DOI-USGS/ISIS3/blob/dev/isis/src/base/objs/GisGeometry/GisGeometry.h) uses the GEOS C-API. Code in [PolygonTools](https://github.com/DOI-USGS/ISIS3/pull/5553/files) uses the C++ API so they are separate. We added a `buffer()` method to GisGeometry.

They are likely related as some GIsFootprints created by ISIS have self-intersections. These *leak* through into isisminer processing and aborts the run. The fix that @kledmundson is providing runs a `buffer( 0 )` on the whole polygon in an attempt to fix them. Most all occurrences are fixed using this technique.

kledmundson:
Closed via https://github.com/DOI-USGS/ISIS3/pull/5620.

PR #5620, opened 2024-09-23
Bug fixed in isisminer to properly handle self-intersecting polygon geometries.

kledmundson:
## Description

The isisminer application detects bad polygons, reports them, and terminates instead of attempting to repair them. Often (though not always), self-intersecting polygons can be repaired by creating a buffer of size 0 around the polygon.

Here, the Strategy and GisGeometry classes have been updated to better manage geometries. The buffer(0) method was added to GisGeometry to repair self-intersecting polygons.  The _RepairInvalidGeometry_ and _InvalidGeometryAction_ keywords have been added to the Strategy class providing improved user control over invalid geometries.

The _RepairInvalidGeometry_ keyword defaults to true so that invalid geometries will be repaired. Options for the _InvalidGeometryAction_
keyword are _continue_, _disable_ or _error_. For InvalidGeometryAction = _continue_, the state of the invalid geometry is retained in the Resource (some GIS operations still function) and it is not disabled and no error occurs - the issue is ignored. If InvalidGeometryAction = _disable_, then the geometry is retained in the Resource but the status is set to "discard". It can be re-enabled by using the ResourceManager strategy. If InvalidGeometryAction = _error_, then an error is thrown and isisminer terminates. The default value is _disable_.

## Related Issue

https://github.com/DOI-USGS/ISIS3/issues/5612

The GisGeometry unit test has been modified to test the validity and repair of a self-intersecting polygon. The Strategy unit test has been updated to test the behavior of the two new keywords _RepairInvalidGeometry_ and _InvalidGeometryAction_.

**GisGeometry unit test result**

ctest -R GisGeometry --output-on-failure
Start 108: isis_unit_test_GisGeometry
1/1 Test #108: isis_unit_test_GisGeometry .......   Passed    2.01 sec

**Strategy unit test result**

ctest -R isis_unit_test_Strategy --output-on-failure
Start 299: isis_unit_test_Strategy
1/1 Test #299: isis_unit_test_Strategy ..........   Passed    0.56 sec

Additionally, all _isisminer_ app tests continue to pass.

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation and I have updated the documentation accordingly.
- [X] I have added my user impacting change to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document. 

## Licensing


isis/src/base/objs/GisGeometry/GisGeometry.cpp
```diff
@@ -207,15 +207,25 @@ namespace Isis {
...
     if (!isDefined()) {
       return (false);
     }
-
-    return (1 == GEOSisValid(this->m_geom));
+    
+    int valid(0);
+    try {
+        valid = GEOSisValid(this->m_geom);
+    } catch (...) {
```
AustinSanders, line 225: If I'm reading the [docs](https://libgeos.org/doxygen/geos__c_8h.html#ae65e55acad28e2b55d793730dd1d4e06) correctly, the catch(...) is not necessary / reachable.

It looks like exceptions are handled within GEOSisValid, and the function returns 2 on exception.

isis/src/base/objs/GisGeometry/GisGeometry.cpp
```diff
@@ -601,9 +611,35 @@ namespace Isis {
...
+ *  
+ * Add a buffer around a geometry with the defined enlargement factor. It is 
+ * common to use this as buffer(0) to fix geometries with self-intersections. 
+ * These cases are often identified by calling isValid() and getting a false 
+ * value returned. 
+ * 
+ * @author 2018-07-29 Kris Becker
+ * 
+ * @param width    With to enlarge or shrink polygon 
+ * @param quadsegs Numnber of segments to define a circle on corners
```
AustinSanders: Numnber -> Number

AustinSanders (changes requested):

kledmundson:
Thanks for reviewing @AustinSanders. I fixed the typo you caught (plus a couple more). Regarding the method _GisGeometry::isValid()_, we have confirmed that it is correct as written. Verified that in the presence of self-intersecting polygon geometry the catch(...) statement is executed properly.

AustinSanders (approved):
I'm not understanding what types of exception can be thrown by GeosIsValid(), which is leading to my confusion about the catch(...).  The documentation states that the function returns "1 on true, 0 on false, 2 on exception."

Ultimately, I trust that the code is functioning properly, I just don't understand the circumstances under which the catch(...) is necessary.  I'll approve for merge, but I'd also like to understand the code better if you don't mind discussing.

kledmundson:
Ok, thanks.

What's going on with this does appear to be a bit cryptic to me as well,
considering the GEOS documentation.

I will rerun again with & without the try/catch in the presence of
self-intersecting polygons and provide more details.

kledmundson:
Hi @AustinSanders, Kris and I went through this yesterday, here are the details (let us know what you think) ... 

1) The GEOS C-API [requires]( https://libgeos.org/usage/c_api/#building-a-program) that users of the library provide handlers for both GEOS notices and errors. Static methods implemented in the GisTopology class receive and process status messages in text. [GisTopology::notice()](https://github.com/DOI-USGS/ISIS3/blob/275ae63f6f3aba93b674816cde8815eb41581200/isis/src/base/objs/GisTopology/GisTopology.cpp#L322-L336) and [GisTopology:error()](https://github.com/DOI-USGS/ISIS3/blob/275ae63f6f3aba93b674816cde8815eb41581200/isis/src/base/objs/GisTopology/GisTopology.cpp#L339-L353) provide GOES/ISIS notice/error handling.

2) The try/catch structure in GisGeometry:: isValid() is required as GEOS notices and errors are relayed back to GisTopology::notice() and GisTopology:error() respectively. They both throw ISIS exceptions when they are called from within the GEOS system when it generates an internal notice or error. See also [GisTopology::geosInit()](https://github.com/DOI-USGS/ISIS3/blob/275ae63f6f3aba93b674816cde8815eb41581200/isis/src/base/objs/GisTopology/GisTopology.cpp#L305-L310) which initializes this condition upon factory creation.

3) A GEOS notice is generated when evaluation of any GEOSGeometry contains a self-intersection. This, in turn, results in an ISIS exception, which is thrown in GisTopology::notice(), even when GEOSisValid() is called. Below is the output generated from _isis_unit_test_GisGeometry_ when we replaced the throw with a print stmt. Note that GisGeometry::isValidReason() can be called after a false return from GisGeometry::isValid() to retrieve that same message.

< GEOS Notice: Self-intersection at or near point 286.74713584288054 51.271685761047472 91d89
< GEOS Notice: Self-intersection at or near point 286.74713584288054 51.271685761047472 94d91
< GEOS Notice: Self-intersection at or near point 286.74713584288054 51.271685761047472 96d92
< GEOS Notice: Self-intersection at or near point 286.74713584288054 51.271685761047472 98d93
< GEOS Notice: Self-intersection at or near point 286.74713584288054 51.271685761047472

4) We think it _might_ be ok to remove the throw in GisTopology::notice() because the desired behavior is retained without the try/catch. However, we would be reluctant to remove the throw in GisTopology:error() as GEOS error behavior is uncertain (e.g., GEOSisValid() may be the only opportunity to handle real GEOS errors).

5) While it's true that the GEOS documentation indicates that GeosIsValid() will not produce an error, the above design/implementation decisions made the catch(...) necessary.

AustinSanders:
That all makes sense!  Thanks for writing that up

Issue #1355, opened 2019-02-12
qnet - can't change reference / no warning when trying to do so


ascbot:
---

Author Name: **Lynn Weller** (Lynn Weller)

Original Assignee: Tracie Sucharski

---

I have a control net where all points have references and I have noticed that while in the point editor if I put a new image in the left window then try to save measure or save point, I am no longer prompted with a warning about changing the reference ("The point already contains a reference measure, would you like to replace it with the measure on the left?").  More concerning is I can't change the reference at all in qnet!

I have to go back to isis3.3.0 to get the warning and be able to change the reference manually.  The latest version of production available is 05-16-2012 which doesn't have the warning and doesn't let the user change the reference (there are no errors with this, the reference simply won't change).  The last time I was in qnet frequently was about 3 weeks ago and I can't say if this problem existed then or not.  I feel something possibly changed during May, but I can't say with certainty, but I know it wasn't that long ago that this capability to change reference existed.

*Steps to reproduce:*

- Open a list and network in qnet (the one supplied contains mostly constrained points with a couple of free points....doesn't make a difference what the point type is though)

- load a point in point editor

- change the image in the left window to something other than reference=true

- modify measure in right window

- click save measure

- note that there is no warning (will get one in isis3.3.0) and
  the reference does not change (meaning, the image in the left does not change 
  to reference=true)

As far as I can tell, there is no way to manually change a points reference.

*Additional information:*

Set triage pretty high; user is already having to revisit networks to double check status of points/measures due to this issue.

Looks like a lost check based on reference to isis3.3.0?

ascbot:
---

Original Redmine Comment
Author Name: **Lynn Weller** (Lynn Weller)
Original Date: 2012-06-06T19:57:15Z

---

There is a small network available under 
/work/users/lweller/Isis3Tests/Jigsaw/Themis_Fit/

where all of the points are free and no weights are set if desired.  It really has no affect on how qnet is operating, but in case you wanted something else work with...

ascbot:
---

Original Redmine Comment
Author Name: **Tracie Sucharski** (Tracie Sucharski)
Original Date: 2012-06-13T16:10:58Z

---

This is fixed and available for testing in:

/work/projects/isis/latest/m00908/isis/src/qisis/apps/qnet

ascbot:
---

Original Redmine Comment
Author Name: **Lynn Weller** (Lynn Weller)
Original Date: 2012-06-13T17:20:15Z

---

[...]

Issue #1354, opened 2019-02-12
phocube and lat/lon


ascbot:
---

Author Name: **Tammy Becker** (Tammy Becker)

Original Assignee: Lynn Weller

---

External post:
https://isis.astrogeology.usgs.gov/IsisSupport/index.php/topic,3487.0.html

So, I would like to create output lat/lon bands to a global image I'm working with for ingestion by another person that uses a different program. I don't think 'Just use ISIS' would go over well. So, can you help me out. I tried 'phocube' but it keeps telling me I need an instrument section in my header. I'm not sure why since it has the corner lat/lons etc in the mapping section. Do you have any tricks so I can get these bands without going to great lengths? I would be helpful for phocube to be able to do this with data that does not have Instrument data. Is there another function I should be using?

Thanks,
Josh 

*Additional information:*

Hi Josh,

In order for 'phocube' to generate information it needs either 1) a camera model with spice ...or 2) a map projected image.  Notice the 'source' parameter in phocube, you have a choice between camera and projection.

If the images you have are 'level 1' camera-space data, Isis requires the camera model and all the ephemeris data (spice) that goes with it in order to know how to intersect with the surface of the target body, etc.  The lat/lon corner information is simply not enough.

If the images you have are 'level 2-map projected' or a digital map, then you have something to work with.  You would need to run 'maplab' with all the information about the map-projected data (through a 'maptemplate' file). This would include the projection, center lat, center lon of the projection, the pixel resolution and the latitude/longitude extents (corners).  'maplab' will then generate a 'Mapping Group' on your cube labels that can then give 'phocube' enough information to generate latitude/longitude output.  Please keep in mind that 'phocube' will not be able to generate photometric information (phase, emission, incidence angles, etc.)..spice is required for this in level1 and level2 images.

Good luck!
Tammy

ascbot:
---

Original Redmine Comment
Author Name: **Tammy Becker** (Tammy Becker)
Original Date: 2012-06-05T20:00:59Z

---

Josh's reply:
Hmmm...  it is map projected, I did make a map template, and I did use map lab.   I'll try it again. But, I've only ever had success with images that had spice data.

ascbot:
---

Original Redmine Comment
Author Name: **Tammy Becker** (Tammy Becker)
Original Date: 2012-06-05T20:01:42Z

---

Trent's contribution:

[...]

PR #5704, opened 2025-01-09
Add new parameters OFFBODY and OFFBODYTRIM to cam2cam

artmathgirl:
Add new parameters OFFBODY and OFFBODYTRIM to cam2cam. Updated functional tests. Updated documentation. 

## Description

Enhance cam2cam to project all pixels in the focal plane - on and/or off the target body - to the output image. This has been extremely useful in detecting and tracking ejected particles from Bennu's surface.

Two program parameters were added - OFFBODY and OFFBODYTRIM.

Enhancement to cam2cam is necessary to support upcoming addition of the same new parameters (OFFBODY, OFFBODYTRIM) to noproj. 

## Related Issue

https://github.com/DOI-USGS/ISIS3/issues/3602

Existing functional test passed.
New tests added to exercise the functionality of the new parameters, confirmed passing.

ctest -R Cam2Cam --output-on-failure
Test project /Users/ssutton/ISIS/GitCheckOuts/cam2camKeywords/ISIS3/build
    Start 1354: DefaultCube.FunctionalTestCam2CamNoChange
1/2 Test #1354: DefaultCube.FunctionalTestCam2CamNoChange ...   Passed    2.74 sec
    Start 1355: DefaultCube.FunctionalTestCam2CamOffbody
2/2 Test #1355: DefaultCube.FunctionalTestCam2CamOffbody ....   Passed    9.33 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =  12.26 sec

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation and I have updated the documentation accordingly.
- [X] I have added my user impacting change to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document. 

## Licensing


isis/src/base/apps/cam2cam/cam2cam.h
```diff
@@ -26,11 +35,14 @@ namespace Isis {
...
       int p_outputLines;
+      bool p_offbody;
+      bool p_trim;
 
     public:
       // constructor
       cam2camXform(const int inputSamples, const int inputLines, Isis::Camera *incam,
-                   const int outputSamples, const int outputLines, Isis::Camera *outcam);
+                   const int outputSamples, const int outputLines, Isis::Camera *outcam,
+                   const bool offbody = false, const bool trim = true);
```
acpaquette: I think both offbody and trim need to be set to true here to avoid api breakage

isis/src/base/apps/cam2cam/cam2cam.h
```diff
@@ -41,6 +53,132 @@ namespace Isis {
...
+   * 
+   * Note this should *not* be used as the processing algorithm directly in
+   * cam2cam because it does not use the RubberSheet mapping algorithm as the
+   * main app uses. This essentially uses the RubberSheet::SlowGeom() method
+   * (w/out interpolation!), which is the least efficicent algorithm.
+   * However it may be applicable in a threaded implementation.
+   * 
+   * @author 2024-10-27 Kris J. Becker
+   */
+  class Cam2CamMapper {
```
acpaquette: Since this class is only used in the tests, can it be moved into ```FunctionalTestsCam2cam.cpp```?

isis/src/base/apps/cam2cam/cam2cam.xml
```diff
@@ -154,6 +229,295 @@
           </option>
         </list>
       </parameter>
+      
+      <parameter name="OFFBODY">
+        <type>boolean</type>
+        <default><item>FALSE</item></default>
```
acpaquette, line 235: Similar to the API comment, I think this should be set to true to maintain default behavior.

acpaquette (changes requested):
Overall looks good! Couple small things to address

isis/src/base/apps/cam2cam/cam2cam.h
KrisBecker: Not sure what you mean. Current cam2cam default behavior is to map only valid lat/lon data. It has never projected RA/DEC data. Thus, offbody=false must remain in order to preserve existing application behavior. 

I don't think this configuration breaks API because these parameters do not exist in the current API and appropriate default value are provide for the new parameters to preserve existing behavior.

isis/src/base/apps/cam2cam/cam2cam.h
KrisBecker: I think this class has general utility other than strictly testing. All one has to do is add the include file in their own code.

It also keeps the class with the context in which it is designed to model.

isis/src/base/apps/cam2cam/cam2cam.h
acpaquette: My mistake, I thought the current behavior did map off-body

isis/src/base/apps/cam2cam/cam2cam.xml
acpaquette: See previous comment for closing

isis/src/base/apps/cam2cam/cam2cam.h
acpaquette: If cam2cam was updated to use it I would agree but it isn't currently being used in the codebase, only the test suite.

It also seems like a rewrite of the cam2camXform, could one not use the cam2camXform in the same way that one would use this new class?

KrisBecker: Yeah, you may be right.

It is odd that it appears the only way to use this externally would require a link to cam2cam.o.  It would be more better/useful if the implementation of cam2camXform is entirely in the header.

acpaquette: We could move the xform implementation into the .h as well

acpaquette:
@artmathgirl @KrisBecker Could the Cam2CamMapper be moved into `FunctionalTestsCam2cam.cpp` or was there a reason that remained in the `cam2cam.h`?

KrisBecker:
@acpaquette It could be moved, sure. I just think this has utility beyond cam2cam and can be easily used to correctly map coordinates between two images, particularly small bodies. Its presence in cam2cam.h makes it readily accessible to developers without them having to code their own version (or copy it).

acpaquette:
From our previous conversation it seemed like we agreed that the `cam2camXForm` could accomplish the same processing one would do with the `Cam2CamMapper`, in which case having both exposed is superfluous.

We should either update cam2cam to use the `Cam2CamMapper` and remove `cam2camXForm` or move the `Cam2CamMapper` into the test file, while exposing `cam2camXForm` in the .h file as Sarah has done. If someone has a use case for the `Cam2CamMapper` that can't be accomplished via `cam2camXForm` we can expose it again in a header file somewhere.

acpaquette (approved):
Sweet, looks good!

PR #5741, opened 2025-03-06
Initial ISIS install script

acpaquette:
## Description
Initial ISIS install script for users, based off of an initial version from @Kelvinrr

## Related Issue
Docs IAA

User testing

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/scripts/install_isis.sh
+++ b/isis/scripts/install_isis.sh
@@ -1,0 +1,480 @@
+#!/bin/bash
+
+# Define environment name and package to install
+DOWNLOAD_DATA="YES"
+CLIENT="mamba"
+
+# Report error and reserves error code from a failed command
+failed_command() {
+    local _status=$?
+    if [ "${_status}" -ne 0 ]; then 
+        echo "Error: $1 failed" 
+        exit ${_status}
+    fi
+}
+
+# Apply check for optional arguments that expect a value
+check_valid_arg() {
+    if [[ "$2" = "-"* ]]; then
+        echo "Invalid argument: $2"
+        echo "Found after $1"
+        exit 1
+    fi
+
+    if [[ -z $2 ]]; then
+        echo "No value for $1 provided"
+        exit 1
+    fi
+}
+
+print_help() {
+    printf "Usage: $0 [options]\n"
+    printf "Options:\n"
+    printf "\t-h, --help            Show this help message and exit\n"
+    printf "\t-l, --anaconda-label  Different ISIS labels as defined by the anaconda\n"
+    printf "\t                      labels at https://anaconda.org/usgs-astrogeology/\n"
+    printf "\t                      isis, examples include \"LTS\", \"dev\", and \"RC\"\n"
+    printf "\t-v, --isis-version    Different ISIS versions as defined by the\n"
+    printf "\t                      anaconda versions for a label at\n"
+    printf "\t                      https://anaconda.org/usgs-astrogeology/isis,\n"
+    printf "\t                      examples include 8.0.3 (LTS/Feature),\n"
+    printf "\t                      8.2.0_RC1 (RC), and 2025.02.22 (dev)\n"
+    printf "\t-n, --env-name        The name of the anaconda environment to create.\n"
+    printf "\t-m, --miniforge-dir   Define the directory to an anaconda package\n"
+    printf "\t                      manager install location. If you have an\n"
+    printf "\t                      anaconda package manager already this\n"
+    printf "\t                      argument will be ignored. If not a version\n"
+    printf "\t                      of miniforge will be installed at this\n"
+    printf "\t                      location\n"
+    printf "\t-p, --data-prefix     The directory where ISISDATA is located. If\n"
+    printf "\t                      this directory doesn't exist then one will\n"
+    printf "\t                      be made at its location\n"
+    printf "\t--no-data             Do not ask to download any data for the\n"
+    printf "\t                      ISIS_DATA area.\n"
+    printf "\n"
+    printf "\tDefining variables on the command line will skip the\n"
+    printf "\tinteractive elements within this script\n"
+    printf "\n\n"
+}
+
+for arg in "$@"; do
+    case $arg in
+        -h|--help)
+            print_help
+            exit 0
+            ;;
+    esac
+done
+
+POSITIONAL_ARGS=()
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        -l|--anaconda-label)
+            check_valid_arg $1 $2
+            ANACONDA_LABEL="$2"
+            shift # past argument
+            shift # past value
+            ;;
+        -v|--isis-version)
+            check_valid_arg $1 $2
+            ISIS_VERSION="$2"
+            shift # past argument
+            shift # past value
+            ;;
+        -n|--env-name)
+            check_valid_arg $1 $2
+            ENV_NAME="$2"
+            shift # past argument
+            shift # past value
+            ;;
+        -m|--miniforge-dir)
+            check_valid_arg $1 $2
+            MINIFORGE_DIR="$2"
+            shift # past argument
+            shift # past value
+            ;;
+        -p|--data-prefix)
+            check_valid_arg $1 $2
+            ISISDATA_PREFIX="$2"
+            shift # past argument
+            shift # past value
+            ;;
+        --no-data)
+            DOWNLOAD_DATA=NO
+            shift # past argument
+            ;;
+        -*|--*)
+            echo "Unknown option $1"
+            exit 1
+            ;;
+        *)
+            POSITIONAL_ARGS+=("$1") # save positional arg
+            shift # past argument
+            ;;
+    esac
+done
+
+set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters
+
+# If an ANACONDA_LABEL was not set, ask the user for it
+if [[ -z $ANACONDA_LABEL ]]; then
+    ANACONDA_LABEL="main"
+    printf "ISIS anaconda label [$ANACONDA_LABEL] will be installed.\n"
+    printf "\\n"
+    printf "  - Press ENTER to confirm the ISIS label to install\\n"
+    printf "  - Press CTRL-C to abort the installation\\n"
+    printf "  - Or specify a different ISIS label below [main, dev, lts, rc]\\n"
+    printf "  - Default is [main]\\n"
+    printf "\\n"
+    printf "[%s] >>> " "$ANACONDA_LABEL"
+
+    read -r anaconda_label
+
+    # If a label was given, set it
+    if [ -n "$anaconda_label" ]; then
+        ANACONDA_LABEL=$anaconda_label
+    fi
+fi
+
+# Force label to uppercase for LTS and RC
+if [ "${ANACONDA_LABEL,,}" = "lts" ] || [ "${ANACONDA_LABEL,,}" = "rc" ]; then
+    ANACONDA_LABEL="${ANACONDA_LABEL^^}"
+fi
+
+printf "\nISIS anaconda label set to [$ANACONDA_LABEL]\n"
+
+if [ "$ANACONDA_LABEL" = "main" ]; then
+    PACKAGE_NAME="usgs-astrogeology::isis"
+else
+    PACKAGE_NAME="usgs-astrogeology/label/$ANACONDA_LABEL::isis"
+fi
+
+# If an ISIS_VERSION was not set, ask the user for it
+if [[ -z $ISIS_VERSION ]]; then
+    ISIS_VERSION="latest"
+    # Skip the version prompt if ANACONDA_LABEL is dev
+    if [ "$ANACONDA_LABEL" != "dev" ]; then
+        printf "Specify an ISIS version to install from the ISIS anaconda label [$ANACONDA_LABEL]\n"
+        printf "\\n"
+        printf "  - Press ENTER to install the latest version of ISIS found under [$ANACONDA_LABEL]\\n"
+        printf "  - Press CTRL-C to abort the installation\\n"
+        printf "  - Or specify an ISIS version below\\n"
+        printf "\\n"
+        printf "[%s] >>> " "$ISIS_VERSION"
+
+        read -r isis_version
+
+        # If a label was given, set it
+        if [ -n "$isis_version" ]; then
+            ISIS_VERSION=$isis_version
+        fi
+        printf "\nISIS version set to [$ISIS_VERSION]\n"
+    else
+        printf "\nUsing latest version from dev channel\n"
+        ISIS_VERSION="latest"
+    fi
+else
+    printf "\nISIS version set to [$ISIS_VERSION]\n"
+fi
+
+if [ $ISIS_VERSION = "latest" ]; then
+    ISIS_VERSION=""
+fi
+
+printf "\nBeginning install of $PACKAGE_NAME\n\n"
+
+# Determine the OS type
+case "$(uname)" in
+    "Linux")
+        MINIFORGE_INSTALLER="Miniforge3-Linux-x86_64.sh"
+        LIBGL_INSTALL=""
+        ;;
+    "Darwin")
+        MINIFORGE_INSTALLER="Miniforge3-MacOSX-x86_64.sh"
+        LIBGL_INSTALL=""
+        export CONDA_SUBDIR=osx-64
+        ;;
+    *)
+        echo "Unsupported OS: $(uname)"
+        exit 1
+        ;;
+esac
+
+# Check if conda is installed but mamba is not
+if command -v conda &> /dev/null && ! command -v mamba &> /dev/null; then
+    echo "WARNING: conda is installed but mamba is not."
+    echo "mamba is a faster alternative to conda and is required for ISIS installation."
+    echo "Otherwise, environment creation will take much longer than what is practical."
+    echo "Would you like to install mamba? [yes|no]"
+    ans="yes"
+    printf "[%s] >>> " "$ans"
+    
+    read -r ans
+    
+    # If no input, use default
+    if [ "$ans" = "" ]; then
+        ans="yes"
+    fi
+    
+    ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+    while [ "$ans" != "YES" ] && [ "$ans" != "NO" ]
+    do
+        echo "Please answer 'yes' or 'no':"
+        printf ">>> "
+        read -r ans
+        ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+    done
+    
+    if [ "$ans" == "NO" ]; then
+        echo "Error: Only mamba is supported. Exiting."
+        exit 1
+    else
+        echo "Will proceed with installing mamba."
+    fi
+fi
+
+# Install Miniforge if it's not already installed
+if ! command -v mamba &> /dev/null; then
+
+    echo "Miniforge not found, installing Miniforge..."
+
+    # If a MINIFORGE_DIR is not set, ask the user
+    if [ -z "$MINIFORGE_DIR" ]; then
+        MINIFORGE_DIR="$HOME/miniforge3"
+
+        printf "Miniforge will be installed at this location:\n\n" 
+        printf "\t$MINIFORGE_DIR/\n\n"
+        printf "\\n"
+        printf "  - Press ENTER to confirm the Miniforge install location\\n"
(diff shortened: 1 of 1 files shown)


isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -h|--help)
+      print_help
+      ;;
+    -n|--env_name)
+      ENV_NAME="$2"
+      shift # past argument
+      shift # past value
```
dpmayerUSGS: Only need one `shift` for command line options that take a single argument. The 2 `shifts` are actually pushing beyond the option and argument in a way that confuses subsequent parsing of the command line.

This can be resolved by deleting Lines 50,55,60

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+    printf "\tDefining variables on the command line will skip the interactive "
+    printf "elements within this script"
+    printf "\n\n"
+    printf "\t\$1 ISIS_LABEL - Different ISIS labels as defined by the "
+    printf "anaconda labels at https://anaconda.org/usgs-astrogeology/isis"
+    printf ", examples include \"LTS\", \"dev\", and \"RC\"\n"
+    printf "\t\$2 ISIS_VERSION - Different ISIS versions as defined by the "
+    printf "anaconda versions for a label at https://anaconda.org/usgs-astrogeology/isis"
+    printf ", examples include 8.0.3 (LTS/Feature), 8.2.0_RC1 (RC), and 2025.02.22 (dev)"
+    exit 1
```
dpmayerUSGS: Consider omitting the `exit` from the `print_help()` function so that exit can be called immediately after invoking the function but using an exit code appropriate for the context. For example, if the user invokes the script with the `-h` option, the script should print the help message and exit normally (code 0) because there is no error.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+    *)
+      POSITIONAL_ARGS+=("$1") # save positional arg
+      shift # past argument
+      ;;
+  esac
+done
+
+set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters
+
+# If a ISIS LABEL was given, set it
```
dpmayerUSGS: `ISIS_LABEL` and `ISIS_VERSION` are not being handled as intended. If a user only intends to specify an ISIS version and therefore only passes a single positional parameter, the script will set that parameter as the value of `ISIS_LABEL`, not `ISIS_VERSION`.

By convention, positional parameters are required. In this script, `ISIS_LABEL` and `ISIS_VERSION` are meant to be optional, and thus should be added as command line options that each take a single argument.

Also: consider naming `ISIS_LABEL` something else. When I see "ISIS label," I think of detached labels associated with an ISIS cube file, not the label on a package in conda.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+        # If input was given, set it
+        if [ -n "$miniforge_install_path" ]; then
+            MINIFORGE_DIR=$miniforge_install_path
+        fi
+    fi
+
+    curl -kL "https://github.com/conda-forge/miniforge/releases/latest/download/$MINIFORGE_INSTALLER" -o Miniforge3.sh || failed_command "Miniforge download"
+    bash Miniforge3.sh -b -p $MINIFORGE_DIR || failed_command "Miniforge installation"
+    export PATH="$MINIFORGE_DIR/bin:$PATH"
+    mamba init || exit 1
```
dpmayerUSGS: Why `mamba` and not `conda` here?

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+fi
+
+# Check if the environment already exists 
+if conda env list | grep -qE "^$ENV_NAME[ ]."; then 
+    printf "\nEnvironment \"$ENV_NAME\" already exists. Not performing any updates.\n" 
+    printf "To delete the old environment, use the following commands:\n\n" 
+    printf "\t$ conda deactivate\n" 
+    printf "\t$ conda remove -n $ENV_NAME --all\n\n" 
+else
+    conda config --set ssl_verify false
```
dpmayerUSGS: Turning off SSL verification in conda is a bad security practice. I know there are unusual situations where this is necessary, but I don't think it's appropriate to do here, especially without warning a user.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+    ENV_PATH="$MINIFORGE_DIR"
+else
+    ENV_PATH="$MINIFORGE_DIR/envs/$ENV_NAME"
+fi
+
+DOWNLOAD_ISIS_DATA_SCRIPT="$ENV_PATH/bin/downloadIsisData"
+if [[ ! -f "$ENV_PATH/bin/downloadIsisData" && ! -f "$ENV_PATH/etc/isis/rclone.conf" ]]; then
+    if ! [ -f "$DOWNLOAD_ISIS_DATA_SCRIPT" ]; then
+        printf "\nInstalled download script: $DOWNLOAD_ISIS_DATA_SCRIPT\n"
+        curl  --output "$ENV_PATH/bin/downloadIsisData" -LJO --insecure https://github.com/USGS-Astrogeology/ISIS3/raw/dev/isis/scripts/downloadIsisData
```
dpmayerUSGS: running `curl --insecure` is bad security practice. Is this really necessary here?

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+            mkdir -p $ENV_PATH/etc/isis/ || failed_command "Creating $ENV_PATH/etc/isis/"
+        fi
+        curl --output "$ENV_PATH/etc/isis/rclone.conf" -LJO --insecure https://github.com/USGS-Astrogeology/ISIS3/raw/dev/isis/config/rclone.conf
+    fi
+else
+    printf "\nFound download script: $DOWNLOAD_ISIS_DATA_SCRIPT\n"
+fi
+
+if [ "$DOWNLOAD_DATA" = "NO" ]; then
+    exit 1
```
dpmayerUSGS: Should be `exit 0` because opting to not download data is normal and not an error.

dpmayerUSGS (changes requested):
Some genuine bugs in the argument parser, but otherwise mostly stylistic/convention considerations.

isis/scripts/install_isis.sh
acpaquette: I think this may be a failure on my end. Some of the arguments aren't flags but expect a value after them. So the `--env_name` arg expects the user to provide a name after the arg. Something like:
```
install_isis.sh -n isis_lts
```

In this case, it seems necessary to do two shifts

isis/scripts/install_isis.sh
acpaquette: To some degree they are interchangeable

isis/scripts/install_isis.sh
acpaquette: Likely a cert issue on my end

isis/scripts/install_isis.sh
acpaquette: Likely a cert issue on my end

isis/scripts/install_isis.sh
dpmayerUSGS: If the user formats the options correctly (`--option argument`), then it doesn't matter whether there is 1 `shift` or 2, but if the user passes an option that should have an argument (`--option`), then the parser as written will grab the next thing on the command line as the argument.

So if the user passes `-m -p my_isis_prefix`, the script will set `MINIFORGE_DIR="-p"` (which isn't valid), and will set `ISIS_PREFIX=""` which is valid, but not the intention.

If the second `shift` is removed on those 3 lines, following the example above the script will still set  `MINIFORGE_DIR="-p"` but will at least set `ISIS_PREFIX=my_isis_prefix` correctly. 

The real issue here may be a need to do some sanity-checking to make sure that options which require an argument have a sane argument and throw an error if they don't. Using a single `shift` should make it easier to identify which option the user forgot to add an argument to.

isis/scripts/install_isis.sh
dpmayerUSGS: Fair enough. The script seems to use both `mamba` and `conda`. Consider picking one and using consistently throughout even though they are interchangeable in this application.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+
+# Verify if the folder exists
+if ! [ -d "$ISISDATA_PREFIX" ]; then
+    if [ "$DOWNLOAD_DATA" = "YES" ]; then
+        echo "Creating folder $ISISDATA_PREFIX"
+        mkdir -p $ISISDATA_PREFIX || failed_command "Creating $ISISDATA_PREFIX"
+    else
+        echo "Not creating directory $ISISDATA_PREFIX since the --no_data flag was set"
+    fi
+fi
```
Kelvinrr: Doing this, it did not expand `~` to home, might be something we want to handle. Using `$HOME` also resulted in creating a local directory called `$HOME` rather than expanding the directory.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,355 @@
...
+    printf "\tdownloadIsisData base \$ISISDATA"
+fi
+
+## Consider offering user the option to install mission spefific data areas at this point:
+printf "\n\n"
+printf "Do you want to install mission-specific ISISDATA now? This can be done later. [yes|no]\n"
+ans="no"
+printf "[%s] >>> " "$ans"
+
+read -r ans
```
Kelvinrr: If users say no to wanting to install ISISDATA, this should also be skipped.

isis/scripts/install_isis.sh
acpaquette: This should be resolved now!

isis/scripts/install_isis.sh
acpaquette: I am a little torn on this. If you say no to base, does it follow that you won't want any of the mission data?

Kelvinrr: considering you can't do anything without base, I imagine so, right? Unless they already have base for some reason. I can't imagine being a common case I guess.

dpmayerUSGS: Yeah, the use case would be if they already have base installed and now want to install mission-specific data. Probably not a common case. 

If they do have base installed already and want to install mission data, it should be pretty harmless to run sync over base again (because there won't be anything to download) before getting to the mission-specific stuff. If a user has a more exotic case (perhaps they have customized their base data area in some way), they probably have sufficient technical skill to install and manipulate data areas without this shell script. 

I  would err on the side of simplicity here and not prompt the user to download mission data if they decline to download base.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,416 @@
...
+# If an ANACONDA_LABEL was not set, ask the user for it
+if [[ -z $ANACONDA_LABEL ]]; then
+    ANACONDA_LABEL="main"
+    printf "ISIS anaconda label [$ANACONDA_LABEL] will be installed.\n"
+    printf "\\n"
+    printf "  - Press ENTER to confirm the ISIS label to install\\n"
+    printf "  - Press CTRL-C to abort the installation\\n"
+    printf "  - Or specify a different ISIS label below\\n"
+    printf "\\n"
+    printf "[%s] >>> " "$ANACONDA_LABEL"
```
Kelvinrr: might want to include dev, main and its as examples here. Since someone who ran this straight probably didn't run -h

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,416 @@
...
+    printf "Usage: $0 [options]\n"
+    printf "Options:\n"
+    printf "\t-h, --help            Show this help message and exit\n"
+    printf "\t-l, --anaconda_label      Different ISIS labels as defined by the anaconda labels "
+    printf " at https://anaconda.org/usgs-astrogeology/isis, examples include \"LTS\", \"dev\", "
+    printf "and \"RC\"\n"
+    printf "\t-v, --isis_version    Different ISIS versions as defined by the anaconda versions "
+    printf "for a label at https://anaconda.org/usgs-astrogeology/isis, examples include 8.0.3 "
+    printf "(LTS/Feature), 8.2.0_RC1 (RC), and 2025.02.22 (dev)\n "
+    printf "\t-n, --env_name        The name of the anaconda environment to create.\n"
```
Kelvinrr: I get unknown option: 

```bash 
(isisdevinstall) igswza281l2305:~ krodriguez$ bash install_isis.sh -l dev -n isisbnash
Unknown option -n
```

isis/scripts/install_isis.sh
acpaquette: I left the two shifts in and did some basic handling if someone provides a flag that expects a value but none are provided

Kelvinrr:
@dpmayerUSGS fixed a few bugs, mind re-reviewing this?

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+echo "ISISDATA path is currently set to $ISISDATA_PREFIX"
+echo "You can do this later, read more at:" 
+printf "\n\thttps://astrogeology.usgs.gov/docs/how-to-guides/environment-setup-and-maintenance/isis-data-area/\n\n"
+echo "Do you want to install base ISISDATA now? This can be done later. [yes|no]"
+ans="no"
+printf "[%s] >>> " "$ans"
+
+read -r ans
+
+# If no input, use default path
```
dpmayerUSGS: This `if` block seems unnecessary because the script defines `ans="no"` as default a few lines above. I think it's okay to keep it if you want to be "belt and suspenders," but maybe update the comment text; the block doesn't have anything to do with a "default path."

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+
+if [ "$ans" == "YES" ]; then
+    printf "\n"
+    printf "Do you want to install mission-specific ISISDATA now? This can be done later. [yes|no]\n"
+    ans="no"
+    printf "[%s] >>> " "$ans"
+
+    read -r ans
+
+    # If no input, use default path
```
dpmayerUSGS, line 393: ~This `if` block seems unnecessary because the script defines `ans="no"` as default a few lines above. I think it's okay to keep it if you wantto be "belt and suspenders," but maybe update the comment text; the block doesn't have anything to do with a "default path."~

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+        echo "Miniforge directory exists at $MINIFORGE_DIR, reusing existing installation"
+    else
+        curl -kL "https://github.com/conda-forge/miniforge/releases/latest/download/$MINIFORGE_INSTALLER" -o Miniforge3.sh || failed_command "Miniforge download"
+        bash Miniforge3.sh -b -p $MINIFORGE_DIR || failed_command "Miniforge installation"
+    fi
+
+    $MINIFORGE_DIR/bin/$CLIENT init || exit 1
+else
+    echo "Miniforge is already installed."
+    if [[ "$CONDA_PREFIX" == *"envs"* ]]; then
```
dpmayerUSGS: Is this trying to look for a directory named `/envs/` somewhere in `$CONDA_PREFIX`? If yes, consider changing the pattern to `*"/envs/"*` to the pattern more specific.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+        echo "Creating folder $ISISDATA_PREFIX"
+        mkdir -p $ISISDATA_PREFIX || failed_command "Creating $ISISDATA_PREFIX"
+    else
+        echo "Not creating directory $ISISDATA_PREFIX since the --no_data flag was set"
+    fi
+fi
+
+$MINIFORGE_DIR/bin/$CLIENT env config vars set -n $ENV_NAME ISISDATA=$ISISDATA_PREFIX ISISROOT=$MINIFORGE_DIR/envs/$ENV_NAME || failed_command "Mamba config var set"
+
+if [[ "$MINIFORGE_DIR" == *"envs/$ENV_NAME"* ]]; then
```
dpmayerUSGS: Consider inserting `/` before `envs` to make the pattern more specific.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+    echo "WARNING: conda is installed but mamba is not."
+    echo "mamba is a faster alternative to conda and is recommended for ISIS installation."
+    echo "Would you like to install mamba? [yes|no]"
+    ans="yes"
+    printf "[%s] >>> " "$ans"
+    
+    read -r ans
+    
+    # If no input, use default
+    if [ "$ans" = "" ]; then
```
dpmayerUSGS, line 216: ~This `if` block seems unnecessary because the script defines `ans="no"` as default a few lines above.~

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+        ans="yes"
+    fi
+    
+    ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+    while [ "$ans" != "YES" ] && [ "$ans" != "NO" ]
```
dpmayerUSGS, line 221: This `while` seems unnecessary because the script defines `ans=no` by default above.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+    *)
+        echo "Unsupported OS: $(uname)"
+        exit 1
+        ;;
+esac
+
+# Check if conda is installed but mamba is not
+if command -v conda &> /dev/null && ! command -v mamba &> /dev/null; then
+    echo "WARNING: conda is installed but mamba is not."
+    echo "mamba is a faster alternative to conda and is recommended for ISIS installation."
```
dpmayerUSGS: The word "recommended" seams to weak here. If the user passes "no" in response to the prompt, the `if` block on Line 221 prints a message that only mamba is supported and the script exits. 

The prompt should be clear that mamba is actually _required_ or the script should be modified to proceed with an existing install of conda without mamba.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+
+read -r ans
+
+# If no input, use default path
+if [ "$ans" = "" ]; then
+    ans="no"
+fi
+
+ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+while [ "$ans" != "YES" ] && [ "$ans" != "NO" ]
```
dpmayerUSGS: This `while` seems unnecessary because the script defines `ans=no` by default above.

isis/scripts/install_isis.sh
```diff
@@ -0,0 +1,472 @@
...
+
+    read -r ans
+
+    # If no input, use default path
+    if [ "$ans" = "" ]; then
+        ans="no"
+    fi
+
+    ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+    while [ "$ans" != "YES" ] && [ "$ans" != "NO" ]
```
dpmayerUSGS, line 399: ~This `while` seems unnecessary because the script defines `ans=no` by default above.~

dpmayerUSGS (changes requested):
I only reviewed `isis/scripts/install_isis.sh`.

I think the script could be slimmed down. There appear to be some unnecessary code blocks.

One of the user prompts says that `mamba` is _recommended_ , but the script will exit with an error if the user has `conda` installed but declines to install `mamba`. What is the intended behavior?

isis/scripts/install_isis.sh
Kelvinrr: I don't think this is true, since user input will overwrite that default to whatever was put in. We can remove it and check for empty string later down the line. Otherwise, if they just press enter, `ans` will be an empty string. 

This applies to all the similar comments.

dpmayerUSGS: Okay, I see what you mean now. 
The repeated handling of yes/no responses could probably be abstracted away into a function that is called where needed rather than repeating the same `if` block though. I would tighten the code up a bit.

isis/scripts/install_isis.sh
Kelvinrr: similar to my previous comment, user input overwrites that value and we need a loop to make sure they didn't input something ambiguous

Kelvinrr:
Addressed comments, seems this needs to be rebased which I can do after these comments are resolved.

dpmayerUSGS (commented):
Only reviewed `isis/scripts/install_isis.sh`
Changes look good.

Kelvinrr (approved):

Issue #5821, opened 2025-05-29
Jigsaw failure with control points


oleg-alexandrov:
**ISIS version(s) affected**: 8.3.0

**Description**  

When running jigsaw, it fails with the error:

**USER ERROR** in method ControlPoint::ComputeApriori(). ControlPoint [point_68644] has no measures which project to the body.

It looks that a control point is being computed via triangulation, and it fails to work. 

Such occurrences are not unexpected. I got this with about 50 images of JunoCam for Europa. Some images are acquired in quick succession with little baseline. Some rays may fail to intersect, or intersect in an inaccurate location. Sometimes things start good, but cameras move a little and rays no longer intersect.

This should be a warning, instead. The function ControlPoint::ComputeApriori(), where this happens, checks for success or failure. So, this should return failed status for this particular control point, but go on with the rest of the work.

Issue #5818, opened 2025-05-28
Jigsaw failure


oleg-alexandrov:
**ISIS version(s) affected**: 8.3.0 and 9.0.0

I have what I believe is a legit ISIS control network for Mars with two cameras and a stereo convergence angle of 9.3 degrees. It gets read in by jigsaw, and it runs very well as:

`jigsaw fromlist=run/list.txt update=no twist=yes radius=yes maxits=3 cnet=run/run-cnet.net onet=run/run-jigsaw.net file_prefix=run/run
`
Yet, as soon as I increase the number of iterations from 3 to 4 (maxits=4), I get the error:
```

 aborting...**USER ERROR** Unable to bundle adjust network [run/run-cnet.net].
**PROGRAMMER ERROR** Could not solve bundle adjust.
**PROGRAMMER ERROR** Negative distances are not supported, the value [-193725241.51332 meters] cannot be stored in the Distance class.
(isis9.0.0) 
```

Why would that be? I got this with both Mars (MOC) and Europa (JunoCam) data.

Here are the interest point matches, and they look good:

<img width="700" alt="Image" src="https://github.com/user-attachments/assets/63a76d4d-b87c-4005-9259-7a9de4df7a5e" />

The most similar issue is https://github.com/DOI-USGS/ISIS3/issues/4111, but that doesn't help figure out what is going on.

Note: The control network is created with ASP's bundle_adjust, but it uses the ISIS API for how to set up and save the measurements, control points, their sigma, etc.

I tested this with ISIS 8.3.0 and ISIS 9.0.0.

Self-contained testcase reproducing the problem, having all inputs, outputs, and the command to run, with a readme, is attached.

Note1: It runs just fine with twist=yes radius=no, or with twist=no radius=yes. Only when these are both set to yes it fails.

Note2: This still fails if adding:

`POINT_RADIUS_SIGMA=10000000
`

but it works with:

`POINT_RADIUS_SIGMA=1000000
`

All this suggests the control network is likely legit, but the solver goes wild.

[jigsaw_bug.tar.gz](https://github.com/user-attachments/files/20471137/jigsaw_bug.tar.gz)

kledmundson:
Hi @oleg-alexandrov. Only took a quick look, so won't swear by what I'm seeing yet, but I think when it's failing spectacularly that you haven't addressed the datum defect of 7. And it goes crazy, as I think I would expect.

[...]

PR #5833, opened 2025-06-13
Jigsaw fixes

oleg-alexandrov:
## Description

These are proposed fixes for jigsaw issues https://github.com/DOI-USGS/ISIS3/issues/5821 and https://github.com/DOI-USGS/ISIS3/issues/5818.

 - When an invalid control point is encountered, it should be ignored with a warning, rather than the program failing.
 - When solving for both twist and radius, and not setting point radius sigma, the tool will print a warning about the problem being under-constrained. This is not a fatal failure, for now, as I don't fully understand the implications of prohibiting this mode, but usually the program will fail soon enough in either case if the solution goes wild, and then the warning at least will explain why that may be.
  
I tested with a control network I had.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/control/apps/jigsaw/jigsaw.cpp
+++ b/isis/src/control/apps/jigsaw/jigsaw.cpp
@@ -224,7 +224,7 @@ namespace Isis {
                                                    under review and is likely resulting \
                                                    in addition error in the bundle adjust. \
                                                    We recommend that you do not solve for radii at this moment."));
-        Application::Log(radiusSolveWarning);
+        Application::AppendAndLog(radiusSolveWarning, log);
       }
     }
     settings->setCubeList(cubeList);
--- a/isis/src/control/apps/jigsaw/jigsaw.cpp
+++ b/isis/src/control/apps/jigsaw/jigsaw.cpp
@@ -209,7 +209,7 @@ namespace Isis {
         "failure may result.";
       PvlGroup radiusSolveWarning("RadiusSolveWarning");
       radiusSolveWarning.addKeyword(PvlKeyword("Warning", msg.c_str()));
-      Application::AppendLog(radiusSolveWarning, log);
+      Application::AppendAndLog(radiusSolveWarning, log);
     }
 
     QString cnetFile = ui.GetFileName("CNET");
--- a/isis/src/control/apps/jigsaw/jigsaw.cpp
+++ b/isis/src/control/apps/jigsaw/jigsaw.cpp
@@ -202,6 +202,19 @@ namespace Isis {
       throw IException(IException::User, msg, _FILEINFO_);
     }
 
+    std::cout << "--twist is: " << (ui.GetBoolean("TWIST") ? "true" : "false") << std::endl;
+    std::cout << "radius is : " << (ui.GetBoolean("RADIUS") ? "true" : "false") << std::endl;
+    std::cout << "radius sigma " << ui.WasEntered("POINT_RADIUS_SIGMA") << std::endl;
+    if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") && 
+        !ui.WasEntered("POINT_RADIUS_SIGMA")) {
+      string msg = "If solving for both twist and radius, must set a value "
+        "for point_radius_sigma, as otherwise the problem is under-constrained and a "
+        "failure may result.";
+      PvlGroup radiusSolveWarning("RadiusSolveWarning");
+      radiusSolveWarning.addKeyword(PvlKeyword("Warning", msg.c_str()));
+      Application::Log(radiusSolveWarning);
+    }
+
     QString cnetFile = ui.GetFileName("CNET");
 
     // retrieve settings from jigsaw gui
--- a/isis/src/control/apps/jigsaw/jigsaw.cpp
+++ b/isis/src/control/apps/jigsaw/jigsaw.cpp
@@ -202,14 +202,14 @@ namespace Isis {
       throw IException(IException::User, msg, _FILEINFO_);
     }
 
-    if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") && 
+    if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") &&
         !ui.WasEntered("POINT_RADIUS_SIGMA")) {
       string msg = "If solving for both twist and radius, must set a value "
         "for point_radius_sigma, as otherwise the problem is under-constrained and a "
         "failure may result.";
       PvlGroup radiusSolveWarning("RadiusSolveWarning");
       radiusSolveWarning.addKeyword(PvlKeyword("Warning", msg.c_str()));
-      Application::Log(radiusSolveWarning);
+      Application::AppendLog(radiusSolveWarning, log);
     }
 
     QString cnetFile = ui.GetFileName("CNET");
--- a/isis/src/control/apps/jigsaw/jigsaw.cpp
+++ b/isis/src/control/apps/jigsaw/jigsaw.cpp
@@ -202,9 +202,6 @@ namespace Isis {
       throw IException(IException::User, msg, _FILEINFO_);
     }
 
-    std::cout << "--twist is: " << (ui.GetBoolean("TWIST") ? "true" : "false") << std::endl;
-    std::cout << "radius is : " << (ui.GetBoolean("RADIUS") ? "true" : "false") << std::endl;
-    std::cout << "radius sigma " << ui.WasEntered("POINT_RADIUS_SIGMA") << std::endl;
     if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") && 
         !ui.WasEntered("POINT_RADIUS_SIGMA")) {
       string msg = "If solving for both twist and radius, must set a value "
--- a/isis/src/control/objs/ControlPoint/ControlPoint.cpp
+++ b/isis/src/control/objs/ControlPoint/ControlPoint.cpp
@@ -968,6 +968,7 @@ namespace Isis {
       msg += "no measures which project to the body. Will ignore it.";
       SetIgnored(true);
       std::cerr << msg << std::endl;
+      adjustedSurfacePoint = aprioriSurfacePoint;
       return Success;
     }
 
--- a/isis/src/control/objs/ControlPoint/ControlPoint.cpp
+++ b/isis/src/control/objs/ControlPoint/ControlPoint.cpp
@@ -962,12 +962,13 @@ namespace Isis {
 
     // if point is Free, we continue to compute a priori coordinates
 
-    // if no good measures, we're done
-    // TODO: is the message true/meaningful?
+    // If no good measures, this point is ignored
     if (goodMeasures == 0) {
-      QString msg = "in method ControlPoint::ComputeApriori(). ControlPoint [" + GetId() + "] has ";
-      msg += "no measures which project to the body";
-      throw IException(IException::User, msg, _FILEINFO_);
+      QString msg = "In method ControlPoint::ComputeApriori(): ControlPoint [" + GetId() + "] has ";
+      msg += "no measures which project to the body. Will ignore it.";
+      SetIgnored(true);
+      std::cerr << msg << std::endl;
+      return Success;
     }
 
     // Compute the averages if all coordinates are free
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ ctest FunctionalTestJigsawApollo to validate this output. [#5710](https://github
 - Changed `StripPolygonSeeder` and `GridPolygonSeeder` to seed individual polygons within each images multipolygon footprint [#5193](https://github.com/DOI-USGS/ISIS3/issues/5193)
 
 ### Fixed
+- Fixed a failure for jigsaw crashing when encountering an invalid control point. Now it is ignored. Added a warning when the problem is underconstrained. 
 - Fixed kaguyatc2isis invalid BandBin values [#5629](https://github.com/DOI-USGS/ISIS3/issues/5629)
 - Fixed SpiceClient to handle redirect requests.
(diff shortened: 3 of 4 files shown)


isis/src/control/apps/jigsaw/jigsaw.cpp
```diff
@@ -202,6 +202,16 @@ namespace Isis {
...
     }
 
+    if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") && 
+        !ui.WasEntered("POINT_RADIUS_SIGMA")) {
+      string msg = "If solving for both twist and radius, must set a value "
+        "for point_radius_sigma, as otherwise the problem is under-constrained and a "
+        "failure may result.";
+      PvlGroup radiusSolveWarning("RadiusSolveWarning");
+      radiusSolveWarning.addKeyword(PvlKeyword("Warning", msg.c_str()));
+      Application::Log(radiusSolveWarning);
```
acpaquette: This should use `Application::AppendAndLog` for CLI and application compatibility

isis/src/control/objs/ControlPoint/ControlPoint.cpp
```diff
@@ -962,12 +962,14 @@ namespace Isis {
...
-    // TODO: is the message true/meaningful?
+    // If no good measures, this point is ignored
     if (goodMeasures == 0) {
-      QString msg = "in method ControlPoint::ComputeApriori(). ControlPoint [" + GetId() + "] has ";
-      msg += "no measures which project to the body";
-      throw IException(IException::User, msg, _FILEINFO_);
+      QString msg = "In method ControlPoint::ComputeApriori(): ControlPoint [" + GetId() + "] has ";
+      msg += "no measures which project to the body. Will ignore it.";
+      SetIgnored(true);
+      std::cerr << msg << std::endl;
```
acpaquette, line 970: I like the warning but for most users this will still dump to the terminal, or most won't pipe the error message to another output. We need a way to silence the warnings in the preference file or some other setting. @Kelvinrr Thoughts on this? I know we had issues when we added a deprecation warning for users, not sure what discussions came out of that

isis/src/control/apps/jigsaw/jigsaw.cpp
```diff
@@ -202,6 +202,16 @@ namespace Isis {
       throw IException(IException::User, msg, _FILEINFO_);
     }
 
+    if (ui.GetBoolean("TWIST") && ui.GetBoolean("RADIUS") && 
+        !ui.WasEntered("POINT_RADIUS_SIGMA")) {
+      string msg = "If solving for both twist and radius, must set a value "
+        "for point_radius_sigma, as otherwise the problem is under-constrained and a "
+        "failure may result.";
```
acpaquette, line 209: Should this value be set to a default? It seems weird that enabling it but failing to set it would be allowed. Currently, ISIS doesn't set any value but defaults the sigma0 to 1e-10. Maybe this could share that default? Just a thought

acpaquette (commented):
Overall looks good! A couple of user facing questions that could use some discussion

kledmundson:
Quick comment regarding the underconstrained warning. We certainly need this in some form, but I suspect that this check is insufficient. I'm not sure that it takes into consideration constraints on individual control point coordinates that may satisfy (or more than satisfy) the datum defect of 7 (set for example in _cneteditor_ or _qnet_). Or constraints that have been placed on spk position and/or pointing.

isis/src/control/apps/jigsaw/jigsaw.cpp
oleg-alexandrov: A sigma of 1e-10 would effectively keep the point fixed. 

It is very hard to set good defaults for sigma, as it is dependent on the current camera configuration and the user also needs to do some guessing. I think no-default and having the user think of it may be better.

isis/src/control/objs/ControlPoint/ControlPoint.cpp
oleg-alexandrov: Sometimes many control points will fail to work. I am not sure what you have in mind, but we probably would not want a popup for each of them. This could even be silent, with failed points simply glossed over. Fine with me whichever way this would be dealt with.

isis/src/control/apps/jigsaw/jigsaw.cpp
oleg-alexandrov: Replaced with AppendAndLog(). Also found an instance of this just under this, where I copied the logic from to start with, so replaced there too.

oleg-alexandrov:
@kledmundson I am not familiar enough with the tool to say for sure. I guess user expertise is needed in either case, and if the user knows some constraints were already set, they may choose to set this one to a relaxed value or simply ignore the warning.

Any sensible use of this tool likely needs some ground constraint in either case (and likely constrain at least 3 points), as for very high focal length any constraint on camera position and orientation _alone_ may still result in too much of displacement on the ground.

isis/src/control/objs/ControlPoint/ControlPoint.cpp
Kelvinrr: yeah we almost universally get complaints about warnings, it can mess up people who parse app outputs which is very common since ISIS often returns PVL. It might be possible to use the log for this since that can be suppressed?

oleg-alexandrov: Can you suggest what this line should be changed to? There appear to be various choices in several places.

lwellerastro: Hi All - could someone please be more explicit about this what is being discussed here as this is a program my job depends on and changes have impact. Is this a situation where measures for a point fail to intersect the body? If so, these should not be silenced and the program either needs to fail outright (as it does now) and/or that information should go to a log file. I want to know about these situations as they indicate some sort of a problem and should be removed and/or my network building needs additional checks. More importantly I need to understand why it is happening and hiding the error does not give me that opportunity.

It is recommended that users run cnetcheck on their network prior to passing it to jigsaw which will identify measures that do not intersect the target via the nolatlon output file.

isis/src/control/apps/jigsaw/jigsaw.cpp
lwellerastro: No default is preferred as the value for any of the weights will vary depending on camera and target body.

isis/src/control/objs/ControlPoint/ControlPoint.cpp
oleg-alexandrov: With control networks I make in the ASP bundle_adjust program, I believe jigsaw starts to fail when the angle between rays is too narrow. I am not fully sure, but I know that if I filter out ahead of time points based on convergence angle, it goes through.

I agree the user should be told about such issues. I'd rather it not fail. The question is, how to log these properly, without pop-ups for example. 

Can one point out to a few lines of code used for warnings in such situations? Above, code like "PvlGroup radiusSolveWarning("RadiusSolveWarning")" is used, maybe something similar?

lwellerastro: I don't know anything about specific code to point out possible warning scenarios, but I don't find pop-ups (without failure) useful in the case of invalid control points. Some of my bundles run hours and in one case days, so I would prefer the program fail outright so that I can do something about the offending measures/points as soon as possible. Plus these longer runs are typically sent to a cluster where I am not monitoring messages to screen in real time. Finding out long after the fact that something is broken is not an efficient use of my time. 

I'm not sure how or where this information could be captured in any existing jigsaw report file. Plus I think changes like that or going to warning versus the current program failure is a breaking change (I'm not certain about the latter though). 

I would discourage depending on jigsaw to validate networks in the way that is being suggested since there are programs that already do this that should be run before passing a network to jigsaw. I can see how a measure may get bundled off the target during jigsaw, but I think that is a different scenario.

oleg-alexandrov: Why programs validate the control network before jigsaw?

And yes, what you point out is correct. Jigsaw itself could create problematic control points that it later itself would not be able to handle. One could have a network with millions of control points and some of the cameras may have borderline acceptable pointing info (such looking in an almost parallel way).

I think it is a sensible choice that if the tool cannot handle a control point, it should be skipped with a warning. Otherwise the user is stuck with a failure, and there seems to be no way around it. The user is expected to inspect warnings and make decisions.

lwellerastro: > Why programs validate the control network before jigsaw?
cnetstats, cnetcheck

oleg-alexandrov: Will cnetcheck go through all the same motions as jigsaw? Can then the failure message refer to these two?

lwellerastro: cnetcheck will report if the measure does not intersect body when nolatlon=True (default) and send the information to an output file (which is not the most user friendly format so you might need a script to parse the output to send to cnetedit to remove the offending measures). It doesn't iterate like jigsaw, it will only check the state of the network based on spice and the line/sample information of the measures (probably similar to what jigsaw does before it starts to iterate). But it will report on off target measures along with free points having single measures, points having nocontrol (because all measures are invalid (ignored), etc. 

cnetcheck is more robust that cnetstats.

It might be useful if jigsaw referred the user to cnetcheck in the case where a measure does not intersect the target body. But I'm not sure the particular bundle error I'm referring to for off-body targets is the same as what you are experiencing with your networks.

I won't go into all of the detail, but this is the jigsaw failure output for voyager data where measures essentially had no spice because they were off the target:

```
  Group = Error
    Program = jigsaw
    Class   = "USER ERROR"
    Code    = 2
    Message = "Unable to bundle adjust network [tb_vgr_voygalnet_Epos.net]"
    File    = main.cpp
    Line    = 157
  End_Group

  Group = Error
    Program = jigsaw
    Class   = "USER ERROR"
    Code    = 2
    Message = "Could not solve bundle adjust"
    File    = BundleAdjust.cpp
    Line    = 982
  End_Group

  Group = Error
    Program = jigsaw
    Class   = "USER ERROR"
    Code    = 2
    Message = "Unable to map apriori surface point for measure
               Voyager2/NAC/2064922 on point 133 into focal plane"
    File    = BundleAdjust.cpp
    Line    = 1887
  End_Group

```
I can't tell if this is the same sort of error you were encountering with your Juno data though.
I will add that for my Voyager and some Galileo disk data of Europa we attempted to avoid this sort of problem by manually moving the spice onto the target for images having issues. Very user intensive, but the process works (I can share those details in an email if you are interested and if applicable to your particular situation).

I have also encountered problems with Galileo SSI Europa, Cassini ISS Enceladus, Titan and MEX SRC and HRSC Phobos data where points end up right on the limb for images (having spice mostly on the target) because a measure is a fraction away of having spice or not. Cnetcheck has found those instances (reported in nolatlon output) and they are removed from the network prior to running the network through the bundle adjustment.  If this is more in line with what you are experiencing with you JunoCam data you might see if cnetcheck spots your known problems.

And so I understand better, in the suggested warning message
```
QString msg = "In method ControlPoint::ComputeApriori(): ControlPoint [" + GetId() + "] has ";
      msg += "no measures which project to the body. Will ignore it.";
      SetIgnored(true);
      std::cerr << msg << std::endl;
```
Does "ignore" mean literally setting a measure in the network to ignore or does it simply mean not using it in calculations during any or all iterations? 

If it is that latter (either really, the user needs to know if jigsaw is modifying the state of measures), maybe it could be flagged as "rejected" which would show up in the residuals.csv file in a similar way as measures are via the use of OUTLIER_REJECTION. But the user would need to know to look in the residuals file for rejected measures and be able to discern rejection due to outlier_rejection use versus rejection due to failing to intersect the target. That seems messy, so maybe ignore that idea. Which is why I think depending on jigsaw to validate a network should be discouraged and cnetcheck should be run to catch some of the most egregious issues before hand.

oleg-alexandrov: >Does "ignore" mean literally setting a measure in the network to ignore or does it simply mean not using it in calculations during any or all iterations?

The control point will be flagged as ignored in the control network, and will be skipped when forming the optimization problem.

>I will add that for my Voyager and some Galileo disk data of Europa we attempted to avoid this sort of problem by manually moving the spice onto the target for images having issues. Very user intensive, but the process works

This is likely not a great solution. Jigsaw should have the ability to filter out various problematic points without failing. It should print a report of the fraction of points that cannot be processed.

I rarely use this, so will defer to actual users and maintainers, but I think this is a case of a tool that should be more robust and give the user more choices of how to handle borderline data.

lwellerastro: > This is likely not a great solution. Jigsaw should have the ability to filter out various problematic points without failing. It should print a report of the fraction of points that cannot be processed.

No, not fun, but in the case of Voyager data of Europa there was no choice - the spice was completely off the target and had to be moved. Extreme case. 

I don't completely disagree that jigsaw should be capable of ignoring problem measures, reporting on it and continuing to iterate if possible. However it may be tough parsing the various possible errors (with some level of specificity) into a single new output report.

And I apologize for commenting on these possible changes so late in the process. I missed the original issue post and only saw the PR today.

oleg-alexandrov: > However it may be tough parsing the various possible errors (with some level of specificity) into a single new output report.

One has to start somewhere. Ignoring failed control points and adding a report file is a good place. But since regular users seem not to see it as a problem, I guess maybe it is fine, most of the time.

BTW, for JunoCam I ran into systematic problems in produced terrain models that were due to intrinsics. ASP can tune those, and we have a [doc](https://stereopipeline.readthedocs.io/en/latest/examples/junocam.html).

oleg-alexandrov:
This can be closed. Looks like jigsaw users know when they need to set constraints. Ideally a warning may be nice but the system to follow seems complicated and likely things are fine as is.

Issue #5984, opened 2026-03-05
cam2map and ASP integration


oleg-alexandrov:
**Description**  

Here starting a discussion on making cam2map being compatible with ASP (on Kelvin's advice).

The proposal seems to be to ensure cam2map and mapproject produce interchangeable products with the right options. The plan I have for now is the following:

 - Add new option asp_map with default of false. If true, it will use new logic.
 - Add a new option called dem with value a string. This dem should be prepped as expected by ISIS.
 - Take a projection string, either with the map option, or with useproj = true and PROJString.. The latter has some issues in my experiments with regular proj str and takes in only wkt I think, this will need looking into.

The produced cub file will have PVL strings with the following info

 BUNDLE_ADJUST_PREFIX = NONE                                                                                                                     
 CAMERA_MODEL_TYPE    = isis                                                                                                                     
 INPUT_IMAGE_FILE     = path to input .cub                                                                                                       
 DEM_FILE             = path to DEM               

These are for use of ASP. 

The precise image created with mapproject and cam2map should be on same grid and same values, up to numerical precision. The extent may differ by a pixel or two given that the two use different logic for that, but I don't think this would be a problem for practical use. I'll try to harmonize these.

ASP should be able to take in mapprojected images from cam2map and produce stereo results that are indistinguishable than the ones with its own mapproject. ASP expects a non-prepped DEM. Not sure what to do about this. Likely may have to have on-the-fly prep in ISIS while loading the un-prepped one.

Thoughts? I think this came from @jlaura if I remember right. Want to hear also what @acpaquette thinks given his work with projections.

Kelvinrr:
I wonder it is worth adding a ISD/State param as well to allow CSM camera model to be input seperately. This is something we plan to add to other aspects of ISIS and it might be good to start here. 

@acpaquette might have some insight on best way to pass in an SRS since he added that functionality to ISIS.

oleg-alexandrov:
Sure, can add a new ISD param to cam2map. If set, the camera will be used from the ISD and passed where needed. In the output, it will then have:

CAMERA_MODEL_TYPE=csm
CAMERA_FILE=isd.json

These are needed for ASP. 

I will see if there's other metadata in the output .cub that needs updating in that case.

acpaquette:
[...]

PR #5988, opened 2026-03-10
Have cam2map create results as ASP's mapproject

oleg-alexandrov:
## Description

This adds to cam2map the option asp_map=true that produces mapprojected cubes in agreement with ASP at each grid point. 

Documentation: https://stereopipeline.readthedocs.io/en/latest/tools/mapproject.html#isis-compatibility

Relevant issue:  https://github.com/DOI-USGS/ISIS3/issues/5984

I added a unit test. I will put some more. 

This does not bring over any classes from ASP. Projections are done with GDAL and the image processing with the ISIS logic.

It would be good to get some review on this as there is quite some new logic. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/apps/cam2map/AspMapProjection.cpp
+++ b/isis/src/base/apps/cam2map/AspMapProjection.cpp
@@ -21,7 +21,6 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
-#include <cstdint>
 #include <limits>
 #include <map>
 #include <memory>
@@ -508,6 +507,11 @@ public:
     m_cols = m_cube.sampleCount();
     m_rows = m_cube.lineCount();
     m_bands = m_cube.bandCount();
+    if (m_cols <= 0 || m_rows <= 0)
+      throw IException(IException::User,
+        "Raster file [" + filename + "] has zero-size dimensions (" +
+        std::to_string(m_cols) + "x" + std::to_string(m_rows) + ")",
+        _FILEINFO_);
     m_portal1 = std::make_shared<Portal>(1, 1, m_cube.pixelType());
     // 2x2 portal for bilinear; hotspot (0,0) per ISIS convention (0-based)
     m_portal2 = std::make_shared<Portal>(2, 2, m_cube.pixelType(), 0, 0);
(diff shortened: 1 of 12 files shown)


oleg-alexandrov:
Looks  like any pull request is failing at: ObservationPair.FunctionalTestJigsawOutputCsmState.

acpaquette:
@oleg-alexandrov That test is also failing on dev so this PR is clean

isis/src/base/apps/cam2map/AspMapProjection.cpp
```diff
@@ -0,0 +1,2083 @@
...
+  if (!useIsisProgress)
+    std::cout << "\r100%" << std::endl;
+
+  outCube.close();
+}
+
+// Load a CSM camera model from an ISD file and inject its state into the
+// cube in memory, so that cube->camera() returns a CSMCamera without
+// modifying the file on disk.
+void loadCsmIntoCube(const QString &isdFile, Cube *cube) {
```
acpaquette: @oleg-alexandrov This feels a bit odd to have. Is there a way we could just return the CSMCamera from this, rather than load the blob onto the cube?

A suggestion might be to add a new constructor on the `CSMCamera` that just takes the model as input. Something like `CSMCamera::CSMCamera(csm::RasterGM* model, cube &cube) : Camera(cube)`

oleg-alexandrov: The reason I chose this approach is because this is exactly what csminit does. It loads a cube and then loads CSM on top from blob.

So this is consistent with existing logic. The philosophy is that we always have a cube, but sometimes the camera info comes not from SPICE but from CSM.

acpaquette: Sure but it seems like one would just run `csminit` on the cube prior to using cam2map rather than pass the ISD. This is partially why I don't understand the idea behind passing the ISD via an argument and I think why `csminit` was created as a counterpart to `spiceinit`.

If we want to pass the ISD directly I think that would be your camera, so we shouldn't need to put it onto the cube before creating a camera from it. I understand that this uses existing infrastructure to create the camera, having some additional infrastructure to create the camera directly from an ISD/State would be a nice to have

oleg-alexandrov: So, the existing workflow that if only the cub is set, we read what is in the cub. Could be spice or csm. If, additionally  isd is passed in, then we read the camera from there. I think this is in line with what is requested above. 

I think current approach is rather straightforward. The cub is amended from isd if need be.

acpaquette: If you think it's okay as is then I will leave it at that.

The main issue I have is that working through the ISIS blob system results in a couple of extra parsing passes. Rather than just reading the ISD into a camera, we create the camera, export it back into a state string, then parse it again when creating the camera second time. It seems like you could avoid the export and second parse altogether.

One could create the `CSMCamera` then set the camera on the cube if we want to maintain that cameras are accessed through the associated cube.

oleg-alexandrov: To avoid this looks like need to add new constructor in CSMCamera. Notable change to rather central class. Is  it worth doing it at this time?

oleg-alexandrov: I added the new constructor. It was enough to call the existing init function with minor changes. This will help us when we support ISD in jigsaw as well. All cam2map tests pass.

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -2321,6 +2333,19 @@ namespace Isis {
...
 
 
+  /**
+   * Add a blob to the cube in memory without modifying the file on disk.
+   * This inserts the blob into the in-memory blob map and adds its PVL
+   * label to the cube label so that hasBlob() and read(Blob) find it.
+   *
+   * @param blob The blob to add
+   */
+  void Cube::addInMemoryBlob(Blob &blob) {
```
acpaquette: Can this function be removed?

oleg-alexandrov: Thanks, I forgot about the old one. Nobody else needs it so yeah, wiped.

oleg-alexandrov:
Getting auth error unrelated to the actual code.

acpaquette:
@oleg-alexandrov I think Amy fixed the auth issues for running tests. Could you rebase this off dev and push to get a clean test run?

oleg-alexandrov:
Thanks. Looks like the workflow needs approval.

acpaquette:
@oleg-alexandrov I think the auth token for our ci may have expired. Working on getting this back up and running

oleg-alexandrov:
The build went through with no failures.

acpaquette (approved):

PR #6039, opened 2026-05-05
Write projection string to GeoTIFF produced by cam2map

### LINKED ISSUE #6033: Cam2map missing projection info when using PROJ
**ISIS version(s) affected**: ISIS 10.0.0RC2

**Description**  

When running cam2map on an image, the output image does not contain any projection information when using GDALINFO. 

**How to reproduce**  

run cam2map on any image using the proj features, then run gdalinfo.  

**Possible Solution**  

We probably have to explcitly call setprojection. Consider only enabling this for non-cubes. e.g. GTIFFs. I imagine cubes would require a sidecart file. In the cube case, we might need to update GDAL to read the proj string. 

**Additional context**

oleg-alexandrov:
Write projection string to GeoTIFF produced by cam2map

## Description

This fixes #6033. When cam2map or any other geotiff writing logic (via Cube::writeLabels) writes a GeoTIFF cube, the proj string is always written. Before, this was true only when a projection string was set as argument.

This uses a function named PvlToWkt that was previously used only for writing a cam2map-ed cub in asp_map format. 

The documentation in cam2map.xml got expanded a bit to reflect that geotiffs can be inputs and outputs.

Manually ran cam2map on an MRO CTX cube (B17_016219_1978_XN_17N282W.cub) before and after the fix. 

Cross-checked against the GDAL ISIS3 driver which can convert a cam2map-ed cub to geotif. 

Added three unit tests to create a geotif with and without setting a projection string, and when writing a geotiff with asp_map = true. 

## Types of changes

## Checklist:

- [ ] I have added myself to the [.zenodo.json](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/.zenodo.json)
  document.
- [x] I have added my user impacting change to the [CHANGELOG.md](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/CHANGELOG.md) document. (Added as `changes/6033.fix.md`.)

## Licensing

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the
  benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an
  overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

### DIFF
--- a/isis/src/base/objs/ProjectionFactory/ProjectionFactory.cpp
+++ b/isis/src/base/objs/ProjectionFactory/ProjectionFactory.cpp
@@ -10,6 +10,9 @@ find files of those names at the top level of this repository. **/
 #include <cfloat>
 #include <cmath>
 #include <iomanip>
+#include <sstream>
+
+#include <ogr_spatialref.h>
 
 #include "Camera.h"
 #include "Cube.h"
@@ -17,8 +20,10 @@ find files of those names at the top level of this repository. **/
 #include "Distance.h"
 #include "FileName.h"
 #include "IException.h"
+#include "IString.h"
 #include "Plugin.h"
 #include "Projection.h"
+#include "PvlGroup.h"
 #include "RingPlaneProjection.h"
 #include "TProjection.h"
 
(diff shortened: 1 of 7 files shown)


oleg-alexandrov:
Note that my PR_6039 and unrelated PR_6027 are both failing with identical errors:

undefined reference to `testing::internal::MakeAndRegisterTestInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char const*, char const*, char const*, testing::internal::CodeLocation, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'

something changed in the env. 

PR_6036 from 3 hours ago worked.

acpaquette:
Only thing I can find is a minizip version change:
Screenshot 2026-05-05 at 7 02 12 PM

isis/src/base/objs/ProjectionFactory/ProjectionFactory.cpp
```diff
@@ -1159,4 +1164,168 @@ namespace Isis {
     }
     return (Isis::Projection *) proj;
   }
+  std::string ProjectionFactory::PvlToWkt(const Isis::PvlGroup &mapGrp,
+                                          double defaultSemiMajor,
+                                          double defaultSemiMinor) {
+
+    std::string projName = mapGrp["ProjectionName"][0].toLower().toStdString();
```
acpaquette, line 1171: @oleg-alexandrov Thoughts on doing something like this rather than copy code from gdal?
```C++
    QString file = "temp.cub";
    Cube cube;
    cube.setDimensions(1, 1, 1);
    cube.create(file);
    PvlObject &isisCube = cube.label()->findObject("IsisCube");
    isisCube.addGroup(mapGrp);
    cube.close();

    GDALDataset *dataset = GDALDataset::FromHandle(GDALOpen(file.toStdString().c_str(), GA_ReadOnly));
    if (!dataset) {
      QString msg = "Failed opening GDALDataset from [" + file + "]";
    }

    const OGRSpatialReference *oSRS = dataset->GetSpatialRef();
    if (!oSRS) {
      QString msg = "Failed obtain spatial reference from [" + file + "]";
    }

    char *wkt = nullptr;
    oSRS->exportToWkt(&wkt);
    std::cout << "Banana " << wkt << std::endl;
    GDALClose(dataset);

    if (!wkt)
      throw IException(IException::Programmer,
        "Failed to export projection to WKT", _FILEINFO_);
    std::string result(wkt);
    CPLFree(wkt);
    return result;
```
I know its code that is unlikely to change but not having to maintain it in two places seems better

oleg-alexandrov: Adam, are you saying we should create on-the fly temporary cubes on disk to be able to parse a a pvl string? 

If that's the case, we likely better reach out to GDAL folks and request they expose a function in the ISIS driver we could use?

acpaquette: Or potentially allow for the [OGRSpatialReference](https://gdal.org/en/stable/api/ogrspatialref.html#_CPPv419OGRSpatialReference) to import a pvl mapping group

oleg-alexandrov: So, more work on the GDAL side. Presumably we can work with them, wait for them to release GDAL with it, pin to that, and wipe our own implementation in due time. Which would be a cycle of some length of time.

oleg-alexandrov:
Anything holding this up?

Kelvinrr:
Just confirming, this needs work on the GDAL side still?

acpaquette:
Ideally yes, I hesitate to add functions on the ISIS side that will eventually be deprecated but there has been zero movement on gdal for this. https://github.com/OSGeo/gdal/issues/14555

oleg-alexandrov:
The GDAL side is now in place. GDAL has a new reusable function, OGRSpatialReference::importFromISISPVL, that reads a projection from an ISIS cube PVL Mapping group. It was merged to GDAL master (OSGeo/gdal#14918) and ships in GDAL 3.14.

So the PvlToWkt copy here is temporary. Once ISIS raises its minimum GDAL to 3.14, this can call importFromISISPVL instead of carrying its own conversion. Until that release and pin update, keeping PvlToWkt avoids holding up the cam2map fix.

PR #6064, opened 2026-05-28
Qt6 warnings

oleg-alexandrov:
## Description

This fixes a merge conflict in https://github.com/DOI-USGS/ISIS3/pull/6005 and makes a small fix for Bullet pointer null check as a test was failing. Nearly all the rest is @acpaquette's changes as as they were there. Nice to fix all those warnings and problematic logic with array on stack.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/UserInterface/UserInterface.cpp
+++ b/isis/src/base/objs/UserInterface/UserInterface.cpp
@@ -610,7 +610,7 @@ namespace Isis {
             if (!matchesDefault) {
               PutAsString(keyword, values);
               commandline += keyword + "=";
-              for(QString val : values) {
+              for (const QString &val : values) {
                 commandline += val + " ";
               }
             }
--- a/isis/src/base/objs/UserInterface/UserInterface.cpp
+++ b/isis/src/base/objs/UserInterface/UserInterface.cpp
@@ -351,7 +351,7 @@ namespace Isis {
 
       QStringList tokens = t.split(",");
 
-      foreach (QString token, tokens) {
+      for(QString token : tokens) {
         // removes quotes from tokens. NOTE: also removes escaped quotes.
         token = token.remove( QRegularExpression("[\"']") );
         p_batchList[i].push_back(token);
@@ -610,7 +610,7 @@ namespace Isis {
             if (!matchesDefault) {
               PutAsString(keyword, values);
               commandline += keyword + "=";
-              foreach(QString val, values) {
+              for(QString val : values) {
                 commandline += val + " ";
               }
             }
--- a/isis/src/mgs/objs/MocLabels/MocLabels.cpp
+++ b/isis/src/mgs/objs/MocLabels/MocLabels.cpp
@@ -99,11 +99,6 @@ namespace Isis {
     AlphaCube a(cube);
     p_ns = a.AlphaSamples();
     p_nl = a.AlphaLines();
-
-    // // Get the two kernels for time computations
-    // PvlGroup &kerns = lab.findGroup("Kernels", Pvl::Traverse);
-    // p_lsk = FileName(kerns["LeapSecond"][0]);
-    // p_sclk = FileName(kerns["SpacecraftClock"][0]);
   }
 
   /**
--- a/isis/src/mgs/objs/MocLabels/MocLabels.cpp
+++ b/isis/src/mgs/objs/MocLabels/MocLabels.cpp
@@ -100,10 +100,10 @@ namespace Isis {
     p_ns = a.AlphaSamples();
     p_nl = a.AlphaLines();
 
-    // Get the two kernels for time computations
-    PvlGroup &kerns = lab.findGroup("Kernels", Pvl::Traverse);
-    p_lsk = FileName(kerns["LeapSecond"][0]);
-    p_sclk = FileName(kerns["SpacecraftClock"][0]);
+    // // Get the two kernels for time computations
+    // PvlGroup &kerns = lab.findGroup("Kernels", Pvl::Traverse);
+    // p_lsk = FileName(kerns["LeapSecond"][0]);
+    // p_sclk = FileName(kerns["SpacecraftClock"][0]);
   }
 
   /**
--- a/isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
+++ b/isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
@@ -1763,12 +1763,12 @@ namespace Isis {
       QString message = "You are saving changes to a measure on an ignored ";
       message += "point.  Do you want to set Ignore = False on the point and ";
       message += "both measures?";
-      int ret = QMessageBox::question(this, "Save Measure", message, 
+      int ret = QMessageBox::question(this, "Save Measure", message,
                                       QMessageBox::Yes | QMessageBox::No,
                                       QMessageBox::Yes);
       switch (ret) {
         // Yes:  set Ignore=false for the point and measures and save point
-        case 0:
+        case QMessageBox::Yes:
           m_editPoint->SetIgnored(false);
           emit ignorePointChanged();
           if (m_leftMeasure->IsIgnored()) {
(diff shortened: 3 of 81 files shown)


isis/src/qisis/objs/QnetTools/QnetNavTool.cpp
```diff
@@ -858,12 +858,14 @@ namespace Isis {
...
+    QString message = "You have chosen to set " +
+                      QString::number(selected.size()) +
+                      " point(s) to ignore. Do you want to continue?";
+    int ret = QMessageBox::question((QWidget *)parent(),
+                                    "Control Network Navigator - Ignore Points",
+                                    message,
+                                    QMessageBox::Yes | QMessageBox::No);
+    switch (ret) {
       case 0: // Yes was clicked or Enter was pressed, delete points
         QApplication::setOverrideCursor(Qt::WaitCursor);
```
Kelvinrr, line 870: The way these are handled are fairly inconsistent, some places it was updated to check against enum values and others they keep the numerical checks.

isis/src/mgs/objs/MocLabels/MocLabels.cpp
```diff
@@ -100,10 +100,10 @@ namespace Isis {
...
     p_nl = a.AlphaLines();
 
-    // Get the two kernels for time computations
-    PvlGroup &kerns = lab.findGroup("Kernels", Pvl::Traverse);
-    p_lsk = FileName(kerns["LeapSecond"][0]);
-    p_sclk = FileName(kerns["SpacecraftClock"][0]);
+    // // Get the two kernels for time computations
+    // PvlGroup &kerns = lab.findGroup("Kernels", Pvl::Traverse);
+    // p_lsk = FileName(kerns["LeapSecond"][0]);
+    // p_sclk = FileName(kerns["SpacecraftClock"][0]);
```
Kelvinrr: we can probably delete these

isis/src/base/objs/UserInterface/UserInterface.cpp
```diff
@@ -610,7 +610,7 @@ namespace Isis {
             if (!matchesDefault) {
               PutAsString(keyword, values);
               commandline += keyword + "=";
-              foreach(QString val, values) {
+              for(QString val : values) {
```
Kelvinrr: probably want this to be a `QString&` to avoid copies.

oleg-alexandrov:
Thanks for the review @Kelvinrr. These should all be addressed now: removed the commented-out kernel lines in MocLabels, switched the UserInterface loop to a const reference, and made the QnetNavTool dialog cases use the QMessageBox enum values consistently.

oleg-alexandrov:
I went through the compiler warnings from the build log and fixed the most obvious, low-risk ones in a separate commit. This pass deliberately only covers the clear-cut compiler warnings, not the larger or judgement-heavy categories.

Worth flagging one real bug it turned up: in mgs/apps/mocuncompress/image_io.cpp the snprintf size argument was sizeof((char *)hdr_buf + IMAGE_LABEL_OFFSET), which is just the size of a pointer (8 bytes), so the image label was being truncated to 7 characters. That is now the actual remaining buffer length, IMAGE_HEADER_LENGTH - IMAGE_LABEL_OFFSET.

The rest of the fixes are cosmetic and safe: removed a few unused variables, changed int loop counters to size_t to silence signedness comparisons, added a (size_t) cast in Spice.cpp, and switched a few catch(std::invalid_argument) blocks to catch by const reference.

Still outstanding, left for a follow-up since they need more thought:
- the nodiscard/unused-result warnings (mostly QFile::open and QTemporaryFile::open, ~40 of them) - each needs a decision on what to do when the call fails.
- the Qt6 deprecated-declarations (QScopedPointer to std::unique_ptr, the QDomDocument::setContent overload) - real API migrations.
- the free-nonheap-object warnings, which come from inside the deprecated QLinkedList in Qt5Compat.
- the cppcheck static-analysis items (dangerousTypeCast, uninitMemberVar, etc.), which are a separate bucket from the compiler warnings.

Happy to take those on in a later round if you want.

(Prepared with the help of Claude.)

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -1761,7 +1763,10 @@ namespace Isis {
       QString message = "You are saving changes to a measure on an ignored ";
       message += "point.  Do you want to set Ignore = False on the point and ";
       message += "both measures?";
-      switch (QMessageBox::question(this, "Save Measure", message, "&Yes", "&No", 0, 0)) {
+      int ret = QMessageBox::question(this, "Save Measure", message, 
+                                      QMessageBox::Yes | QMessageBox::No,
+                                      QMessageBox::Yes);
+      switch (ret) {
```
acpaquette, line 1769: ~~Another instances where we should change from numeric to enum.~~

I don't know if we want to change all of these to use the enum verses integer comparison

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -1893,7 +1898,9 @@ namespace Isis {
     if (origMeasure->IsIgnored() && m->IsIgnored()) {
       QString message = "The " + side + "measure is ignored.  ";
       message += "Do you want to set Ignore = False on the measure?";
-      switch (QMessageBox::question(this, "Save Measure", message, "&Yes", "&No", 0, 0)) {
+      int response = QMessageBox::question(this, "Save Measure",
+                                           message, QMessageBox::Yes | QMessageBox::No);
+      switch (response) {
```
acpaquette, line 1903: Another one

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
```diff
@@ -1918,8 +1925,10 @@ namespace Isis {
           message += "may need to move all of the other measures to match the new ";
           message += " coordinate of the reference measure.  Do you really want to ";
           message += " change the reference measure's location? ";
-          switch(QMessageBox::question(this, "Save Measure",
-                                       message, "&Yes", "&No", 0, 0)){
+          int ret = QMessageBox::question(this, "Save Measure",
+                                          message,
+                                          QMessageBox::Yes | QMessageBox::No);
+          switch(ret) {
```
acpaquette, line 1931: Another one

isis/src/qisis/objs/ControlPointEditWidget/ControlPointEditWidget.cpp
acpaquette: It seems like most use the integer comparison with some using the enum comparison

acpaquette:
Overall this PR seems fine. If @Kelvinrr approves it can get merged

oleg-alexandrov:
Thanks @acpaquette, good catch on the inconsistency. Looking closer, the numeric cases weren't just a style mismatch, they were an actual bug. The new StandardButton overload of QMessageBox::question returns the enum value (QMessageBox::Yes is 0x4000, QMessageBox::No is 0x10000), not the 0/1 button index the old string-button overload returned. So any switch still written as case 0 / case 1 matched nothing and the Yes/No handling silently did nothing. The worst instances were the "do you really want to change the reference measure?" dialogs in QnetTool.cpp and ControlPointEditWidget.cpp, where clicking No should revert and return false but instead fell through and saved anyway.

I made it uniform across the whole PR rather than just the file you commented on. Every case 0 became case QMessageBox::Yes and every case 1 became case QMessageBox::No, in ControlPointEditWidget.cpp (3 switches), MatchTool.cpp (4) and QnetTool.cpp (4). I also fixed GuiEditFile.cpp, where an if(ret) check on a Save|Cancel dialog was always true since both enum values are nonzero, so it saved even on Cancel; that now checks ret == QMessageBox::Save, matching closeFile() right below it. I confirmed every remaining case 0/case 1 in those files belonged to a converted QMessageBox switch (the unrelated switch(column) blocks use named enums), so the change is safe and predictable. Builds clean here; just pushed, so CI will pick it up.

(Prepared with the help of Claude.)

oleg-alexandrov:
The build passes.

Kelvinrr (approved):

PR #6067, opened 2026-05-28
Fix access to target body radii in Target/ShapeModel constructors

### LINKED ISSUE #6066: Target body radii unavailable in Target/ShapeModel constructors

**Description**  
The target body radii is not available when the Target object is created by the Spice class. Shape models are created in the Target class during construction which creates a problem for ShapeModels that need the radii to create reference ellipsoids at creation time. 

This problem occurs because the Spice class retrieves and stores the target body radii after the target body code is determined. The body code can be directly specified in the labels within the Kernels group, but this code is read from the labels by Target. Upon return from Target instantiation, the Spice class then populates the radii array and stores these data within the label's NaifKeywords object in the BODY_CODE and BODY{BODY_CODE}_RADII keywords. 

However, the body code is provided to the Spice object from the Target object after instantiation. Hence the circular dependency preventing the body radii being available to the ShapeModelFactory class that instantiates the appropriate ShapeModel. Recall ShapeModels take Target as a constructor argument but the Spice object has not initialized the target body radii at this point.

**How to reproduce**  
In ShapeModel objects, you can call target->radii() method in any method after it's created, but the array _m_radii_ is initialized to a Distance() defaults in the constructor, which are NULLs.

**Possible Solution**  
I have a PR coming that creates a new method, Spice::body_radii(), that will retrieve the body radii on behalf of the Target object at construction and make the data available to the ShapeModel at construction time.

KrisBecker:
## Description
The PR provides a new method called Spice::body_radii() to retrieve target body radii using the code provided as an argument. This does method does not record the radii to the labels, but tries first to retrieve them using NAIF. If that is unsuccessful, it attempts to read the radii from the NaifKeywords object. I these attempts fail, an exception is thrown.

## Related Issue
Fixes #6066.

Tests were added to the SpiceTests gtest to determine correct behavior of the method.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/Target/Target.cpp
+++ b/isis/src/base/objs/Target/Target.cpp
@@ -79,29 +79,28 @@ namespace Isis {
 
       *m_systemCode = (*m_bodyCode/100)*100 + 99;
 
-      SpiceChar naifBuf[40] = { "unknown" };
+      SpiceChar naifBuf[40];
       SpiceBoolean found;
       bodc2n_c((SpiceInt) *m_systemCode, sizeof(naifBuf), naifBuf, &found);
       NaifStatus::CheckErrors();
       string s(naifBuf);
       (*m_systemName).append(s.c_str());
 
-      // QString radiiKey = "BODY" + QString((BigInt) naifBodyCode()) + "_RADII";
-      // m_radii[0] = Distance(getDouble(radiiKey, 0), Distance::Kilometers);
-      // m_radii[1] = Distance(getDouble(radiiKey, 1), Distance::Kilometers);
-      // m_radii[2] = Distance(getDouble(radiiKey, 2), Distance::Kilometers);
+
     }
     // Override it if it exists in the labels
     if (kernels.hasKeyword("NaifBodyCode")) {
       *m_bodyCode = (int) kernels["NaifBodyCode"];
     }
-
-    // Gotta do this for shape models that need radii for reference ellipsoids.
-    // Note if the spice pointer is null (i.e., old unit tests), then don't 
-    // make the call.
+    
+    // Ensure the reference ellipsoid radii are present for ShapeModels
     if ( nullptr != spice ) {
-      if ( !m_sky ) m_radii = spice->body_radii( *m_bodyCode );    
+      QString radiiKey = "BODY" + toString( *m_bodyCode ) + "_RADII";
+      m_radii[0] = Distance( spice->getDouble(radiiKey, 0), Distance::Kilometers);
+      m_radii[1] = Distance( spice->getDouble(radiiKey, 1), Distance::Kilometers);
+      m_radii[2] = Distance( spice->getDouble(radiiKey, 2), Distance::Kilometers);      
     }
+
     m_shape = ShapeModelFactory::create(this, lab);
   }
 
--- a/isis/src/base/objs/Target/Target.cpp
+++ b/isis/src/base/objs/Target/Target.cpp
@@ -79,7 +79,7 @@ namespace Isis {
 
       *m_systemCode = (*m_bodyCode/100)*100 + 99;
 
-      SpiceChar naifBuf[40];
+      SpiceChar naifBuf[40] = { "unknown" };
       SpiceBoolean found;
       bodc2n_c((SpiceInt) *m_systemCode, sizeof(naifBuf), naifBuf, &found);
       NaifStatus::CheckErrors();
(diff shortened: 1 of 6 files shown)


isis/src/base/objs/Target/Target.cpp
acpaquette: Is there a reason that this section of code couldn't be uncommented rather than adding a new method to the Spice class? This seems to take care of the issue around uninitialized radii. The `getDouble` call would just need to be replaced with `readStoredValue` and a `toDouble()` conversion.

Alternatively, we could call `spice.getDouble()` and setting the radii would just have to check that `nullptr != spice`.

Something like:
```C++
if ( nullptr != spice ) {
  m_radii[0] = Distance(spice.getDouble(radiiKey, 0), Distance::Kilometers);
  m_radii[1] = Distance(spice.getDouble(radiiKey, 1), Distance::Kilometers);
  m_radii[2] = Distance(spice.getDouble(radiiKey, 2), Distance::Kilometers);
}
```

acpaquette (commented):
Thanks for getting this in Kris! Just one comment to address

isis/src/base/objs/Target/Target.cpp
KrisBecker: There are several reasons why I did not take that route.

The main problem with just calling Spice::readStoredValue() there is that it reads only reads from the NaifKeywords object in the label. When running spiceinit, those values are not present in the label until Spice calls getDouble() upon return from the Target call. However, Target, thus ShapeModel, needs them at spiceinit time so you must get the those values from the kernel pool.

I also did not try to divine why those calls were commented out, so there could be other reasons. This seemed a reasonable approach as it has a better chance of success and does not populate/pollute the label every time Spice::getDouble() is called.

acpaquette: I don't see a big difference between `fetch_radius` lamba and the spice `getDouble` function. The way the `Target` object is created in the Spice class is that either you have loaded kernels or you have a naif keywords group with populated values.

If you have loaded kernels and are already passing a Spice object to the Target for construction, you're not polluting the keywords, you're adding the key to the Spice object from Target there rather than in the Spice class on [lines 537-549](https://github.com/DOI-USGS/ISIS3/blob/38d410daef3517b7b46b78b79d2a9f2d15566843/isis/src/base/objs/Spice/Spice.cpp#L537). I guess I don't understand why we are avoiding using existing functionality and writting more code to maintain.

KrisBecker: The Spice::body_radii() can be called from any routine to get radii for any target (e.g., moons of a planet body rather than the planet) and not write them to the label.

If you prefer I just make the call to Spice::getDouble() instead, I will make the change.

acpaquette: Yes, I would prefer that. A good middle ground seems like keeping the `body_radii` function on Spice but make the internals use `toDouble` rather than `fetch_radius`

Kelvinrr: it's a philosophical question on whether you write this function that get reused or copy paste simple code, imho the problem with the former is we would have to track down everywhere getting radii to make it make sense (otherwise we are adding to the bloat). 

imho in this case, probably better off using the tools that exist than writing a new function.

Kelvinrr:
There seems to be some tests failures, but dev builds are also not clean rn (the one two days ago is, but the most recent is not), but it's not running. I suspect it might be an intermitten issue so after we get a dev build going we can see if the errors are in this PR or pre-existing.

KrisBecker:
This does appear to be caused by the new implementation in the Target class. The Target code cannot duplicate the behavior of the [Spice implementation](https://github.com/DOI-USGS/ISIS3/blob/38d410daef3517b7b46b78b79d2a9f2d15566843/isis/src/base/objs/Spice/Spice.cpp#L537-L550) because Spice first directly stores the body code in the NaifKeywords object before getting the body radii. The Spice:: storeValue() method is called to do this which is a protected method, and therefore cannot be called in Target.

This changes the order of the NaifKeyword object keywords and appears to be causing the test failures in the isis_unit_test_Spice. I suspect this is also the cause of the other test failures as well, but I cannot figure out how to confirm this. I don't know how to evaluate output results of the old unitTests in the current cmake environment.

acpaquette:
I can confirm that the `isis_unit_test_Spice` test is failing due to the keywords not being in the same order. I don't think order within a PVL group should be an issue. The other failures are due to not checking if the target is sky. We need to add a sky check along side the spice check:
```C++
if ( nullptr != spice && isSky != true) {
...
}
```

acpaquette:
I think we can update the spice unit test and this PR should be good to go

acpaquette:
@KrisBecker I can update the test truthdata if you would like

KrisBecker:
Yeah the would be a great help.

Thanks!

acpaquette:
@KrisBecker I couldn't push directly so there is a PR open on the U of A repo
https://github.com/UA-LPL/ISIS3/pull/1

KrisBecker:
Tests pass. Thanks for your help.

acpaquette (approved):

PR #6071, opened 2026-06-08
Fix variable line-rate camera ground-to-image round trip near a rate change

oleg-alexandrov:
## Description

The HRSC camera, and any camera that scans with a variable line rate (Dawn VIR, Rosetta VIRTIS, CRISM, Clipper push-broom), uses the VariableLineScanCameraDetectorMap class. Its SetDetector method, which maps an ephemeris time back to a parent line, selected the line-rate section by comparing the time against GetStartEt() - 0.5. That 0.5 is the half-pixel offset of the line convention used by SetParent and lineRate (line numbers are pixel-centered), but here it is subtracted from a time in seconds, so it really means half a second. For a time within half a second before a line-rate change, SetDetector selected the next section while SetParent had selected the previous one. When the line rate differs across that boundary, the image-to-ground-to-image round trip does not produce the orignal result.

The fix removes that incorrect 0.5 offset, comparing currEt directly to GetStartEt(), so SetDetector selects the same section as SetParent.

This also converts the old-style objs/HrscCamera unitTest into a gtest at isis/tests/HrscCameraTests.cpp, reusing the existing h2254_0000_s12.cub test data (no new test cube is added). The gtest keeps the original coverage (kernel IDs, instrument and spacecraft names, four-corner round trips, center ground point) and adds a regression test that round-trips a line just before the first line-rate change and asserts it arrives back at the starting pixel within a tolerance of 0.02 pixel; that assertion fails before the fix and passes after. The legacy isis/src/mex/objs/HrscCamera/unitTest.cpp and its HrscCamera.truth are removed as part of the conversion.

## Related Issue

No existing issue; the bug was found while validating the HRSC camera model. The changelog fragment is timestamped accordingly.

Using the publicly available HRSC product H2254_0000_S12 (dataset MEX-M-HRSC-3-RDR-V2.0, from the PDS Geosciences HRSC archive, which is also the cube used by the regression test):

```
hrsc2isis from=h2254_0000_s12.img to=h2254_0000_s12.cub
spiceinit from=h2254_0000_s12.cub ckpredicted=true
```

This cube's first line-rate change is at line 633. An ISIS image-to-ground-to-image round trip just before it, at sample 1300, line 585, with campt:

```
# image -> ground
campt from=h2254_0000_s12.cub sample=1300 line=585 to=fwd.pvl
# ground -> image (feed the resulting lat/lon back in)
campt from=h2254_0000_s12.cub type=ground latitude=<lat> longitude=<lon> to=rev.pvl
```

returns Line = 583.989 (off by 1.011 px) before this fix and Line = 584.99999963 (off by 4e-7 px) after it. Lines away from a rate change, and the rate-change line itself, are unaffected.

As an additional cross-check, a CSM camera was built and compared against the ISIS camera:

```
isd_generate -k h2254_0000_s12.cub h2254_0000_s12.cub        # writes h2254_0000_s12.json
cam_test --image h2254_0000_s12.cub --cam1 h2254_0000_s12.cub \
         --cam2 h2254_0000_s12.json --session1 isis --session2 csm --sample-rate 100
```

The maximum pixel difference, in the ground-to-image direction this bug affects, went from about 0.9 pixel before the fix to about 1e-4 pixel after it, consistent with the round trip above.

Effect on other tests: the variable-line-rate camera tests (HRSC, Dawn VIR, Rosetta VIRTIS, CRISM, Clipper push-broom) were run with the fix applied and with it reverted, and the results are identical except for the new HRSC regression test, which flips from failing to passing. No existing truth or gold files needed regenerating: the change only moves results that were already wrong (about a pixel off, near a rate change) onto the correct value, and no other test sampled that region. SetParent (image to ground) is untouched, so all absolute latitude/longitude values are unchanged.

This change was developed with assistance from Claude (Anthropic).

## Types of changes

- [x] Infrastructure change (changes to things like CI or the build system that do not impact users)

## Checklist:

## Licensing

This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain.


oleg-alexandrov:
Any thoughts on this? This has been carefully validated with the CSM implementation. The CSM implementation is generic, a single line scan camera for all sensors. The bug is in VariableLineScanCameraDetectorMap in ISIS and seems to be a mixup of pixels and seconds.

amystamile-usgs (approved):

PR #6023, opened 2026-04-23
Allow GDAL PDS format to be read

chkim-usgs:
## Description

Enables native support for virtual file systems and remote streaming (/vsicurl/) by using GDAL's VSI API. Local file usage performance should not be affected since VSI is basically a pass-through to standard C `fopen` calls. 

## Related Issue

Addresses https://github.com/DOI-USGS/ISIS3/issues/5864

## How Has This Been Validated?

Tested `/vsicurl` URLs:

**mroctx2isis**
```
$ISISROOT/bin/mroctx2isis from=/vsicurl/https://planetarydata.jpl.nasa.gov/img/data/mro/ctx/mrox_5011/data/V13_084840_0996_XN_80S159W.IMG to=./mroctx2.cub
```

**vims2isis**
```
$ISISROOT/bin/vims2isis from=/vsicurl/https://pds-imaging.jpl.nasa.gov/api/data/atlas:pds3:cas:cassini_orbiter:/covims_0094/data/2017255T000819_2017257T195837/v1884113035_1.qub vis=v1884113035_1.vis.cub ir=v1884113035_1.ir.cub
```

**mrf2isis**
```
$ISISROOT/bin/mrf2isis from=/vsicurl/https://pds-geosciences.wustl.edu/lro/lro-l-mrflro-4-cdr-v1/lromrf_0001/data/sar/00200_00299/level1/lsb_00222_1cd_xiu_79s202_v1.lbl to=lromrf.lbl.cub
```

**hi2isis**
```
$ISISROOT/bin/hi2isis from=/vsicurl/https://pds-imaging.jpl.nasa.gov/api/data/atlas:pds3:mro:mars_reconnaissance_orbiter:/HiRISE/EDR/ESP/ORB_075400_075499/ESP_075499_1945/ESP_075499_1945_BG12_0.IMG to=hirise.cub
```

**marci2isis**
```
$ISISROOT/bin/marci2isis from=/vsicurl/https://pds-imaging.jpl.nasa.gov/api/data/atlas:pds3:mro:mars_reconnaissance_orbiter:/mrom_0001/data/CRU_000001_9999_MA_99N999W.IMG to=marci.cub
```

**lo2isis**
```
$ISISROOT/bin/lo2isis from=/vsicurl/https://pds-imaging.jpl.nasa.gov/api/data/atlas:pds3:lo:lunar_orbiters:/LO_1001/DATA/LO1/FRAME_1005/FRAME_1005_H1.LBL to=lo.cub
```

[...]

PR #6079, opened 2026-06-22
tgocassis2isis: populate SpacecraftClockStartCount for PSA-exported CaSSIS products

oleg-alexandrov:
## Description

When `tgocassis2isis` ingests a TGO CaSSIS calibrated product that was exported to the ESA PSA (the `em16_tgo_cas` namespace labels downloaded from the archive or ODE), it selects the `TgoCassisExportedInstrument_PSA.trn` fallback translation. That table fills `StartTime` but has no `SpacecraftClockStartCount` group, so the cube comes out without that keyword.

`SpacecraftClockStartCount` is required on the cube regardless of camera model. The ISIS `TgoCassisCamera` constructor reads it, and `spiceinit` builds its cache through the camera, so both fail when it is missing. The instrument-team products map it from the FSW header via `TgoCassisInstrument.trn`, but the PSA export drops it, so until now the workaround was a manual `editlab` step to inject it.

This populates the keyword directly in the ingester. The PSA path already opens and parses the same XML to choose its translation; the clock value sits in that document as a hex-ASCII `em16_tgo_cas:exposuretimestamp` field. A few lines read it, hex-decode it, and set `SpacecraftClockStartCount` on the Instrument group. The translation table alone cannot do this because it cannot hex-decode, which is likely why it was left out. The block is guarded by `if (!inst.hasKeyword("SpacecraftClockStartCount"))`, so the instrument-team path, where the keyword is already present, is untouched.

The keyword value is not used for geometry: per the long-standing TODO in `TgoCassisCamera.cpp`, the camera computes time from `StartTime` in UTC. The keyword only has to be present for the camera to construct. Injecting the real decoded timestamp is the clean choice and stays correct if CaSSIS is ever switched to SCLK as that TODO intends.

Result: PSA-exported calibrated products ingest for both the ISIS camera and CSM with no manual step.

A changelog entry was added (`changes/6079.fix.md`).

## Related Issue

Follows up on the discussion in DOI-USGS/ale#720, where this `editlab` workaround came up and the in-ingester fix was proposed.

**New automated test.** The existing `TgoCassis2IsisTestPSALabel` gtest ingests a PSA-exported label but did not previously check the clock (the old PSA path never produced it). Its test label (`MY36_015782_024_0_PAN_cropped.xml`) had been cropped down and no longer contained the `exposuretimestamp` field, so a representative `em16_tgo_cas:exposuretimestamp` was added back to it, and an assertion was added that the ingested cube's `Instrument/SpacecraftClockStartCount` decodes to the expected value. This exercises the new decode path directly. The full `tgocassis2isis` suite passes (10/10), including this case.

**Real data, reproducible.** Fetch one publicly available PSA-exported calibrated framelet (anonymous, no login) and ingest it. Using framelet 0 of orbit 5684, acquisition 280077162:

```bash
PSA=https://archives.esac.esa.int/psa/ftp/ExoMars2016/em16_tgo_cas/data_calibrated/Science_Phase/Orbit_Range_5600_5699/Orbit_5684/Science/280077162/PAN
base=cas_cal_sc_20190303T070517-20190303T070521-5684-8-PAN-280077162-0-0__4_0
wget "$PSA/$base.xml" "$PSA/$base.dat"

tgocassis2isis from=$base.xml to=$base.cub
getkey from=$base.cub grpname=Instrument keyword=SpacecraftClockStartCount
# -> 2f0595a051db5fd9
```

That last `getkey` now returns the clock count. It is the hex-decode of the label's `exposuretimestamp` field, which can be checked independently:

```bash
grep -ioE '<em16_tgo_cas:exposuretimestamp>[0-9a-f]+' $base.xml | sed 's/.*>//' | xxd -r -p
# -> 2f0595a051db5fd9
```

Before this change, the produced cube had no `SpacecraftClockStartCount`, so `spiceinit` failed, and the keyword had to be injected by hand from that same decoded value:

```bash
clk=$(grep -ioE '<em16_tgo_cas:exposuretimestamp>[0-9a-f]+' $base.xml | sed 's/.*>//' | xxd -r -p)
editlab from=$base.cub options=addkey grpname=Instrument keyword=SpacecraftClockStartCount value=$clk
```

With this change that `editlab` step is no longer needed: `tgocassis2isis` sets the keyword during ingest, and `spiceinit from=$base.cub spkpredicted=true` then succeeds directly (kernels required for the `spiceinit` step only; the keyword population itself needs none). The instrument-team (non-PSA) products are unaffected, since the new block only runs when the keyword is absent.

## Types of changes

## Checklist:

## Licensing
- [x] I dedicate any and all copyright interest in this software to the public domain.

### DIFF
--- a/isis/src/tgo/apps/tgocassis2isis/tgocassis2isis.cpp
+++ b/isis/src/tgo/apps/tgocassis2isis/tgocassis2isis.cpp
@@ -10,7 +10,6 @@ find files of those names at the top level of this repository. **/
 
 #include <QString>
 #include <QtMath>
-#include <QFile>
 
 #include "AlphaCube.h"
 #include "Cube.h"
@@ -328,6 +327,34 @@ namespace Isis {
     if (inst.hasKeyword("ExposureDuration")){
       inst.findKeyword("ExposureDuration").setUnits("seconds");
     }
+
+    // The PSA-exported calibrated products (em16_tgo_cas namespace) do not carry
+    // SpacecraftClockStartCount. Their fallback translation table fills only
+    // StartTime. The instrument-team products handled by TgoCassisInstrument.trn
+    // map this keyword from the FSW header, so it is already present for them.
+    // The camera (TgoCassisCamera) and spiceinit require it regardless of the
+    // product, so without it the PSA products fail to ingest. The clock value is
+    // present in the XML as a hex-ASCII em16_tgo_cas:exposuretimestamp field.
+    // Decode it and add the keyword here, so the PSA products work with no manual
+    // editlab step. The translation table alone cannot do this because it cannot
+    // hex-decode. Read the label through OriginalXmlLabel, the same VSI-aware path
+    // the rest of this app uses, so remote (/vsicurl) inputs work too.
+    if (!inst.hasKeyword("SpacecraftClockStartCount")) {
+      // Parse without namespace processing so the prefixed
+      // em16_tgo_cas:exposuretimestamp tag is matched as a literal name.
+      OriginalXmlLabel expLabel;
+      expLabel.readFromXmlFile(inputLabel);
+      QDomDocument expDoc = expLabel.ReturnLabels();
+      QDomNodeList tsNodes = expDoc.elementsByTagName("em16_tgo_cas:exposuretimestamp");
+      if (!tsNodes.isEmpty()) {
+        QString hexAscii = tsNodes.at(0).toElement().text().trimmed();
+        QString clockCount = QString::fromLatin1(QByteArray::fromHex(hexAscii.toLatin1()));
+        if (!clockCount.isEmpty()) {
+          inst.addKeyword(PvlKeyword("SpacecraftClockStartCount", clockCount));
+        }
+      }
+    }
+
     // Translate BandBin group
     FileName bandBinTransFile(missionDir + "TgoCassisBandBin.trn");
     XmlToPvlTranslationManager bandBinXlater(inputLabel, bandBinTransFile.expanded());
--- a/changes/6079.fix.md
+++ b/changes/6079.fix.md
@@ -1,0 +1,1 @@
+Fixed `tgocassis2isis` to populate `SpacecraftClockStartCount` for PSA-exported CaSSIS calibrated products by decoding it from the label `exposuretimestamp` field, so these products ingest without a manual `editlab` step.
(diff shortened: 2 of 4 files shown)


isis/src/tgo/apps/tgocassis2isis/tgocassis2isis.cpp
```diff
@@ -343,6 +343,34 @@ namespace Isis {
...
+    // product, so without it the PSA products fail to ingest. The clock value is
+    // present in the XML as a hex-ASCII em16_tgo_cas:exposuretimestamp field.
+    // Decode it and add the keyword here, so the PSA products work with no manual
+    // editlab step. The translation table alone cannot do this because it cannot
+    // hex-decode.
+    if (!inst.hasKeyword("SpacecraftClockStartCount")) {
+      QFile expFile(inputLabel.expanded());
+      if (expFile.open(QIODevice::ReadOnly)) {
+        QDomDocument expDoc;
+        expDoc.setContent(&expFile);
```
amystamile-usgs: Are we assuming the xml is correct here?

amystamile-usgs (commented):
Overall looks good. Just one question.

isis/src/tgo/apps/tgocassis2isis/tgocassis2isis.cpp
Kelvinrr: we also want to make sure it supports VSI to match the PR that was just merged.

oleg-alexandrov: Both addressed. Regarding VSI, now the read is one through OriginalXmlLabel::readFromXmlFile, the same VSI-aware path #6023 switched the rest of this app to, so a remote /vsicurl input gets the clock keyword populated just like a local file.

 One detail: it reads without namespace processing, so the prefixed em16_tgo_cas:exposuretimestamp tag is matched as a literal name (parsing with namespace processing on made the prefixed lookup miss). The now-unused QFile include was removed.

To address Amy's point regarding what we assume about XML correctness. This logic does not add any new trust in the file beyond what the ingest already relies on. By the time it runs, tgocassis2isis has already opened and parsed this same XML earlier (to detect the PSA product and pick the fallback translation), so the document is known-readable here, and readFromXmlFile throws on a genuine parse error rather than proceeding on a bad file.

It is also guarded so a missing or malformed field is a no-op: it only runs when SpacecraftClockStartCount is absent, so instrument-team products that already carry it are untouched.

The value is also not used for geometry: per the long-standing TODO in TgoCassisCamera.cpp the camera computes time from StartTime in UTC, so the keyword only has to be present for the camera to construct, and an unexpected value cannot change results today.

The tgocassis2isis gtest suite passes locally, including the PSA case that asserts the decoded SpacecraftClockStartCount. I force-pushed to this PR to integrate with dev.

The bot did the work here and I inspected.

oleg-alexandrov:
Tests passs.

amystamile-usgs (approved):

Issue #5861, opened 2025-08-03
Modeling of lens distortion in ISIS


oleg-alexandrov:
**ISIS version(s) affected**: latest

**Description**  

A handful of cameras in ISIS have the same issue as what occurred in the CSM implementation and since fixed. They treat lens distortion around the origin as special, and assume pixels close enough to origin do not need the distortion operation.

There's no reason for that special treatment, as applying a formula such as:

x1 = x0 * (1 + c1*r^2 + c2*r^4) 

or so is a well-behaved numerical operation for small r, and same for the division by that. 

In fact, having those as special cases would result in issues with numerical differentiation needed for some camera operations. 

Problematic distortion and undistortion:

JunoDistortionMap

OsirisRexDistortionMap

OsirisRexOcamsDistortionMap

CameraDistortionMap

LoHighDistortionMap

Hyb2OncDistortionMap

MarciDistortionMap

MocWideAngleDistortionMap

The fix is to remove in both distortion and undistortion operation any check such as "if (r < 1e-6)".

By contrast, for the cameras below the distortion is implemented correctly. A formula is applied as-is, and the model is quite similar to the above.

DawnFcDistortionMap 

RadialDistortionMap

NewHorizonsLorriDistortionMap

ApolloMetricDistortionMap

NewHorizonsMvicTdiCameraDistortionMap

NewHorizonsMvicFrameCameraDistortionMap

KaguyaMiCameraDistortionMap

KaguyaTcCameraDistortionMap

ClementineUvvisDistortionMap

Chandrayaan1M3DistortionMap

LroWideAngleCameraDistortionMap

LroNarrowAngleDistortionMap

ReseauDistortionMap

RosettaOsirisCameraDistortionMap

OsirisRexOcamsOpenCVDistortionMap

OsirisRexTagcamsDistortionMap

ThemisIrDistortionMap

TaylorCameraDistortionMap

acpaquette:
@oleg-alexandrov Any idea how fixing this will change the ground intersection point or the resulting projection? If you are working with Juno, a small detailed example of what one would expect to change when this is addressed would be nice to have

oleg-alexandrov:
The practical effect of this fix would likely be negligible. The "ground intersection" algorithm relies on the secant method with big steps, which are on the order of 0.1 m or more, and such algorithms also stop early, with a height tolerance of 0.01 meters or so.

[...]

PR #6102, opened 2026-08-01
Extrapolate the reseau distortion model smoothly outside the image area

oleg-alexandrov:
## Description

The reseau distortion model (*ReseauDistortionMap*, used by the Viking, Voyager, Mariner 10, and Apollo Metric cameras) fits a local inverse-distance-weighted affine to the five nearest reseaus. Both directions (the forward and inverse distortion maps, *SetFocalPlane* and *SetUndistortedFocalPlane*) reject (returning false) any point whose fitted position fell outside the image box. So a ground point that projects a few pixels off the sensor has no image position.

The result is that  bundle adjustment and map projection fail at the image boundary, and the camera projection is discontinuous there. This is very problematic for numerical differentiation and inconsistent with polynomial lens distortion models that continue nicely for a while outside the precise image box.

The proposed fix anchors the local affine fit at a query point clamped to the image box, then evaluates that affine at the true input point, and drops the out-of-bounds rejection.

- Inside the image the clamp is the identity, so the same neighbors are chosen and the same affine is fit and evaluated. Results are unchanged (identical output).
- Outside the image the query slides onto the nearest box edge, where the five nearest reseaus are always a two dimensional neighborhood (never colinear), so the affine is well conditioned. Evaluating it at the true point extrapolates the distortion linearly: continuous across the boundary and bounded (it grows only linearly, it does not diverge).

Cross section of the displacement magnitude along the detector mid line. The current curve lies on top of the new curve inside the detector and stops at the edges; the new curve continues smoothly and grows only linearly outside.
reseau_deformation_magnitude

Displacement magnitude across the plane, current model (defined only near the detector) versus new model (one continuous field past the black 1204x1056 box).
reseau_deformation_profile

The nonlinear part of the distortion (global affine removed, arrows scaled up). The interior is identical in both panels; the new model extends the field smoothly outside.
reseau_deformation_vectors

This PR was created with Claude (AI assistance) and reviewed and approved by me.

## Related Issue

Continues the theme of #5861 (closed), making the ISIS distortion models well-behaved analytic operations instead of relying on special cases near the edges. That issue did not touch the out-of-box rejection. There is no open issue specifically for reseau extrapolation outside the image box.

*ReseauDistortionMap* has only a stub unit test and is exercised through the camera unit tests.

- Tests that pass unchanged: the Viking camera unit test passes, and its interior lines (the four corner round trips and the center latitude/longitude) are byte identical to the previous truth. Running the test with this PR reverted produces the same interior output, confirming inside the image is unchanged. Apollo Metric passes. Mariner 10 and Voyager reproduce their existing truths on every line that executed (their only failures on my machine are missing local kernel/reseau data, not this PR), which confirms the shared code preserves interior behavior for all four cameras.

- Test modified: *isis/src/viking/objs/VikingCamera/unitTest.cpp*. I also added *GDALAllRegister()* at the top of *main*, matching three other object unit tests. Without it the unmodified test segfaults on the current dev under the GDAL backed Cube IO (a driver is not registered in the raw test), independent of this PR.

- New test for the outside logic: a small section round trips two pixels outside the 1204x1056 detector (sample 1300 line 1200, and sample -100 line -100). With the previous code *SetImage* returns false and these print ERROR; with this PR they project and round trip within a pixel and print OK. Only OK/ERROR is printed so the result is stable across platforms. The truth was regenerated for the added lines only.

- Figure validation: the deformation figures were generated by a transcription of the algorithm checked against a grid dumped from the real ISIS C++ model; they agree to 0.037 pixel over 13,431 inside and outside points.

## Types of changes

## Checklist:

## Licensing

This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain.

### DIFF
--- a/isis/src/base/objs/ReseauDistortionMap/ReseauDistortionMap.cpp
+++ b/isis/src/base/objs/ReseauDistortionMap/ReseauDistortionMap.cpp
@@ -88,14 +88,31 @@ namespace Isis {
     double focalLine = dy / p_pixelPitch +
                        p_camera->FocalPlaneMap()->DetectorLineOrigin();
 
-    // Find distance from input point to all nominal reseaus
+    // The local affine fit is anchored at a query point clamped to the distorted
+    // image box. Inside the image the query equals the input point, so the result
+    // is identical to before. Outside the image the query slides onto the nearest
+    // box edge, so the five closest reseaus always form a two dimensional
+    // neighborhood (never colinear) and yield a well conditioned affine. That
+    // affine is then evaluated at the true input point below, which extrapolates
+    // the distortion smoothly and linearly past the image edges. This lets ground
+    // points that project just off the sensor still return a usable focal plane
+    // position, as needed by bundle adjustment and map projection, instead of a
+    // hard failure at the image boundary.
+    double querySamp = focalSamp;
+    double queryLine = focalLine;
+    if(querySamp < 0.5) querySamp = 0.5;
+    if(querySamp > p_distortedSamps + 0.5) querySamp = p_distortedSamps + 0.5;
+    if(queryLine < 0.5) queryLine = 0.5;
+    if(queryLine > p_distortedLines + 0.5) queryLine = p_distortedLines + 0.5;
+
+    // Find distance from the query point to all nominal reseaus
     std::vector<double> distances(p_numRes, 0.0);
     double wt[5];
     int closepts[5];
     double ldiffsq, sdiffsq;
     for(int i = 0; i < p_numRes; i++) {
-      sdiffsq = (focalSamp - p_rsamps[i]) * (focalSamp - p_rsamps[i]);
-      ldiffsq = (focalLine - p_rlines[i]) * (focalLine - p_rlines[i]);
+      sdiffsq = (querySamp - p_rsamps[i]) * (querySamp - p_rsamps[i]);
+      ldiffsq = (queryLine - p_rlines[i]) * (queryLine - p_rlines[i]);
       distances[i]  =  ldiffsq + sdiffsq;
     }
 
@@ -154,16 +171,15 @@ namespace Isis {
       lsqX.Solve();
       lsqY.Solve();
 
+      // Evaluate the fitted affine at the true input point. Inside the image this
+      // is the usual interpolation. Outside, since the query was clamped to the
+      // box edge while the evaluation point is the true (off image) location, the
+      // affine is extrapolated smoothly rather than rejected.
       known[1] = focalSamp;
       known[2] = focalLine;
 
-      // Test to make sure the point is inside of the image
       double undistortedFocalSamp = lsqX.Evaluate(known);
       double undistortedFocalLine = lsqY.Evaluate(known);
-      if(undistortedFocalSamp < 0.5) return false;
-      if(undistortedFocalLine < 0.5) return false;
-      if(undistortedFocalSamp > p_undistortedSamps + 0.5) return false;
-      if(undistortedFocalLine > p_undistortedLines + 0.5) return false;
 
       // Convert undistorted sample, line position to an x,y position
       p_undistortedFocalPlaneX = (undistortedFocalSamp - p_undistortedSamps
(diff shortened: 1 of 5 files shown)


oleg-alexandrov:
All test pass apart from 	

1486 - DefaultCube.SpiceinitWebTrue
1516 - DemCube.FunctionalTestSpiceinitWebAndShapeModel 

which fail on dev too due to unrelated issue: An error occurred when talking to the server. The server encountered an unexpected error.

isis/src/base/objs/ReseauDistortionMap/ReseauDistortionMap.cpp
```diff
@@ -88,14 +88,31 @@ namespace Isis {
...
+    // The local affine fit is anchored at a query point clamped to the distorted
+    // image box. Inside the image the query equals the input point, so the result
+    // is identical to before. Outside the image the query slides onto the nearest
+    // box edge, so the five closest reseaus always form a two dimensional
+    // neighborhood (never colinear) and yield a well conditioned affine. That
+    // affine is then evaluated at the true input point below, which extrapolates
+    // the distortion smoothly and linearly past the image edges. This lets ground
+    // points that project just off the sensor still return a usable focal plane
+    // position, as needed by bundle adjustment and map projection, instead of a
+    // hard failure at the image boundary.
```
Kelvinrr: Claude likes to make very verbose comments, I recommend tuning it so it cuts these down.

isis/tests/FunctionalTestsCampt.cpp
```diff
@@ -251,8 +251,9 @@ TEST_F(DefaultCube, FunctionalTestCamptCoordList) {
   groundPoint = appLog.group(2);
   EXPECT_DOUBLE_EQ( (double) groundPoint.findKeyword("Sample"), 100.0);
   EXPECT_DOUBLE_EQ( (double) groundPoint.findKeyword("Line"), 10000.0);
-  QString ErrorMsg = "Requested position does not project in camera model; no surface intersection";
-  EXPECT_PRED_FORMAT2(AssertQStringsEqual, groundPoint.findKeyword("Error"), ErrorMsg);
+  // Line 10000 is far outside the image; the reseau distortion now
+  // extrapolates outside the image domain, so the point projects (intended).
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, groundPoint.findKeyword("Error"), "NULL");
```
Kelvinrr, line 256: why would we not want it to error with the detailed error that it's off the image? If the goal is handling the boundaries of the image during bundle and projection, should we localize to those cases?

isis/tests/FunctionalTestsCampt.cpp
```diff
@@ -274,8 +275,9 @@ TEST_F(DefaultCube, FunctionalTestCamptAllowError) {
 
   campt(testCube, options, &appLog);
   PvlGroup groundPoint = appLog.findGroup("GroundPoint");
-  QString ErrorMsg = "Requested position does not project in camera model; no surface intersection";
-  EXPECT_PRED_FORMAT2(AssertQStringsEqual, groundPoint.findKeyword("Error"), ErrorMsg);
+  // Sample/line -100 are outside the image; the reseau distortion now
+  // extrapolates outside the image domain, so the point projects (intended).
+  EXPECT_PRED_FORMAT2(AssertQStringsEqual, groundPoint.findKeyword("Error"), "NULL");
```
Kelvinrr: This test should either change to create an actual error with some other examples or have it still error when completely off the image as the point of this test is to test allow errors..

isis/src/base/objs/ReseauDistortionMap/ReseauDistortionMap.cpp
oleg-alexandrov: Agreed. Trimmed the four comment blocks down to a couple of lines each in the latest commit.

isis/tests/FunctionalTestsCampt.cpp
oleg-alexandrov: By design the distortion map is a pure coordinate transform, so it should stay continuous and analytic everywhere. Rejecting a point inside the transform makes it discontinuous and non-differentiable at the image edge, which is exactly what breaks numerical differentiation in bundle adjustment and iterative projection. Extrapolating a few pixels past the edge is the natural, expected behavior, and it matches the polynomial lens models.

The same holds for linescan sensors even without any distortion: a ground point outside the field of view, including earlier or later in time than the recorded lines, still maps to a well defined position on the sensor plane. Keeping that convention everywhere is preferable to treating the edges as a special case.

So rather than localize the extrapolation to bundle and projection, the cleaner split is to keep the model continuous and let the caller decide bounds, which is where it already lives. campt has allowoutside and allowerror, and CameraPointInfo checks HasSurfaceIntersection and InCube. allowoutside defaults to true for campt, which is why this point now projects; leaving allowoutside off makes InCube report an out-of-cube error instead.

isis/tests/FunctionalTestsCampt.cpp
oleg-alexandrov: Agreed the allowerror path needs a real error, not an off-image point. I split this test in the latest commit. The off-image case (sample and line just outside) now asserts that the point projects, which is the intended new behavior. A separate allowerror test uses a pixel whose look ray misses the body, so a genuine no-surface-intersection error still exercises allowerror and confirms campt reports it in the group instead of throwing.

isis/tests/FunctionalTestsCampt.cpp
Kelvinrr: similar to comments, if claude is being used for responses to comments, keep them brief. In general, though, it's best to not use Claude for comments and docs since it can worsen hallucinations when they are read by other LLMs down the road, unlike code which can be easily tested empiracally to work, written technical writing cannot.

oleg-alexandrov: Yeah, but here verbosity is justified and I did review. This is an important point. Proper extrapolation is big deal, and the middle paragraph is what I suggested. In code, sure, we want it brief.

oleg-alexandrov: Claude's technical writing is quite accurate. I do always review. Its biggest problem is not accuracy, but it gets long-winded and makes unnecessary connections and digressions.

Kelvinrr: yeah, the long windedness is partly what can create issues since it can increase the chances of drifting, for example. It's why we dont use LLMs for our docs only for generating code and summeries like llm.txt/llms-full.txt. If you have to review it, might be worth writing it yourself. Either way, I think this PR looks good now.

Kelvinrr (approved):

Issue #5501, opened 2024-05-17
Support multi-band "from" images in map2cam 


rvwagner:
**Description**  
When using map2cam, currently if you specify a multiband image as the "from" image, only band 1 gets reprojected to match the "match" image.  I have a couple of projects where it would be useful if all of the "from" image bands were reprojected, creating a multi-band output image (currently I'm looping over all bands of the "from" image, running map2cam for each, and then cubeiting them together again, but making sure I keep the band ordering the same is a hassle).  I have the same need to get proper areal statistics in camera space as in #3856.

A potential complication I could see is that if the from and match images both have the same >1 number of bands, the user expectation might be that map2cam would pair them up, projecting from+1 to match+1, from+2 to match+2, etc.  Currently, it appears that if you provide two multi-band images, map2cam only projects from.cub+1 to match match.cub+1, and all other bands in both files are ignored, so any multiband support would be brand new behavior with no existing usage to support.

Kelvinrr:
Gonna bump this, I think in the past it hasn't been worked on because it has a workaround but seems like a good feature to have.

rvwagner:
Bump!

At least as an initial version, I'd suggest requiring that the "match" cube be single-band (error out otherwise, which is effectively an API change so this would have to go in a major-version update), and explicitly project all bands of the "from" cube to match that band in the match cube.  I can see enough confusion arising from any default multiband-to-multiband behavior that it's probably not worth trying to handle it ("I map2cam'd a WAC multi-band mosaic into a WAC image, why is only one band lining up?" "When I map2cam an image with four backplanes onto a WAC image, the bands get misaligned, but it's fine with 3 or 5 backplanes- what gives?"). Especially for a program as niche as this.

rvwagner:
Bump.

Kelvinrr:
@chkim-usgs can we make a note for this one for next support sprint, it's been open forvever. Should be a relatively simple change.

Kelvinrr:
@rvwagner just want to clarify the desired behavior. What we are planning to adding to map2cam: 

1. If Match nbands == From nbands, map2cam will process all bands by pairing fband1 and mband1, fband2 and mband2, etc. 
2. If bands are not equal, throws an exception   
3. To prevent this from being API breaking, we will put it behind a new flag (something like `bool pairbands`). 

Does this sounds reasonable? 

[...]

PR #6105, opened 2026-08-05
Update map2cam with multiband support

chkim-usgs:
## Description

Adds multi-band support for the FROM image in `map2cam`. Previously, if a multi-band map-projected image was passed as FROM, only band 1 was reprojected and the remaining bands were silently dropped. Now every band of the FROM cube is reprojected into the MATCH geometry, producing an output with the same number of bands as FROM.

~~Marked fragment file as **breaking** as the single-band MATCH requirement is an API breaking change.~~
^The API breaking change will be in a separate PR.

## Related Issue

Addresses https://github.com/DOI-USGS/ISIS3/issues/5501

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/tests/Map2camTests.cpp
+++ b/isis/tests/Map2camTests.cpp
@@ -5,7 +5,6 @@
 
 #include "Cube.h"
 #include "CubeAttribute.h"
-#include "IException.h"
 #include "PixelType.h"
 #include "Pvl.h"
 #include "PvlGroup.h"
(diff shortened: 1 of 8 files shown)


isis/tests/Map2camTests.cpp
```diff
@@ -29,4 +30,24 @@ TEST_F(DefaultCube, FunctionalTestMap2camTest) {
 
   ASSERT_TRUE(ocube.label()->findObject("IsisCube").hasGroup("Kernels"));
   ASSERT_FALSE(ocube.label()->findObject("IsisCube").hasGroup("Mapping"));
+
+  // test multi-band
+  ASSERT_GT(projTestCube->bandCount(), 1);
+  EXPECT_EQ(ocube.bandCount(), projTestCube->bandCount());
+}
```
Kelvinrr, line 36: We should add a test that has multiple bands as output.

isis/tests/Map2camTests.cpp
```diff
@@ -29,4 +30,24 @@ TEST_F(DefaultCube, FunctionalTestMap2camTest) {
...
+
+TEST_F(DefaultCube, FunctionalTestMap2camMultiBandMatchError) {
+  // a multi-band MATCH cube is not supported and should produce a user error
```
Kelvinrr: Why must it error? the previous behavior was to only use the first band correct? Why not keep that intact? Otherwise we cannot merge this until 11.0

isis/tests/Map2camTests.cpp
chkim-usgs: This `ASSERT_GT(projTestCube->bandCount(), 1);` checks that the FROM cube has a band count greater than 1 and the line after expects the output cube band count to match the FROM cube count. Should we add another test that checks multiple bands?

isis/tests/Map2camTests.cpp
chkim-usgs: I was basing these changes off of the [most recent comment](https://github.com/DOI-USGS/ISIS3/issues/5501#issuecomment-3483052994) on the issue by the OP. Did we actually want to allow multi-band MATCH cubes?

isis/tests/Map2camTests.cpp
Kelvinrr: oof, I just read ASSERT, didn't notice the _GT

isis/tests/Map2camTests.cpp
Kelvinrr: We could leave the default behavior and move this in another PR that can stick until we do an 11.0

Kelvinrr (approved):

Issue #5786, opened 2025-04-30
Allow use of high-vertical-resolution DTMs as shapemodels


rvwagner:
Resurrecting #4604 (I would have done it directly, but apparently once an issue is auto-closed as "completed" the original poster doesn't have permission to re-open it) because it's still impossible to use a high-resolution DTMs that were produced as elevation data as a radius source without either quantizing them badly, doing an ugly conversion mess to take advantage of the scale and offset metadata available to 16-bit data, or both if the data has a large range.

This is a particular issue with shape-from-shading DTMs, because the ~25 cm vertical quantization of 32-bit floating-point data at the lunar radius generally completely eliminates the improved detail that SfS adds over photogrammetric DTMs. (We've been dealing with this issue recently in our testing of the Ames Stereo Pipeline, which works in 64-bit floating-point to avoid exactly this problem.)

If you want to have a DTM that you use for evaluating small-scale features and precise slopes and also for projecting images on (both common use-cases when analyzing landing sites for missions), you currently need to maintain two versions of the file, one in elevation form for direct analysis and one in radius form for map-projection (this is what I've ended up doing to support recent lunar CLPS missions).

Based on the old issue's discussion, I see a few potential solutions for this, none of them perfect:

1) Add base/multiplier support to 32-bit floating-point data (and make sure that internally the representation is 64-bit floating-point).  To avoid having the nobody-will-ever-figure-this-out issue of `editlab grpname=Pixels keyword=Base value=1737400` being the right way to convert from elevation to radius, give `demprep` a way to add a fixed offset to elevation-type pixel values, something like `demprep SPHERICALDATUMRADIUS=1737400`, that will add that value to the cube's Base keyword.  If someone has a more complex datum, have the demprep documentation describe the correct suffixes to add to the `to=` of an fx call to preserve details appropriately.

2) Add support for 64-bit floating-point files, albeit not as the default format. (With the same demprep addition, but instead of modifying Base, have it convert to 64-bit.)

3) Add some sort of metadata for "add this offset to the recorded data when using this file as a radius source" (e.g. for the Moon, it would be a tag to add 1,737,400 to all pixels).  This would be nice for direct analysis of the DTM, as it would leave the underlying pixel values as easy-to-read-and-compare-to-existing-data elevation values. However, as previously described, this option gets really ugly for planets where elevation-type terrain models are typically defined relative to a more complex shape than a sphere, and apparently involves touching a lot of disparate bits of code).

[...]

PR #6112, opened 2026-08-07
Add ability to Read and Write Cubes with 64 Bit Floats

acpaquette:
## Description
This PR adds the ability to read and write cubes with 64-bit floats (doubles)

## Related Issue

#5786 

New tests added to directly test all pixel type read writes on the cube IO handler.

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/base/objs/Cube/Cube.cpp
+++ b/isis/src/base/objs/Cube/Cube.cpp
@@ -313,7 +313,7 @@ namespace Isis {
 
       if (newFileAttributes.propagateMinimumMaximum()) {
         if(result->pixelType() > pixelType()) {
-          // If we are an integer type, directly apply the base and multiplier
+          // If we are translating between integer types, directly apply the base and multiplier
           if (isIntegerType(result->pixelType())) {
             result->setBaseMultiplier(base(), multiplier());
           }
--- a/isis/src/base/objs/Cube/Cube.cpp
+++ b/isis/src/base/objs/Cube/Cube.cpp
@@ -1423,6 +1423,12 @@ namespace Isis {
       m_multiplier = (max - min) / (x2 - x1);
       m_base = min - m_multiplier * x1;
     }
+    else if (m_pixelType == Real) {
+      x1 = VALID_MIN4;
+      x2 = VALID_MAX4;
+      m_multiplier = (max - min) / (x2 - x1);
+      m_base = min - m_multiplier * x1;
+    }
   }
 
 
--- a/isis/src/base/objs/Cube/Cube.cpp
+++ b/isis/src/base/objs/Cube/Cube.cpp
@@ -312,10 +312,7 @@ namespace Isis {
       }
 
       if (newFileAttributes.propagateMinimumMaximum()) {
-        if(result->pixelType() == Isis::Real) {
-          result->setBaseMultiplier(0.0, 1.0);
-        }
-        else if(result->pixelType() >= pixelType()) {
+        if(result->pixelType() >= pixelType()) {
           result->setBaseMultiplier(base(), multiplier());
         }
         else {
--- a/isis/src/base/objs/Cube/Cube.cpp
+++ b/isis/src/base/objs/Cube/Cube.cpp
@@ -312,19 +312,15 @@ namespace Isis {
       }
 
       if (newFileAttributes.propagateMinimumMaximum()) {
-        if(result->pixelType() > pixelType()) {
-          // If we are translating between integer types, directly apply the base and multiplier
-          if (isIntegerType(result->pixelType())) {
-            result->setBaseMultiplier(base(), multiplier());
-          }
-          // Going from integer to float32/64 or float32 to float64, make the assumption that
-          // float32/64 can handle the data with the default base and multiplier
-          else {
+        // For new cubes created as float32 or float64, make the assumption that
+        // float32/64 can handle the data with the default base and multiplier
+        // Even if the data being copied has a base and multiplier (legacy behavior)
+        if (!isIntegerType(result->pixelType())) {
             result->setBaseMultiplier(0.0, 1.0);
-          }
         }
-        else if (result->pixelType() == pixelType()) {
-            result->setBaseMultiplier(base(), multiplier());
+        // If we are translating between integer types, directly apply the base and multiplier
+        else if(result->pixelType() >= pixelType()) {
+          result->setBaseMultiplier(base(), multiplier());
         }
         else {
           QString msg =
--- a/isis/src/base/objs/Cube/Cube.cpp
+++ b/isis/src/base/objs/Cube/Cube.cpp
@@ -312,8 +312,19 @@ namespace Isis {
       }
 
       if (newFileAttributes.propagateMinimumMaximum()) {
-        if(result->pixelType() >= pixelType()) {
-          result->setBaseMultiplier(base(), multiplier());
+        if(result->pixelType() > pixelType()) {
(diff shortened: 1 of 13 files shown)


isis/src/base/objs/Cube/CubeIoHandler.cpp
```diff
@@ -1361,13 +1361,21 @@ namespace Isis {
...
             double &bufferVal = buffersDoubleBuf[bufferIndex];
 
-            if(m_pixelType == Real) {
+            if (m_pixelType == Double) {
+              double raw = ((double *)chunkBuf)[chunkIndex];
+              if(m_byteSwapper)
+                raw = m_byteSwapper->Double(&raw);
+
+              bufferVal = (double)raw;
+            }
```
Kelvinrr, line 1372: Does this not need something like ` ((double *)buffersRawBuf)[bufferIndex] = raw;`? Other branches here end with something similar.

isis/src/base/objs/Cube/CubeIoHandler.cpp
```diff
@@ -1361,13 +1361,21 @@ namespace Isis {
...
+
+            else if(m_pixelType == Real) {
               float raw = ((float *)chunkBuf)[chunkIndex];
               if(m_byteSwapper)
                 raw = m_byteSwapper->Float(&raw);
 
               if(raw >= VALID_MIN4) {
-                bufferVal = (double) raw;
+                bufferVal = ((double)raw) * m_multiplier + m_base;
```
Kelvinrr: does the double branch need anything with base and mult?

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -1423,6 +1420,12 @@ namespace Isis {
       m_multiplier = (max - min) / (x2 - x1);
       m_base = min - m_multiplier * x1;
     }
+    else if (m_pixelType == Real) {
+      x1 = VALID_MIN4;
+      x2 = VALID_MAX4;
+      m_multiplier = (max - min) / (x2 - x1);
+      m_base = min - m_multiplier * x1;
+    }
```
Kelvinrr, line 1440: why is this added? Do we need to scale Reals here? I imagine ISIS wasn't doing this before for a reason.

isis/src/base/objs/Cube/CubeIoHandler.cpp
acpaquette: It likely should just for consistency but since we aren't translating values into special pixels I don't think it's necessary

isis/src/base/objs/Cube/CubeIoHandler.cpp
acpaquette: Nope, multiplying and adding doubles that get stored in a double can lead to weird behavior. We would need a larger default storage type if we are going to multiply doubles together

isis/src/base/objs/Cube/Cube.cpp
acpaquette: Reals were the largest and often default storage type. If we want to add doubles as a storage type and you have a double cube and convert it to a real, you would want to apply a base and multiplier to attempt to keep the values similar to the double values.

Kelvinrr: wait so are Double cubes being downcast to 32-bit floats in memory when read?

acpaquette: No, if the user chooses to downcast from a double to a float it is necessary/a good idea to apply a base and multiplier to the output cube to attempt to keep the values in the new float cube as close to the original double cube. Something like:
`cubeatt from= my_double_cube.cub to= my_float_cube.cub+Real+-very_large_number:very_large_number`

Kelvinrr:
Just needs test failures addressed

isis/src/base/objs/Process/Process.cpp
```diff
@@ -355,20 +355,23 @@ Isis::Cube *Process::SetOutputCubeStretch(const QString &parameter, const int ns
       }
 
       if(att.propagateMinimumMaximum()) {
-        if(cube->pixelType() == Isis::Real) {
+        if(InputCubes.size() == 0) {
```
Kelvinrr, line 358: do we need to still check if inputcubes.size() == 0?

isis/src/base/objs/Cube/Cube.cpp
```diff
@@ -763,7 +767,8 @@ namespace Isis {
     setLabelsAttached(att.labelAttachment());
     if (!att.propagatePixelType())
       setPixelType(att.pixelType());
-    setMinMax(att.minimum(), att.maximum());
+    if (!att.propagateMinimumMaximum())
+      setMinMax(att.minimum(), att.maximum());
```
Kelvinrr, line 771: was this a unrelated bug fix?

isis/tests/FunctionalTestsDsk2isis.cpp
```diff
@@ -73,24 +73,15 @@ TEST_F(Dsk2isisDefault, FunctionalTestDsk2isisGridMethod) {
...
-
+  QVector<QString> args = {"from="+dskFile, "map=" + mapFile, "to="+outputCube, "method=GRID"};
+
+  UserInterface options(APP_XML, args);
+  try {
+    dsk2isis(options);
+  }
+  catch (IException &e) {
+    FAIL() << "Unable to convert dsk to dem: " <<e.toString().toStdString().c_str() << std::endl;
+  }
```
Kelvinrr: was NAIF global state issues not a factor here?

isis/src/base/objs/Process/Process.cpp
acpaquette: Yes, for cases where we only set an output cube. Most ISIS apps/processes that operate on only an output cube defaulted to `Real` so the check for a `Real` short circuited the check for no input cubes

isis/tests/FunctionalTestsDsk2isis.cpp
acpaquette: I didn't see an issue. I also did some digging and the `NaifDskPlateModel` doesn't use the kernel pool but DSA/DSK utility functions for file handling/opening. It seems like this avoids the kernel pool and allows for read access across applications/processes. I can make this a separate PR. The reason for the change was trying to debug the dsk2isis tests in a debugger is a pain when the app gets run in a process that doesn't inherit the debugging information, so the app runs without hitting any break points

isis/src/base/objs/Cube/Cube.cpp
acpaquette: Nope, this was largely being avoided because of how the `Cube` class handled `Real` bit types in `setMinMax` where a base and multiplier would not be computed if the bit type was a `Real`. Now that we are doing something with `Real`s in that function, we should only call `setMinMax` when we actually have something to propagate

chkim-usgs (approved):
Looks good. With this PR, since users can create a Double cube, will we add support for export apps? I think ProcessExport ([example](https://github.com/DOI-USGS/ISIS3/blob/dev/isis/src/base/objs/ProcessExport/ProcessExport.cpp#L1047)) and ProcessExportPds ([example](https://github.com/DOI-USGS/ISIS3/blob/dev/isis/src/base/objs/ProcessExportPds/ProcessExportPds.cpp#L267)) do not really handle 64-bit yet, perhaps other processes as well?

Kelvinrr:
@acpaquette A documentation update for this functionality might be useful in asc-astro-docs. At least updating existing docs on pixel types.

Issue #5721, opened 2025-02-03
Jigsaw - error propagation not completing


lwellerastro:
**ISIS version(s) affected**: 8.3.0

**Description**  
Error propagation has been running for nearly 100 days on my Kaguya TC global control network and I have no idea when it will finish. This network bundles without err prop in under 50 hours. I don't know where it is in the process because I have disabled the slurm standard out since a previous run that was approaching 30+ days was generating many hundreds of gigs in size (I needed to kill that run due to cluster and scratch maintenance). My current job appears to be running according to top (not dead or sleeping). This run time seems excessively long but I don't know how long it should take.

This is not necessarily a bug, but I do not want to post to the discussion board where this problem will receive zero attention.

**How to reproduce**  
All of my data are on /scratch due to its size. The run is occurring under my scratch area Kaguya_TC/Global/Morning/GlobalNetwork/ErrProp/. Here are the specs for the network followed by my jigsaw command to see what is being solved for:

**#** of images: Kaguya_TC/Global/Morning/GlobalNetwork/ErrProp
**#** of points: 11573190
**#** of measures: 45876606

```
jigsaw froml=KTC_Morning_Global_Combined_ImageValDNAdd_GCP_Fix2.lis cnet=KTC_Morning_Global_Combined_ImageValDNAdd_GCP_Fix2.net \
        onet=JigOut_ErrProp_KTC_Morning_Global_Combined_ImageValDNAdd_GCP_Fix2.net \
        radius=yes update=no errorprop=yes \
        sigma0=1.0e-5 maxits=10 \
        camsolve=accelerations twist=yes overexisting=yes \
        spsolve=position overhermite=yes \
        camera_angles_sigma=0.25 \
        camera_angular_velocity_sigma=0.1 \
        camera_angular_acceleration_sigma=0.01 \
        spacecraft_position_sigma=1000 \
        point_radius_sigma=100 \
        file_prefix=ErrProp_RadAccelTwist_SpkPos_ValDNAddGCP_Fix2
```
Yes, it's enormous, but how long should this take? Is there a way to calculate the amount of time necessary? Other than disabling the standard output as I have done, is it possible (via another post) to disable all of the text going to screen which is resulting in a massive log file? It's useful knowing something is happening during the error prop process, but can users get that information in with less verbosity?

I would like like my bundle to finish but it has been hogging a resource for months now and it's not clear to me there isn't any end in sight.
Any suggestions on how guesstimate how long this needs are appreciated.

[...]

PR #6117, opened 2026-08-14
Jigsaw optimizations

Kelvinrr:
## Description

Several fixes to Jigsaw that should cut down jigsaw runtimes up to 10x.

Results from Lynn's Kaguya network, baseline is 10.0.0, "now" is this PR. 

  | baseline | now
-- | -- | --
bundle solve (Total Elapsed Time) | 1197.72 s | 80.4–83.2 s
error propagation | 1053.97 s | 12.6–12.7 s
whole jigsaw process, wall | ~1300 s | 127–132 s
peak RSS | 3.2 GB | 3.2 GB

Summary of Fixes: 

* Error propagation: It used to invert the whole reduced normal matrix then scan all points once per block column looking for the few that image touches. Now only uses what's needed
* One geometry sweep per iteration instead of two.
* CHOLMOD analyzed once instead of reconputing every iteration 
* Redundant calls for camera geometry per iteration. 
* Label and control net reading. jigsaw no longer opens all 3361 cubes to build an image list only the GUI reads.
* Redundant reads, copies and parsing in PVL reader
* Sparse matrix iteration. numberOfRows(), it now returns the cached last block size. 

## Related Issue

#5721 

Local benchmarks give the same result for Lynn's Kaguya network compared to 10.0.0 jigsaw. Sigmas and residuals are within floating point rounding precesion. 

## Types of changes

## Checklist:

## Licensing


isis/src/base/objs/Spice/Spice.cpp
```diff
@@ -1138,7 +1138,7 @@ namespace Isis {
     m_uB[1] = uB[1];
     m_uB[2] = uB[2];
 
-    computeSolarLongitude(*m_et);
```
acpaquette, line 1141: Should probably just remove the comment here. It explains why the line was removed

isis/src/base/objs/SpicePosition/SpicePosition.cpp
```diff
@@ -1353,24 +1368,14 @@ namespace Isis {
...
   void SpicePosition::SetEphemerisTimePolyFunction() {
-    // Create the empty functions
-    Isis::PolynomialUnivariate functionX(p_degree);
-    Isis::PolynomialUnivariate functionY(p_degree);
-    Isis::PolynomialUnivariate functionZ(p_degree);
-
-    // Load the coefficients to define the functions
-    functionX.SetCoefficients(p_coefficients[0]);
-    functionY.SetCoefficients(p_coefficients[1]);
-    functionZ.SetCoefficients(p_coefficients[2]);
```
acpaquette, line 1364: Removing `PolynomialUnivariate` and replacing it with a function for both SpicePosition and SpiceRotation seems like a bit much. Could we instead look at ways to optimize `PolynomialUnivariate` directly?

isis/src/base/objs/SpiceRotation/SpiceRotation.h
```diff
@@ -477,6 +477,9 @@ namespace Isis {
       std::vector<double> p_av;           //!< Angular velocity for rotation at time p_et
       bool p_hasAngularVelocity;          /**< Flag indicating whether the rotation
                                                includes angular velocity*/
+      std::vector<double> m_cachedCJ;     //!< p_CJ the cached Euler angles were decomposed from
+      std::vector<double> m_cachedAngles; //!< Euler angles cached by Angles()
+      std::vector<int> m_cachedAxes;      //!< Axis order the cached Euler angles use
```
acpaquette, line 482: These values should likely have defaults set in the SpiceRotation constructors

isis/src/base/objs/SpiceRotation/SpiceRotation.cpp
```diff
@@ -1304,15 +1319,27 @@ namespace Isis {
...
-    angles.push_back(ang2);
-    angles.push_back(ang3);
+    std::vector<double> angles(3);
+    angles[0] = ang1;
+    angles[1] = ang2;
+    angles[2] = ang3;
 
     NaifStatus::CheckErrors();
+
+    m_cachedCJ = p_CJ;
```
acpaquette, line 1330: I don't think we need to cache `p_CJ`. It gets changed whenever `SetEphemerisTime` is called, so it may be better to invalidate the values for `m_cachedAngles` and `m_cachedAxes` in `SetEphemerisTime`. `SetAngles` and `SetAxes` also might need to be tweeked to accomadate `m_cachedAngles` and `m_cachedAxes`

isis/src/control/apps/jigsaw/jigsaw.xml
```diff
@@ -989,6 +994,24 @@
...
+      <description>
+        By default error propagation computes only the entries of the inverse of the
+        reduced normal equations that the parameter uncertainties actually require,
+        which is a small fraction of the whole matrix on a large network. Select this
+        option to invert the entire matrix instead. The results are the same to
+        round-off, so this is a fallback for comparison and is much slower.
+      </description>
+      <type>boolean</type>
+      <default>
+        <item>No</item>
```
acpaquette: This should likely be defaulted to `Yes` to retain backwards compatibility

isis/src/control/objs/BundleSettings/BundleSettings.cpp
```diff
@@ -54,6 +54,7 @@ namespace Isis {
     m_updateCubeLabel      = false;
     m_errorPropagation     = false;
     m_createInverseMatrix  = false;
+    m_denseInverse         = false;
```
acpaquette: Similar to the XML change, this should likely default to `true`

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -952,9 +961,31 @@ namespace Isis {
       double vtpv = 0.0;
       double previousSigma0 = 0.0;
 
+      // the symbolic factorization is retained across iterations, so discard any factor
+      // left over from a previous solve before a new pattern is analyzed
+      cholmod_l_free_factor(&m_L, &m_cholmodCommon);
```
acpaquette: It's unclear why this is being called outside of the loop, there has been no "previous solve"

isis/src/control/objs/ControlPoint/ControlPoint.cpp
```diff
@@ -1280,6 +1284,12 @@ namespace Isis {
   }
 
 
+  // avoids the SurfacePoint copy, which heap allocates coordinates and covariance
+  const SurfacePoint &ControlPoint::AdjustedSurfacePointRef() const {
```
acpaquette: Rather than add a method, we should probably just update `GetAdjustedSurfacePoint` to return a reference. The bigger question will be if we can make it a `const` reference without creating problems

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
```diff
@@ -952,9 +961,31 @@ namespace Isis {
...
 
+      // the residual sweep and the next iteration's partials sweep evaluate the same geometry at
+      // the same parameters, so fuse them unless something changes the normals in between
+      bool fuseResiduals = !m_bundleSettings->outlierRejection()
+                           && m_bundleResults.numberMaximumLikelihoodModels() == 0
+                           && m_bundleLidarControlPoints.isEmpty()
+                           && !hasRejectedMeasures();
+
+      if (fuseResiduals) {
+        m_focalPlaneComputedStored = true;
```
acpaquette, line 970: I'm not sure how much the `fuseResiduals` and `m_focalPlaneComputedStored` get us in terms of speed. It seems like moving `ComputeResiduals_Millimeters` out of the measures loop so we aren't doing m^2 camera operations seems like the bigger win. It would be nice to see what the speed penalty is like without this particular optimization

isis/src/core/src/Pvl.cpp
```diff
@@ -743,13 +743,13 @@ namespace Isis {
           }
         }
 
-        if(readKeyword == PvlKeyword("Group")) {
+        if(readKeyword.isNamed("Group")) {
```
acpaquette, line 746: We should likely update the equality operator rather than replace all of the equalities with a different function. Looking over`PvlKeyword` we can likely change the body of `operater==` from:

```C++
if(!m_name && !key.m_name) return true;
if(!m_name || !key.m_name) return false;

return (stringEqual(m_name, key.m_name));
```
to:
```C++
return (stringEqual(name(), key.name()));
```

isis/src/core/src/PvlKeyword.cpp
```diff
@@ -688,20 +741,35 @@ namespace Isis {
    */
   bool PvlKeyword::stringEqual(const QString &QString1,
                                const QString &QString2) {
-    Isis::IString s1(QString1);
```
acpaquette, line 691: Having to maintain two seperate string equality functions in ISIS (IString and now here) seems a bit excessive for efficiency. Is there someway we can port this into IString as static methods?

isis/CMakeLists.txt
```diff
@@ -207,8 +207,14 @@ if(NOT CMAKE_BUILD_TYPE)
...
 
+# if(APPLE)
+#   set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -O0" CACHE STRING "" FORCE)
+# else()
+#   set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Og"  CACHE STRING "" FORCE)
+# endif()
+
 if(APPLE)
-  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -O0" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_FLAGS_DEBUG "-O3 -DNDEBUG" CACHE STRING "" FORCE)
```
acpaquette: This seems like it should be reverted? Looks like Claude was a bit too trigger happy and change the flags in the CMakefile rather than update the cmake command used to configure the build

acpaquette (commented):
Initial review! Overall looks good, lots to chew on

isis/CMakeLists.txt
Kelvinrr: Nah this was me, I was manually removing flags to get it to run in release mode. FLAGS_DEBUG kept adding the debug flags so I manually remoed them.

isis/src/base/objs/SpicePosition/SpicePosition.cpp
Kelvinrr: I dont think you can really optimize it. This is technically the optimization since it gives the exact same result.

isis/src/base/objs/SpiceRotation/SpiceRotation.cpp
Kelvinrr: a more complicated invalidation scheme would complicate things so the chances of bugs go up. Everywhere thay changes p_CJ should cause this to be recomputed without causing issues.

isis/src/base/objs/SpicePosition/SpicePosition.cpp
acpaquette: This is the current `Expand` function for `PolynomialUnivariate`:
```C++
  void PolynomialUnivariate::Expand(const std::vector<double> &vars) {
    p_terms.clear();
    p_terms.push_back(1.0);

    for(int i = 1; i <= p_degree; i++) {
      p_terms.push_back(p_terms[i-1]*vars[0]);
    }
  }
```
This could be changed to avoid doing push_backs on an empty array. Similarly in the `BasisFunction` the `Evaluate` function looks like:
```C++
  /**
   * Compute the equation using the input variable.
   *
   * @param var A single double value to use for the equation.
   *
   * @return The output double value resulting from the equation.
   */
  double BasisFunction::Evaluate(const double &var) {
    
    std::vector<double> vars;
    vars.push_back(var);
    return BasisFunction::Evaluate(vars);
  }
```
And could be changed to actually allocate the vector without the push_back.

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
Kelvinrr: I can remove it and test seperately

isis/src/core/src/Pvl.cpp
Kelvinrr: maybe? `isnamed` seems more intuitive

isis/src/base/objs/SpicePosition/SpicePosition.cpp
Kelvinrr: it's not really an issue with simply appending to an array in a silly way, it's that every iteration was doing expensive heap allocations that were thrown away and reallocated. Even preallocating p_terms removes maybe two of the five-plus mallocs while leaving the QString name re-encoding, the coefficient copy, and the vars temporary untouched. This change gives the exact same result without any of that.

Kelvinrr:
tests pass

isis/src/base/objs/SpicePosition/SpicePosition.cpp
acpaquette: Okay, maybe a different approach then. Could we add a new `Evalute` function to the `PolynomialUnivariate` that basically copies what is being done in `evaluatePolynomial`? Something like:
```C++
  double PolynomialUnivariate::Evaluate(const std::vector<double> &coeffs, double var) {

    if ( (int)coeffs.size() != p_numCoefs ) {
      QString msg = "Unable to evaluate function for the given vector of values. "
                    "The number of terms in the expansion [" 
                    + toString( (int)coeffs.size() ) + "] does not match number of coefficients "
                    "in the basis equation [" + toString(p_numCoefs) + "]";
    }

    double result = 0.0;
    double term = 1.0;

    for (int i = 0; i <= degree; i++) {
      result += coefficients[i] * term;
      term *= var;
    }
    
    return result;
  }
```
Ideally we don't want to maintain the same function in two places

amystamile-usgs:
Needs a changelog

amystamile-usgs (approved):

isis/src/control/objs/BundleAdjust/BundleAdjust.cpp
acpaquette: @Kelvinrr It may have been nice to get an answer to this before merging

Issue #5947, opened 2026-01-26
Warning for TIFFs not made with the updated ISIS metadata (in GDAL or ISIS)


thareUSGS:
**Description**  
With ISIS's update to support GeoTIFF natively, there will likely be an extended period of time where users will try to use TIFF/GeoTIFFs that were not created with the latest needed version. It would be nice if a warning or tip is issued to make sure GDAL binaries are upgraded to 3.12.2+ (or whatever the final minor release is).

It is noted that, map projected GeoTIFFs should work in ISIS v10 correctly. This is more to address TIFF/GeoTIFFs created outside of ISIS v10 or older versions of GDAL which don't have the new "json:ISIS3 metadata". This was added to support applications like cam2map which require SPICE and camera model metadata.

Overview: Newer versions of GDAL and ISIS will embed ISIS3 metadata inside a GeoTiff.  This ultimately allows ISIS v10+ to support the TIFF format natively (including TIFF enhancements like compression, cloud streaming, sparse NoDATA blocks, and better interoperability). Updates are briefly described is this [ISIS RFC 13](https://github.com/DOI-USGS/ISIS3/discussions/5698) or in [GDAL v3.12.0 release notes](https://github.com/OSGeo/gdal/blob/v3.12.0/NEWS.md#gdal-3120---overview-of-changes) under GeoTIFF and ISIS3 format sections.

**Example Warning**  
This could be positive warning or tip. Maybe something like.

"This GeoTIFF has not been initialized with version GDAL version 3.12.2+ or ISIS v10+. It is recommended to upgrade GDAL and/or ISIS and recreate this file to allow for native TIFF/GeoTIFF support within ISIS v10+ applications."

Kelvinrr:
@chkim-usgs add this to the support board.

chkim-usgs:
@acpaquette made [some observations](https://github.com/DOI-USGS/ISIS3/pull/6121#pullrequestreview-4975424240) that ISIS v10 reads older GeoTIFFs fine and that the main problem was really the metadata missing the blob `_data`. Instead of a broad warning, a change was made to throw an error in `Blob::ReadGdal` if the blob data is absent.

PR #6121, opened 2026-08-19
Added readGdal() warning message for non-ISIS3 metadata

chkim-usgs:
## Description

Added this warning message in `readGdal()` that appears if the incoming file's `GDAL` metadata domain does not include either `json:ISIS3` or `json:PDS`: 
```
This file <filename> has not been initialized with GDAL version 3.12.2+ or ISIS v10+. It is recommended to upgrade your software and recreate this file to allow for native TIFF/GeoTIFF support in ISIS v10+.
```

## Related Issue

Addresses https://github.com/DOI-USGS/ISIS3/issues/5947

Added gtest and ran `catlab` on an image with no ISIS3/PDS metadata to see the exception warning message. 

## Types of changes

## Checklist:

## Licensing

### DIFF
--- a/isis/src/core/src/Pvl.cpp
+++ b/isis/src/core/src/Pvl.cpp
@@ -183,13 +183,6 @@ namespace Isis {
     else if (CSLFindString(metadataDomains.List(), domainPds) != -1) {
       metadata = CPLStringList(dataset->GetMetadata(domainPds), false);
     }
-    else {
-      QString msg = "This file [" + dataFilename.name() + "] has not been initialized "
-                    "with GDAL version 3.12.2+ or ISIS v10+. It is recommended to upgrade "
-                    "your software and recreate this file to allow for native TIFF/GeoTIFF "
-                    "support in ISIS v10+.";
-      IException(IException::User, msg, _FILEINFO_).print();
-    }
 
     if (metadata.Count() > 0 && metadata[0] != nullptr) {
       const char *metadataJsonString = metadata[0];
--- a/isis/src/core/src/Pvl.cpp
+++ b/isis/src/core/src/Pvl.cpp
@@ -183,6 +183,13 @@ namespace Isis {
     else if (CSLFindString(metadataDomains.List(), domainPds) != -1) {
       metadata = CPLStringList(dataset->GetMetadata(domainPds), false);
     }
+    else {
+      QString msg = "This file [" + dataFilename.name() + "] has not been initialized "
+                    "with GDAL version 3.12.2+ or ISIS v10+. It is recommended to upgrade "
+                    "your software and recreate this file to allow for native TIFF/GeoTIFF "
+                    "support in ISIS v10+.";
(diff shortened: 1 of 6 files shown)


isis/src/core/src/Pvl.cpp
```diff
@@ -183,6 +183,13 @@ namespace Isis {
     else if (CSLFindString(metadataDomains.List(), domainPds) != -1) {
       metadata = CPLStringList(dataset->GetMetadata(domainPds), false);
     }
+    else {
+      QString msg = "This file [" + dataFilename.name() + "] has not been initialized "
+                    "with GDAL version 3.12.2+ or ISIS v10+. It is recommended to upgrade "
+                    "your software and recreate this file to allow for native TIFF/GeoTIFF "
+                    "support in ISIS v10+.";
+      IException(IException::User, msg, _FILEINFO_).print();
+    }
```
acpaquette: I think we still want to be able to process Gtiffs that don't have an existing ISIS3 metadata area. This prevents users from bringing in GeoTiffs that have been processed elsewhere.

acpaquette (commented):
The original intent behind Trents issue makes sense in theory, but in practice I don't think we actually need it. I experimented a bit with GeoTiffs created using GDAL 3.11 by translating a cube into a GeoTiff and running various ISIS apps on the image.

ISIS 10/dev can read the labels and image just fine and you can run ISIS apps on the old geotiffs (GDAL 3.11 and lower) without issue. Where problems tend to arise is with old GeoTiff metadata that contain tables/blobs since they do not contain the actual data for the tables. See [here](https://github.com/DOI-USGS/ISIS3/blob/418b227f01f0af84fbc42f88df0bfe43c373f75f/isis/src/core/src/Blob.cpp#L292) for blob data extraction from the GDAL metadata.

In general, we should keep GeoTiff reading fairly frictionless. For this issue I think adding an exception in the `ReadGdal` function for blobs would provide enough information to users. Basically check if there is the `_data` key within the json metadata, if not throw an exception:
```C++
if (!jsonblob[key].contains("_data")) {
  QString msg = "Gdal ISIS3 metadata does not contain blob data for [" + QString::fromStdString(key) + "].\n"
                "The metadata was likely created in a version of GDAL prior to 3.12.2.\n"
                "The image will likely need additional ISIS processing to attach the same blob with data.";
  throw IException(IException::Io, msg, _FILEINFO_);
}
```

isis/src/core/src/Pvl.cpp
chkim-usgs: Since the exception is only printed and not thrown, the process should continue with the warning message appearing towards the beginning of the process. But I can remove this warning message and add the `_data` check mentioned above instead. ~~This check will probably go after the `const char *metadataJsonString = metadata[0];` line in the metadata validation if block and will have to verify the metadata domain is `json:ISIS3`.~~  Misread feedback comment, will add the check to the Blob.cpp `ReadGdal()` function.

acpaquette (approved):

Issue #4240, opened 2021-01-06
Duplicate serial numbers for some data sets


jlaura:
**ISIS version(s) affected**: <=4.3.0

**Description**  
This a reporting of an issue original reported here: https://github.com/USGS-Astrogeology/PDS-Pipelines/issues/512#issuecomment-754951546

The issue is that two different Clementine (in this case) images produce the same ISIS serial number when serial numbers should be unique. 

Here is a pair of Clementine HiRes EDRs that yield the same serial number:

https://pdsimage.wr.usgs.gov/Missions/Clementine/cl_0077/lun328/lhxxxxxx/lhxxxxxj/lha1365j.328
https://pdsimage.wr.usgs.gov/Missions/Clementine/cl_0077/lun328/lhxxxxxx/lhxxxxxj/lha1366j.328

They both result in serial numbers of "Clementine1/HIRES/1994-04-29T08:48:21.170"

**How to reproduce**  
```
(isis4.3.0) -bash-4.2$ clem2isis from=lha1365j.328 to=lha1365j.cub
(isis4.3.0) -bash-4.2$ clem2isis from=lha1366j.328 to=lha1366j.cub
(isis4.3.0) -bash-4.2$ getsn from=lha1365j.cub
Clementine1/HIRES/1994-04-29T08:48:21.170
(isis4.3.0) -bash-4.2$ getsn from=lha1366j.cub
Clementine1/HIRES/1994-04-29T08:48:21.170
```
**Possible Solution**  
Make a modification to the clementine entry in the translation table to add another unique key to the serial generation process.

**Additional context**  
The original issue suggest that other missions exhibit the same behavior. @dpmayerUSGS can you please provide other examples? 

Also, I did not see this being reported previously, but I may have missed it. Searching for 'duplicate' generated, as expected, a lot of results.

scsides:
@dpmayerUSGS, Proposed solution needs to be examined further. I could be missing something, but the labels of both images indicate they are from the HiRes camera. Unless there is an operating mode I'm not aware of, it is impossible for one instrument (as defined by ISIS) to take two images at the same time. That being said, looking at the labels suggests they are the same image, one is encoded as jpeg the other is encoded as N/A (possible PDS BSQ or something like that). If these are the same image, then by definition the SNs should match.

dpmayerUSGS:
[...]

PR #6134, opened 2026-08-29
Keep the instrument serial number for CSM cubes so jigsaw can read them

### LINKED ISSUE #5482: Serial Numbers Differ when csminit and spiceinit are used
**ISIS version(s) affected**: 8.0.0, isislts

**Description**  

`getsn` returns different serial numbers, depending on where spiceinit or csminit have been run.

**How to reproduce**  

Run `getsn` on both of these labels (changed `lbl` to `txt` to allow upload):

[TC1W2B0_01_05202S188E3481.txt](https://github.com/DOI-USGS/ISIS3/files/15081243/TC1W2B0_01_05202S188E3481.txt)
[TC1W2B0_01_05199S201E3512.txt](https://github.com/DOI-USGS/ISIS3/files/15081245/TC1W2B0_01_05199S201E3512.txt)

**Possible Solution**  

**Additional context**

oleg-alexandrov:
## Description

jigsaw reads its input cubes into a SerialNumberList, which requires a unique serial number per cube. A cube that has a CSM sensor model attached (a CsmInfo group) had its serial number built only from the CSMPlatformID, CSMInstrumentId, and ReferenceTime fields. That serial number drops the per-image identity, so two images that share a reference time, as sub-images of a framing instrument do, receive the same serial number, and jigsaw fails to read them with a duplicate serial number error.

This change makes the serial number code prefer the instrument-based serial number whenever the cube has a recognized instrument label (which csminit keeps), and use the CSM serial number only when there is no usable instrument label. This restores a unique serial number per image and makes it agree with the serial number the same cube had before csminit. A cube with no recognized instrument label still gets the CSM serial number, so that behavior is unchanged.

This came up while attaching CSM sensor models to JunoCam cubes with csminit. A JunoCam image is split into many framelets, and framelets from the same image share a reference time, so under the CSM serial number they all received the same serial number and jigsaw could not read them into a SerialNumberList.

## Related Issue

Fixes #5482 and #4240.

Added a unit test that reproduces the failure at the point jigsaw fails: two distinct cubes are given the same CsmInfo group and added to a SerialNumberList. Before this change the second cube gets the same CSM serial number and the add fails with a duplicate serial number error; after this change each cube keeps its unique instrument-based serial number and both are added. The test fails before the change and passes after it. The change was built and tested against the current dev branch.

To address the concern raised by amystamile-usgs and jlaura in #5482, that removing the CSM serial number breaks the CSM jigsaw tests: that happened with a full revert, which makes every CSM cube fall back to the instrument serial, including the test cubes that have no recognized instrument. This change is narrower, switching to the instrument serial only when the cube has a recognized instrument, so the CSM serial number is unchanged for cubes without one. All 94 CSM tests, the CSM jigsaw functional test, the existing serial number tests, and the new test pass.

## Types of changes

## Checklist:

## Licensing

Disclaimer: this work was done with AI assistance (Claude). The author was very heavily involved and edited and offered suggestions as need be.


oleg-alexandrov:
This branch fails with unrelated errors that also show up in the nightly dev failures. So something else changed even though there have been no actual recent commits in dev.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
```diff
@@ -97,25 +97,48 @@ namespace Isis {
...
+
+        static QString instTransFile = "$ISISROOT/appdata/translations/Instruments.trn";
+        static PvlToPvlTranslationManager instrumentXlater(instTransFile);
+        instrumentXlater.SetLabel(label);
+        instrument = instrumentXlater.Translate("InstrumentName");
+      }
+      catch (IException &e) {
+        mission = "";
+        instrument = "";
+      }
```
Kelvinrr, line 128: isn't this going to cause issues downstream?

isis/src/base/objs/SerialNumber/SerialNumber.cpp
```diff
@@ -97,25 +97,48 @@ namespace Isis {
...
+                            && instrument != "Unknown";
+
+    // Prefer the instrument-based serial number when the cube has a recognized ISIS
+    // instrument: it matches what spiceinit produces for the same cube, and unlike
+    // the CSM serial number (built only from CSMPlatformID, CSMInstrumentId, and
+    // ReferenceTime, which can be identical for multiple framelets of a framing
+    // instrument) it is unique per image. Use the CSM serial number only when there
+    // is no usable instrument serial (for example a generic CSM cube with no ISIS
+    // instrument label).
+    if(label.findObject("IsisCube").hasGroup("CsmInfo") && !instrumentUsable) {
```
Kelvinrr, line 143: I dont recall the original conversations about these CSM serial numbers. I think is was a Jesse thing a long time ago. I think the correct thing to do is remove the use of this weird serial numbers in it's entirety and only use the traditional ones ISIS always used. This could be considered API breaking (this would force anyone using the next version of ISIS to re-generate serials) but if it's so broken that it prevents CSM from being useful in bundle, I am fine waving it as more of a bug fix. If it's a once in a while conflict. We would need to do this in a non-API breaking way.

Kelvinrr (commented):
comments in line.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
oleg-alexandrov: Yeah, here likely it is better to just throw if we can't find a good serial number.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
oleg-alexandrov: Having CSM produce a serial number alone is a good thing. Jigsaw now accepts csm cameras as in input, and should be able to function if the input images are not even cub files but plain tif. 

Claude found the true source of problem being a trunctated time and a fixed is proposed in https://github.com/DOI-USGS/usgscsm/pull/534.

I would think prioritizing the cub serial number is a good thing, with fallback to manufactured camera-only serial number.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
oleg-alexandrov: There is a new commit that throws rather than quietly returning empty strings.

oleg-alexandrov:
All dev regressions have been failing for the last 6 days, btw. Makes it tricky to eval the PR.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
Kelvinrr: If we want camera-only serials, is there data in the ISD we can use to make them more unique? That could would allow this to be a bug fix.

oleg-alexandrov: It looks that the current approach of combining platform id, instrument id, and reference time, is enough, assuming no sensor fires more frequently than once per millisecond. This is a rare instance to start with, and we only tripped up on JunoCam which is a quirky pushframe. 

I would be inclined to keep this as is, once the PR for the millisecond is merged in. We could dig up for someting in the ISD, such as some binary hash of whole isd but that may be a solution for which we have no problem yet. That would also result in a massive fail in tests.

oleg-alexandrov: I modified that PR to save the calendar time to the microsecond, not the millisecond. This is not an academic fix. Claude says:

CSMCamera::initFromModel (CSMCamera.cpp:156-159) parses the string back into a numeric iTime m_refTime, then computes every pixel's absolute time as et = m_refTime + getImageTime(imagePt) (lines 221 and 469). This is the one that treats the string as a number, so its resolution sets the absolute-time accuracy the CSM camera hands to SPICE for target rotation, sun and spacecraft position, and illumination angles.

I think this is still minor, but measurable. Then, a millisecond in orbit is a lot. Earth-based satellites use microseconds for their time reporting, because they can scan a handful of scan lines in one millisecond.

oleg-alexandrov:
This passes now that the dev build got fixed. Together with https://github.com/DOI-USGS/usgscsm/pull/534 which got merged it appears solid.

Kelvinrr:
apologies with the delay on follow up review, been in and out sick for a long while.

oleg-alexandrov:
@Kelvinrr sorry to hear that, get well and rest up. No rush.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
```diff
@@ -97,25 +97,55 @@ namespace Isis {
...
+      catch (IException &e) {
+        // Re-throw if there is no CSM fallback, rather than continuing with an
+        // empty mission and instrument.
+        if (!label.findObject("IsisCube").hasGroup("CsmInfo")) {
+          throw IException(e, IException::Unknown,
+                           "Unable to find a serial number translation for this cube.",
+                           _FILEINFO_);
+        }
+        mission = "";
+        instrument = "";
```
Kelvinrr: still seem like dead code

Kelvinrr (commented):
running claude to test this PR locally, it found that some missions now do not get a SN. 

> New hard regression. The instrument branch has no CSM fallback, so any CSM cube with a recognized instrument but no <Mission><Instrument>SerialNumber.trn now fails completely instead of getting a working CSM serial number. Two confirmed instances by cross-referencing every Camera.plugin group against both translation tables and all 60 serial-number .trn files: Rosetta OSIRIS and Hayabusa NIRS. Those cubes now yield "Unknown" and jigsaw refuses them — the same failure this PR set out to fix.

might be worth checking out.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
oleg-alexandrov: Removed the dead code, and the instrument branch now falls back to the CSM serial number when a  recognized instrument has no SerialNumber.trn (Rosetta Osiris, Hayabusa Nirs). 

Added a regression test covering that fallback.

Regressions pass.

isis/src/base/objs/SerialNumber/SerialNumber.cpp
```diff
@@ -97,56 +97,101 @@ namespace Isis {
...
+    if(!useCsm) {
+      try {
+        // We want to use this instrument's translation manager. It's much faster for
+        //   SerialNumberList if we keep the translation manager in memory, so re-reading
+        //   from the disk is not necessary every time. To do this, we'll use a map to store
+        //   the translation managers with a string identifier to find them. This identifier
+        //   needs to have the mission name and the instrument name.
+
+        //  Create the static map to keep the translation managers in memory
+        static std::map<QString, PvlToPvlTranslationManager> missionTranslators;
```
Kelvinrr: reduce these comments. Too long

isis/tests/CSMCameraTests.cpp
```diff
@@ -463,6 +463,59 @@ TEST_F(CSMCameraSetFixture, SerialNumber) {
...
 
 
+// jigsaw reads its input cubes into a SerialNumberList, which requires a unique serial
+// number per cube. A cube run through csminit gains a CsmInfo group. If the serial
+// number is taken from that group (CSMPlatformID, CSMInstrumentId, ReferenceTime), two
+// images that share a reference time, as sub-images of a framing instrument do, receive
+// the same serial number, and SerialNumberList reports "Duplicate serial number" - the
+// failure jigsaw reports when reading such cubes (ISIS #5482, #4240). This test
+// replicates that mechanism: two distinct cubes given the same CsmInfo must still get
+// distinct, instrument-based serial numbers and both be added to the list.
```
Kelvinrr: same here

isis/tests/CSMCameraTests.cpp
```diff
@@ -463,6 +463,59 @@ TEST_F(CSMCameraSetFixture, SerialNumber) {
...
+}
+
+
+// Some cubes have a recognized ISIS instrument label but no matching
+// <mission><instrument>SerialNumber.trn (for example Rosetta Osiris or Hayabusa
+// Nirs). The instrument-based serial number then cannot be built. When such a cube
+// also has a CSM sensor model, the serial number must fall back to the CSM serial
+// number rather than yield "Unknown" and be rejected by jigsaw. This test gives an
+// MGS cube a recognized-but-untranslatable instrument (Rosetta Osiris) together with
+// a CsmInfo group and checks the CSM serial number is produced.
```
Kelvinrr: and here

Kelvinrr (commented):
I think in order to merge this in good consciousness, we need to add an additional error message if jigsaw (or any other app that generates SNs) has a SN mismatch error. That is, Jigsaw tries to match SNs from cubes to a network and it fails to find the images in the network, the error message mentions to regenerate SNs for the network. 

Problem there is we would also need an app for a flag that can do that so that the user doesn't have to recreate the entire network from scratch because they updates ISIS from 10.1 to 10.2.

oleg-alexandrov:
I shortened the comments and added an error when serial numbers for the images can't be found.

I would think introducing a mechanism to fix the serial numbers would be premature. This would apply only to jigsaw with csm cameras, and this feature is so new that likely no such cnets exist.

Kelvinrr (approved):
that's a fair point.

Issue #6137, opened 2026-08-31
Web SPICE not working for ShadowCam


rvwagner:
**ISIS version(s) affected**: 10.1.0_RC1

**Description**  
Now that ShadowCam has made it into the public version of ISIS, it's time to test it!  So far, it seems good except for SPICE.
```
wget https://pds.shadowcam.im-ldi.com/observation/2023/190/M029240108S/M029240108SE.cub
shadowcam2isis from=M029240108SE.cub to=M029240108S.raw.cub
spiceinit from=M029240108S.raw.cub web=true
```

Result:
```
spiceinit: Requesting Spice Data
100% Processed
**ERROR** An error occurred when talking to the server. The server encountered an unexpected error in SpiceClient.cpp at 592.
```

I can spiceinit a NAC image from this computer with the same version of ISIS with no problems, so I'm guessing the issue is that web SPICE isn't fully aware that ShadowCam exists.  downloadIsisData seems to know how to get the ShadowCam data directory (at least the non-kernel parts), so it seems like this is hopefully just a matter of adjusting a flag or adding KPLO kernel-syncing to the server settings.

(Also maybe you could make the error messages more informative?)

PR #6154, opened 2026-09-09
Better error messaging in SpiceClient

chkim-usgs:
## Description

When `spiceinit web=true` failed, the client discarded the server's actual error and showed only a generic message like `An error occurred when talking to the server` because SpiceClient only knew how to decode two response formats: hex-encoded XML and PVL. The current spiceserver web service reports failures as JSON `{"body": {"error": "..."}}`, which fell through both decoders and landed in the generic network-error branch, so the real cause was thrown away.

## Related Issue

Partially addresses https://github.com/DOI-USGS/ISIS3/issues/6137

## Types of changes

## Checklist:

## Licensing


Kelvinrr:
can you update the PR description?

isis/src/base/apps/spiceinit/SpiceClient.cpp
```diff
@@ -207,7 +210,26 @@ namespace Isis {
...
+          QJsonValue body = jsonDoc.object().value("body");
+          if (body.isObject()) {
+            serverError = body.toObject().value("error").toString();
+          }
+        }
+
+        if (!serverError.isEmpty()) {
+          *p_error = "spiceserver was unable to initialize the cube.  "
+                     "The error reported was: ";
+          *p_error += serverError;
```
Kelvinrr, line 232: might be worth reporting the HTTP status code

Kelvinrr (commented):
can we test this? Should be easy since it doesn't have to hit a real server to get an error back. A 404 would test it.

isis/src/base/apps/spiceinit/SpiceClient.cpp
chkim-usgs: Was able to get a local instance of spiceserver up to test various changes, modified it so it responded only 404 error and got the specified error message and the added `(HTTP status code <code>)` as expected:
```
(isis-dev) ➜  shadowcam spiceinit web=true url=http://localhost:5000/ from=M029240108S.raw.cub
**ERROR** spiceserver was unable to initialize the cube.  The error reported was: Forced test failure (SPICESERVER_FORCE_STATUS=404) (HTTP status code 404).
```

Kelvinrr: For 404 I dont think you even need a test server, you can just put in a fake URL and it'll fail since the server doesn't exist. If we go the test server route then we can have more comprehensive tests than just 404.

Kelvinrr (approved):

Issue #6161, opened 2026-09-19
cam2map USEPROJ: non-PROJ4-format PROJStrings can silently corrupt ground range or crash due to axis order


ljbade:
## Summary

`cam2map`'s USEPROJ/PROJString option accepts any string PROJ's generic `proj_create()` recognises -- not just a raw PROJ4-style `+key=value` string, but also WKT (WKT1 and WKT2), authority codes (`EPSG:...`, `IAU_2015:...`), the `urn:ogc:def:crs:...` form, and PROJJSON. For a CRS built from any of these non-`+proj=` formats, this can silently corrupt the resulting map projection's ground range, or crash `cam2map` outright.

## Root cause

`IProj`'s constructor (`isis/src/base/objs/IProj/IProj.cpp`) extracts the base geodetic CRS directly from the user-supplied output CRS:

```cpp
m_llaProj = proj_crs_get_geodetic_crs(m_C, m_outputProj);
...
m_llaProj2outputProj = proj_create_crs_to_crs_from_pj(m_C, m_llaProj, m_outputProj, 0, 0);
```

`m_llaProj`'s axis order is whatever the output CRS's authority declares. A raw PROJ4-style string is always legacy (lon, lat) order, but a WKT2/authority-code/PROJJSON CRS can declare (lat, lon) per ISO 19111 -- for example `IAU_2015:49920` ("Mars (2015) - Sphere / Ocentric / Sinusoidal, clon = 0") declares its base geographic CRS as `AXIS["geodetic latitude (Lat)", ORDER[1]]` then `AXIS["geodetic longitude (Lon)", ORDER[2]]`.

`IProj::SetGround()` then unconditionally assumes traditional order:

```cpp
c_in.lpz.lam = m_longitude;
c_in.lpz.phi = m_latitude;
PJ_COORD c_out = proj_trans(m_llaProj2outputProj, PJ_FWD, c_in);
```

When the transform pipeline actually expects (lat, lon) order, this feeds a longitude value into the slot PROJ treats as latitude, causing `proj_trans` to reject it (`sinu: Invalid latitude`, since e.g. -104 is not a valid latitude). `cam2map`'s default (camera-derived, un-overridden) ground-range computation swallows these per-point failures and proceeds with a corrupted range, ultimately crashing with `Invalid cube size specifications [ns=-2147483648,nl=-2147483648,nb=1]`.

[...]

PR #6163, opened 2026-09-20
Restrict cam2map USEPROJ PROJString to PROJ4 string format

ljbade:
## Description
This PR contains a bug fix for `cam2map`'s USEPROJ/PROJString functionality:

The USEPROJ/PROJString option passed the user's string straight to PROJ's generic `proj_create()`, which accepts far more than a raw PROJ4-style `+key=value` string -- authority codes (`EPSG:...`, `IAU_2015:...`), WKT, and PROJJSON all work too. `IProj` takes its transform's axis order straight from the output CRS's own base geodetic CRS: a raw PROJ4-style string is always legacy (lon, lat) order, but other formats can declare (lat, lon). `IProj` always assumes (lon, lat) order when feeding coordinates to `proj_trans()`, so a (lat, lon)-declared CRS gets fed a longitude value where PROJ expects a latitude, and `cam2map`'s default ground-range computation ends up crashing with `"Invalid cube size specifications"`. GDAL's own reader for this same `ProjStr` keyword (added upstream in [OSGeo/gdal#14894](https://github.com/OSGeo/gdal/issues/14894)) only ever calls `importFromProj4()` on it, so a cube written in any of these other formats could never round-trip through GDAL anyway. This restricts `PROJString` to strings starting with `+`, and adds `proj_normalize_for_visualization()` in `IProj` so the axis order is forced to (lon, lat) regardless of how a source CRS declares its own axes.

## Related Issue
Fixes issues #6161

Added test coverage for the issue fixed in this PR:
- `FunctionalTestCam2mapUseProjIauCodeFails`: an IAU authority code (e.g. `IAU_2015:49920`) is rejected rather than passed to PROJ
- `FunctionalTestCam2mapUseProjWktFails`: a hand-written WKT CRS is rejected rather than passed to PROJ
- `FunctionalTestCam2mapUseProjJsonFails`: a PROJJSON CRS is rejected rather than passed to PROJ
- `IProjTest.TestIProjLatLonAxisOrder`: constructs `IProj` directly from a CRS that declares (latitude, longitude) axis order and verifies `SetGround`/`SetCoordinate` still produce correct, round-tripping results

Ran the full test suite and all tests pass.

## Types of changes

**NOTE:** I checked the "breaking change" box as previosuly `cam2map` would accept a `projstring` that wasn't in the `+key=value` format such as an authority code, WKT or PROJJSON. However as issue #6161 describes using these other formats would either result in incorrect output or a crash.

## Checklist:
- [x] My code follows the code style of this project.

*Note:* I used generative AI, specifically Anthropic Claude Code to help me create this PR, but I have reviewed all the code and test results myself.

## Licensing


isis/tests/IProjTests.cpp
```diff
@@ -106,4 +106,33 @@ TEST_F(IProjMappingGroup, TestIProjXYRange) {
...
\ No newline at end of file
+}
+
+// IAU_2015:49920 declares its base geodetic CRS axes as (latitude, longitude) per ISO 19111,
+// unlike a raw "+proj=" string which is always (longitude, latitude). This exercises
+// proj_normalize_for_visualization() in IProj's constructor directly: without it,
+// SetGround()/SetCoordinate() would silently swap latitude and longitude.
+TEST(IProjTest, TestIProjLatLonAxisOrder) {
+  PvlGroup mapping("Mapping");
+  mapping.addKeyword(PvlKeyword("ProjStr", "IAU_2015:49920"));
```
acpaquette, line 117: From this it seems like authority codes work under the current system. It seems like it would be worth allowing them through the same `projStr` parameter with some updated docs

isis/src/base/apps/cam2map/cam2map.xml
```diff
@@ -402,8 +402,12 @@
         <type>string</type>
         <brief>A PROJ4 string that defines the projection</brief>
         <description>
-          This parameters is used instead of the map file to define the projection to
-          transform into. For help with PROJ strings visit https://proj.org/en/9.3/index.html
+          This parameter is used instead of the map file to define the projection to
```
acpaquette, line 405: I have a comment about this on one of the tests, but it seems like allowing authority codes would be useful especially since we are accounting for the lat/lon flip

acpaquette (commented):
Changes look good! A couple of comments to address

ljbade:
@acpaquette I am hesitant about allowing other formats now in the `ProjStr` as it will cause compatibility issues with the ISIS cube decoder in GDAL.

In the GDAL PR that implements the `ProjStr` https://github.com/OSGeo/gdal/pull/14894 the value is passed through GDAL's `importFromProj4` at https://github.com/OSGeo/gdal/pull/14894/changes#diff-42688dcf36351fdaa63b2a5ddbd46e9bdfe7b39d5642d27f1278d7aae4c5f073R2044

If you look at source for that function you will see that it basically restricts strings to start with `+` character.

So if if authority code is allowed in `ProjStr` it will work fine amongst the ISIS tools, but will fail to import if some uses the GDAL library or tools to work on a cube file with an authority code.

My suggestion is that if we want to support authority codes, instead we add a new `WKTStr` option which is mutually exclusive with `ProjStr`. This new string will support WKT strings, which in the PROJ library is what authority codes are translated to anyway.

And then introduce command line options to set a full WKT string, or an authority code, both of which will then be encoded to `WKTStr`. Of course GDAL will then need an update to support this, but the GDAL maintainer already mentioned in the PR that this was the recommended approach for adding CRS to new file formats anyway. See https://github.com/OSGeo/gdal/pull/14894#issuecomment-4863642742

